PR #8, opened 2014-12-30
fix getChannelsWithUnreads

natesilva:
Instead of checking whether the `unread_count` property exists, check whether it is `> 0`.


natesilva:
To clarify: This fix is needed because `getChannelsWithUnreads` is broken due to a bad assumption in the code about the CoffeeScript question mark operator.

evansolomon:
Nice find

PR #10, opened 2015-01-13
emit "open" event after receive PONG

shokai:
I have a hubot that notify after it has been deployed.
## bug (?)

In following hubot script, first `robot.send` will not be sent.
```coffee
# Description:
#   deply notifer

module.exports = (robot) ->

  robot.send {room: "news"}, "Hubot deployed!"  ## this message will not send

  setTimeout ->
    robot.send {room: "news"}, "wait 5(sec), Hubot deployed!"  ## this is ok
  , 5000
```

## current implementation
node-slack-client emits `open` event after websocket was connected.
hubot-slack adapter receives `open` event, then emit `connected` event.
hubot receives `connected` event from adapter, then load `scripts/*.coffee`.

## fixed
emit `open` event after send PING and receive first PONG

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -99,10 +99,9 @@ class Client extends EventEmitter
     else
       @ws = new WebSocket @socketUrl
       @ws.on 'open', =>
-        @emit 'open'
         @connected = true
         @_connAttempts = 0
-        @_lastPong = Date.now()
+        @_lastPong = null
 
         # start pings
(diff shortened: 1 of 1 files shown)


paulhammond:
Thanks for noticing this and pointing it out with an easy to reproduce bug report.

I've done some digging internally about what went wrong here, and I've taken a slightly different approach to fixing it, which is in #19. We'll get that merged into the next release!

PR #13, opened 2015-01-26
Implement star_added and star_removed

sillygwailo:
Emit the star_added and star_removed events and send along the object of the message that was starred (or had its star removed).

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -342,6 +342,14 @@ class Client extends EventEmitter
       when "error"
         @emit 'error', message.error
 
+      when "star_added"
+        m = new Message @, message
+        @emit 'star_added', m
+
+      when "star_removed"
+        m = new Message @, message
+        @emit 'star_removed', m
+
       when "message"
         # is this the special message we get on reconnect?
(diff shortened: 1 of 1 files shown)


evansolomon:
Seems fine to me but from [this comment](https://github.com/slackhq/node-slack-client/blob/master/src/client.coffee#L319) it seems like maybe these specially-handled events are not meant to be public API's?  I'm not sure though, maybe one of the Slack guys can weigh in.

paulhammond:
[star_added](https://api.slack.com/events/star_added) and [star_removed](https://api.slack.com/events/star_removed) are both documented, supported parts of our public API!

paulhammond:
Oops, I just hit the merge button by mistake and messed up this pull request. I'm sorry about that!

SO: while star_added and star_removed are both documented events, the code here only handled the case where a message was starred. But, other things can be starred in Slack too - including files, file comments, and channels.

I'd love to merge a pull request that handles star_added or star_removed on any of these objects!

paulhammond:
#27 is the follow up pull request

PR #14, opened 2015-01-30
Add Capability to modify messages via the api

fabianthoma:
This adds the capability to edit the messages by calling a method on the message object itself. The send message function on the Client also sends back the message to the caller, this even works in hubot-slack.


paulhammond:
Nice! Thanks for sharing your code!

PR #15, opened 2015-01-30
Add support for chat.postMessage API method

paulhammond:
This pull request adds support for the [chat.postMessage](https://api.slack.com/methods/chat.postMessage) API method

### DIFF
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -86,6 +86,19 @@ class Channel
     m = new Message @_client, {text: text}
     @sendMessage m
 
+  postMessage: (data) ->
+    params = data
+    params.channel = @id
+    if data.attachments
+      params.attachments = JSON.stringify(data.attachments)
+
+    @_client.logger.debug data
+    @_client.logger.debug params
+    @_client._apiCall "chat.postMessage", params, @_onPostMessage
+
+  _onPostMessage: (data) =>
+    @_client.logger.debug data
+
   sendMessage: (message) ->
     message.channel = @id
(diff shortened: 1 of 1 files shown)


silentrob:
This is super useful. Just a side note, Bot users are not authorized to call this endpoint.

paulhammond:
Actually, as of a few minutes ago, they are. As I mentioned over in slackhq/hubot-slack#148 this is provisional and might change but probably won't...

silentrob:
:beers: :+1:

src/channel.coffee
```diff
@@ -86,6 +86,19 @@ class Channel
...
+    params = data
+    params.channel = @id
+    if data.attachments
+      params.attachments = JSON.stringify(data.attachments)
+
+    @_client.logger.debug data
+    @_client.logger.debug params
+    @_client._apiCall "chat.postMessage", params, @_onPostMessage
+
+  _onPostMessage: (data) =>
```
evansolomon: Just a thought...

There are a bunch of these `_on{{ApiMethod}}` methods that just call the logger.  Not really *bad* but might be easier to just pass around the logger directly instead of a bunch of different methods that wrap it (unless there's some other purpose I'm missing).

paulhammond: A fair point - I was just copying the existing pattern. I'm open to other approaches here!

evansolomon: Yea you definitely didn't start it here.  Not a big deal, just something I noticed.

PR #17, opened 2015-01-31
Add ability to delete messages

paulhammond:
This pull requests builds on the pattern from #14, and adds the ability to delete messages.

### DIFF
--- a/src/message.coffee
+++ b/src/message.coffee
@@ -70,6 +70,18 @@ class Message
   _onUpdateMessage: (data) ->
     @_client.logger.debug data
 
+  deleteMessage: =>
+    params = {
+      "ts": @ts,
+      "channel": @channel,
+    }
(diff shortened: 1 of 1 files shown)


src/message.coffee
```diff
@@ -70,7 +70,19 @@ class Message
...
     @_client.logger.debug data
 
+  deleteMessage: =>
+    params = {
+      "ts": @ts,
+      "channel": @channel,
+    }
+    @_client.logger.debug "Sending message delete request"
+    @_client.logger.debug params
+    if @ts then @_client._apiCall "chat.delete", params, @_onDeleteMessage
```
evansolomon: If the request is going to be conditionally sent I think the logging should go in the same conditional.

paulhammond: Good point, fixed in bdfb0fa60dabb09988e331a4aade2f6fb8cd9b36

PR #19, opened 2015-02-03
Emit open event after hello, not before

paulhammond:
This pull request takes a different approach to fixing the bug identified in #10. Previously messages sent immediately after the client has emitted the "open" event wouldn't appear in channel.

Waiting until after we've received a "hello" before emitting `open` fixes the bug. This is better than waiting for the next ping/pong timeout as it means you can respond immediately to the connection being established, instead of waiting a few seconds.


grantmd:
:+1:

PR #20, opened 2015-02-04
Optionally specify a callback to client.openDM to get DM channel id

bertrandom:
im.open returns a DM channel id like `D024BFF1M` but the caller of client.openDM has no way to get this information in order to pass to getDMByID and send messages via the channel. This allows you to specify an optional callback to get the channel ID returned by the API.

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -175,12 +175,15 @@ class Client extends EventEmitter
   _onJoinChannel: (data) =>
     @logger.debug data
 
-  openDM: (user_id) ->
+  openDM: (user_id, callback) ->
     params = {
       "user": user_id
     }
 
-    @_apiCall 'im.open', params, @_onOpenDM
+    if typeof callback is 'undefined'
+      callback = @_onOpenDM
+
+    @_apiCall 'im.open', params, callback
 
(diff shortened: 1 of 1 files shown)


evansolomon:
Makes sense.  I think it would actually be a good idea to support this for anything that ends up hitting `_apiCall`.

paulhammond:
We should probably do this everywhere instead of this one method.

We should also find a way to keep the internal callback and have it call the user specified callback. Today the internal callback only does logging, but there's no guarantee that will be true in the future...

jtatum:
What if _apiCall took two callback parameters? Or an array of callbacks?

evansolomon:
Here's a simple pass at exposing callbacks in anything that hits apiCall 92fcbf0e09768fc4d0344818b0082f3b4852e484

It's probably not the best solution in the world, but the API/behavior seems about right to me.

sillygwailo:
I definitely want user-specified callbacks on anything that's an API call, channel.send() in particular.

paulhammond:
I like @evansolomon's approach, so I just merged it! I think it solves all of the needs people have brought up here...

PR #22, opened 2015-02-14
Add files and comments

yahsan2:
I added file and comment Classes.
and, I added eventCamelName function.

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -513,6 +513,7 @@ class Client extends EventEmitter
       when "file_comment_added", "file_comment_edited"
         if @files[message.file.id]
           c = message.comment
+          c.file_id = message.file.id
           @files[message.file.id].comments[c.id] = @files[message.file.id].newComment c
           @emit @eventCamelName(message.type), c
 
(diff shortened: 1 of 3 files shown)


paulhammond:
Hi! thanks for sharing this!

Unfortunately some of the other changes we've just merged mean that we can't merge this as is. Would you mind updating so it's merge-able? While you're at it, you'll probably want to include the changes to add custom callbacks to your new functions (see #20 and #30 for details)

I'm also not sure about the `eventCamelName` function - it means it's much harder to grep the code for where a particular function is called, and it doesn't improve the readability of the code. I think

    @emit @eventCamelName(message.type), f

is more confusing than:

    @emit "fileCreated", f

If you can fix both I'd be happy to get this merged!

ekmartin:
Are you still planning to finish this @yahsan2? Would be nice, especially since it adds file events :)

Tehnix:
Was wondering what the progress on this was? Else I might try to take a stab at this, since I would really like attachments etc to show up (am currently using @ekmartin's [slack-irc](https://github.com/ekmartin/slack-irc) integration) :)

PR #25, opened 2015-02-17
Updating .travis.yml and package.json to require node 0.10 or newer

jtatum:
The versions of grunt specified in package.json do not run on node 0.8. Grunt and its cohorts aren't required for the package itself or the test cases, so .travis.yml will remove them from package.json prior to installing dependencies and running tests.

Successful Travis build: https://travis-ci.org/jtatum/node-slack-client/builds/51145556


evansolomon:
Is anyone using Node 08 these days?  I think I'd actually drop 0.8 support/testing, but someone from Slack should probably make that choice.

paulhammond:
We're ok with dropping it!

jtatum:
Updated Travis build at https://travis-ci.org/jtatum/node-slack-client/builds/51156823

PR #27, opened 2015-02-20
Implement stars and removed stars

sillygwailo:
This is for everything, not just messages. So files, channels, etc. are covered. It's up to the app to react to what's being starred/unstarred in the item.type property.

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -135,7 +135,7 @@ class Client extends EventEmitter
       @ws.on 'star_added', (data, flags) =>
         @onStarAdded JSON.parse(data)
 
-      @ws.on 'star_added', (data, flags) =>
+      @ws.on 'star_removed', (data, flags) =>
         @onStarRemoved JSON.parse(data)
 
       return true
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -132,6 +132,12 @@ class Client extends EventEmitter
       @ws.on 'ping', (data, flags) =>
         @ws.pong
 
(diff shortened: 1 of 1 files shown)


src/client.coffee
```diff
@@ -132,6 +132,12 @@ class Client extends EventEmitter
       @ws.on 'ping', (data, flags) =>
         @ws.pong
 
+      @ws.on 'star_added', (data, flags) =>
+        @onStarAdded JSON.parse(data)
+
+      @ws.on 'star_added', (data, flags) =>
+        @onStarRemoved JSON.parse(data)
```
jtatum: Is this a typo? Calling onStarRemoved when star_added is emitted?

sillygwailo:
Typo fixed. Now that I test it again, Slack's API isn't sending me 'star_added' or 'star_removed' events in my test app. Hmm.

sillygwailo:
I think I have it now. The confusion lay in what what the ws module calls messages and what the Slack app calls messages.

The latest commit added the event to the onMessage() handler, which handles ws messages. That handler switches on events Slack sends along. It stores the raw message object in the 'message' variable, so the star_added and star_removed events send that along too (and not, as I had in a previous PR, just sent along the object of class Message).

Does that make sense? This commit sends along the more abstract 'message' that ws sends to onMessage, and the app using this library will have to handle the object it receives by looking at the item.type property.

paulhammond:
Thanks for working through this - this updated version looks great (and sorry again for messing up your last pull request!)

PR #28, opened 2015-02-20
Ported simple.js to coffeescript

jvelliott:
Tried to do a direct conversion to help any other noobs out there trying to learn coffeescript.

A couple list comprehensions. Not sure if there's a nicer way to do ternary in coffeescript?


evansolomon:
Your coffeescript "ternary" is generally right in that you're using an if/else as an expression.  The maybe "nicer" way would be to one line it: `result = if condition then truthyResult else falselyResult`

I'll leave a couple other comments on your commit with some more coffeescript tips (not that anything you did was wrong).

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,73 @@
...
+
+  console.log "You have #{unreads} unread #{messages}"
+
+
+slack.on 'message', (message) ->
+  type = message.type
+  channel = slack.getChannelGroupOrDMByID(message.channel)
+  user = slack.getUserByID(message.user)
+  time = message.ts
+  text = message.text
```
evansolomon: There's a nice desctructuring syntax in cs: `{type, ts, text} = message`

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,73 @@
...
+slack.on 'open', ->
+  channels = []
+  groups = []
+  unreads = slack.getUnreadCount()
+  key
+
+  # Get all the channels that bot is a member of
+  channels = ("##{value.name}" for key, value of slack.channels when slack.channels[key].is_member)
+  # Get all groups that are open and not archived 
+  groups = (value.name for key, value of slack.groups when slack.groups[key].is_open and not slack.groups[key].is_archived)
```
evansolomon: This is probably getting on the verbose side for what I'd consider for a one liner.  It is correct, just might be a little more readable if it was split up.

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,73 @@
...
+autoReconnect = true
+autoMark = true
+
+slack = new Slack(token, autoReconnect, autoMark)
+
+slack.on 'open', ->
+  channels = []
+  groups = []
+  unreads = slack.getUnreadCount()
+  key
```
evansolomon: You don't need this, CoffeeScript will handle the variable declaring for you.  This may actually throw a reference error at runtime, not sure though.

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,73 @@
...
+
+slack.on 'message', (message) ->
+  type = message.type
+  channel = slack.getChannelGroupOrDMByID(message.channel)
+  user = slack.getUserByID(message.user)
+  time = message.ts
+  text = message.text
+  response = ''
+
+  channelName =
```
evansolomon: Would probably one line this one, `channelName = if channel.is_channel then '#' else ''`

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,73 @@
...
+  channelName += channel.name
+
+  console.log "Received: #{type} #{channelName} @#{user.name} #{time} \"#{text}\""
+
+  # Respond to messages with the reverse of the text received.
+
+  if type is 'message'
+    response = text.split('').reverse().join('')
+    channel.send response
+    console.log "@#{slack.self.name} responded with \"#{response}\""
```
evansolomon: Escaping double quotes is fine, or you can use multi-line strings and not have to.

```coffeescript
console.log """
  @#{slack.self.name} responded with "#{response}"
"""
```

jvelliott:
Thanks for the feedback.  I'll update it over the weekend.

Do you prefer I make a separate commit, or rebase onto this one?

evansolomon:
> Do you prefer I make a separate commit, or rebase onto this one?

Whatever you like.

I think it also makes sense to actually replace the JS version.  Having both in the source, especially when they're not actually the "same" code, seems confusing.  To do that it should be added to the `grunt prepublish` build step.

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env coffee
```
evansolomon: You don't need this unless the file is meant to be executable (which it probably shouldn't be).

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,73 @@
...
+# This is a simple example of how to use the slack-client module in CoffeeScript. It creates a
+# bot that responds to all messages in all channels it is in with a reversed
+# string of the text received.
+#
+# To run, copy your token below, then:
+#  npm install
+#   cd examples
+#   coffee simple_reverse.coffeescript
```
evansolomon: Wrong extension.  Also probably simpler to compile this to JS and use `node` to run it (these instructions as is would also require having coffee-script globally installed).

jvelliott:
>I think it also makes sense to actually replace the JS version.

OK, that sounds good to me.  Should I go through and comment on the new compiled version, or just leave it alone?

You do actually want it committed, right?

evansolomon:
Let's not commit the compiled JS. It can be compiled by the same grunt task that handles the rest of the library. So the JS file should also be removed from Git and gitignored.

Let me know if any of that doesn't make sense or if you want any help.

Gruntfile.coffee
```diff
@@ -26,6 +26,12 @@ module.exports = (grunt) ->
       index:
         files:
           'index.js': 'index.coffee'
+      examples:
+        expand: true
+        cwd: 'examples'
+        src: ['*.coffee']
+        dest: 'examples'
+        ext: '.js'
```
jvelliott: Hope this looks good. Never worked with grunt before!

evansolomon: :+1:

examples/simple_reverse.coffee
```diff
@@ -0,0 +1,76 @@
...
+  channels = []
+  groups = []
+  unreads = slack.getUnreadCount()
+
+  # Get all the channels that bot is a member of
+  channels = ("##{value.name}" for key, value of slack.channels when slack.channels[key].is_member)
+  # Get all groups that are open and not archived 
+  groups = (for key, value of slack.groups
+    if slack.groups[key].is_open and not slack.groups[key].is_archived
+      value.name)
```
jvelliott: Just noticed that this expanded loop, as opposed to the comprehension, seems to be adding an "void 0" to the end of the groups array.

I can prevent this with an empty *else* block, but that seems gross.  Is there a nicer way?

```
groups = (function() {
  var ref, results;
  ref = slack.groups;
  results = [];
  for (id in ref) {
    group = ref[id];
    if (group.is_open && !group.is_archived) {
      results.push(group.name);
    } else {
      results.push(void 0);
    }
  }
  return results;
})();
```

evansolomon: Not really.  You could leave it as a comprehension, you could do something like `Object.keys(slack.channels).reduce(...)`, or you could stick `.filter(Boolean)` onto the end (I personally use this method a lot).

jvelliott: I'm going to put it back as a comprehension, then.  I rewrote it slightly to be a bit more obvious what it was doing.

jtatum:
Not sure if this belongs here, but there was something I was going to poke at the sample code about. There are a variety of things that will crash the javascript sample, like deleting a message. Some 'message' subtypes have no user or text associated with them so the console.log throws.

jvelliott:
Hah, it's a fair point.  Guess I never noticed that deleting/editing/a bunch of other stuff all hits the on "message" callback.

Seems to me like either the project or the APIs are being pretty overbroad with those things, but it can't hurt to do a bit of null checking.

evansolomon:
@jvelliott is this all ready to merge?

jvelliott:
As far as I know!

jvelliott:
Rebased commits so they all show up together after merge.

evansolomon:
Thanks!

PR #31, opened 2015-03-03
Test for valid (non-null) socket @ws in Client.reconnect() before @ws.cl...

kachizukanne:
Test for valid (non-null) socket @ws in Client.reconnect() before @ws.close() to avoid system failure if the bot is disabled

image

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -154,7 +154,10 @@ class Client extends EventEmitter
       @_pongTimeout = null
 
     @authenticated = false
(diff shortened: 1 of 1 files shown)


kachizukanne:
Hi,

I reckon this update may be necessary to elegantly manage this exception. Given the trace for the call (see above image) I'm unable to trap and manage the exception from the caller

davidkassa:
+1. Also, should this check be made on the disconnect method?

PR #36, opened 2015-04-18
Fix Message Callbacks

CharlieHess:
Posting a message to a channel and then calling `updateMessage` or `deleteMessage` on that instance would throw an exception:
```
/Users/charlie/slack-poker-bot/node_modules/slack-client/src/message.js:106
    return this._client.logger.debug(data);
                       ^
TypeError: Cannot read property 'logger' of undefined
    at Message._onUpdateMessage (/Users/charlie/slack-poker-bot/node_modules/slack-client/src/message.js:106:24)
    at IncomingMessage.<anonymous> (/Users/charlie/slack-poker-bot/node_modules/slack-client/src/client.js:668:22)
    at IncomingMessage.emit (events.js:117:20)
    at _stream_readable.js:943:16
    at process._tickCallback (node.js:419:13)
```

PR #37, opened 2015-04-19
Use fat arrows for Message update/delete callbacks

programble:
Callbacks were getting called unbound, resulting in undefined `@_client`.


programble:
Oops, duplicate of #36.

PR #42, opened 2015-06-01
Add "sendRaw" method to the Channel class

Kami:
This method works the same as "send", but it disables Slack message formatting. This allows users to send "raw" messages.

API reference where `parse` attribute is documented:

* https://api.slack.com/methods/chat.postMessage
* https://api.slack.com/docs/formatting (see "Parsing modes" section)

### DIFF
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -86,6 +86,10 @@ class Channel
     m = new Message @_client, {text: text}
     @sendMessage m
 
+  sendRaw: (text) ->
+    m = new Message @_client, {text: text, parse: "none"}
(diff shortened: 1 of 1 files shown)


l12s:

It decouples the model objects (channels and so forth) from the client layer, moving the `send` and `sendMessage` functionality onto [the RTM client itself](https://github.com/slackhq/node-slack-client/blob/master/lib/clients/rtm/client.js#L507).

However, the RTM client also doesn't support raw messages right now, so I've created an issue to track this: https://github.com/slackhq/node-slack-client/issues/138

PR #45, opened 2015-06-11
Make Channel#_onFetchHistory work

atmos:
I wanted this in a recent project and @lstoll was kind enough to find what was missing and implement it.

The dev process for running this repo is kinda complicated so a new release would be greatly appreciated. :smiley:

### DIFF
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -118,7 +118,9 @@ class Channel
     @_client._apiCall method, params, @_onFetchHistory
 
   _onFetchHistory: (data) =>
-    @_client.logger.debug data
+    for message in data['messages']
+      message['channel'] = @id
+      @_client.onMessage message
 
   mark: (ts) ->
     params = {
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -118,7 +118,10 @@ class Channel
     @_client._apiCall method, params, @_onFetchHistory
 
   _onFetchHistory: (data) =>
-    for message in data['messages']
+    messages = data['messages']
+    messages.sort (a,b) ->
+      a.ts - b.ts
+    for message in messages
       message['channel'] = @id
       @_client.onMessage message
 


97-109-107:
@atmos could you please share an example of how you used this change to read history? Is this just a regular event that can be interpreted with:

````

  this.slack.on('message', function(msg) {
    ...
  });

````

If yes, how do you differentiate between a 'real' message in the room versus a result of fetch history?

atmos:
I'm calling this method after the `login` event is emitted and backfilling the channels that I want.

One drawback is definitely the interaction w/ resetting messages as marked as read. I'm unsure if that's something I need to fix in the client library or if there's a better way to query the API to avoid changing this state.

lstoll:
@97-109-107 we don't differentiate between historical and new messages on events, they're just messages to be handled. Maybe we could set a field on it to indicate that it's historic?

atmos:
I figured out the problem w/ marking messages as unread, we were using the `autoMark` attribute on the client incorrectly.

atmos:
@97-109-107 Here's an example of how I'm using it. https://github.com/atmos/peonies/blob/71cdceecc98af8b8496e4424d3395a38fdc88565/src/main.coffee#L22-L28

97-109-107:
@lstoll I believe that might be beneficial to other users. In my attemts to solve this for my own use case I tried adding a subtype like so:

````
  _onFetchHistory: (data) =>
    for message in data['messages']
      message['channel'] = @id
      message['subtype'] = 'history'
````

@atmos thanks for the example

atmos:
Are there a lot of places that switch on message subtype? I was trying to avoid having to add another special case and I believe subtype is useful to preserve.

brianantonelli:
+1

l12s:

That update doesn't include support for what you're describing here yet, but unfortunately deprecates your PR. I've created https://github.com/slackhq/node-slack-client/issues/140 to track the work required to make this happen.

PR #47, opened 2015-06-11
bump ws 0.4.31 -> 0.8.1

vanm:
Bumps the ws npm to a more recent version.  Noticing quite a few reconnects / connection issues when connected for extended periods of time, so updating `ws` was low hanging fruit.

Verified that the API calls and events in use by slack-client are unchanged: https://github.com/websockets/ws/blob/master/doc/ws.md

There is a change in default protocol, though:
```
Protocol support
Hixie draft 76 (Old and deprecated, but still in use by Safari and Opera. Added to ws version 0.4.2, but server only. Can be disabled by setting the disableHixie option to true.)
HyBi drafts 07-12 (Use the option protocolVersion: 8)
HyBi drafts 13-17 (Current default, alternatively option protocolVersion: 13)
```
In my testing I haven't seen any issues, but would be good to hear from the Slack folks if we should specify protocolVersion: 8.

This is definitely a critical component for the library, so I'm not in a rush to get it merged until we get more eyes on it ;)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,12 +11,7 @@
     "slack"
   ],
   "author": "Slack Technologies, Inc.",
-  "licenses": [
-    {
-      "type": "MIT",
-      "url": "http://github.com/slackhq/node-slack-client/raw/master/LICENSE"
-    }
-  ],
+  "license": "MIT",
   "repository": {
     "type": "git",
     "url": "git://github.com/slackhq/node-slack-client.git"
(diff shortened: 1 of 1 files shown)


vanm:
Haven't experienced any issues caused by this version bump.  Been running in production since around mid-June.

l12s:
Thanks, I'll check this out and confirm the upgrade looks good.

ChrisCinelli:
Can we bump ws version please?
It is still broken on node 0.12.

vanm:
Rebased with master and updated to ws 0.8.1.

Issue #51, opened 2015-06-19
Handle team_migration_started events


andebor:
The client seems to be missing handling for [team_migration_started](https://api.slack.com/events/team_migration_started) events.

```
[Tue May 26 2015 18:24:22 GMT+0000 (UTC)] DEBUG Unknown message type: team_migration_started
[Tue May 26 2015 18:24:22 GMT+0000 (UTC)] DEBUG { type: 'team_migration_started' }
```

When this event occurs the client gets disconnected and doesn't get back online. As stated in the docs, a reconnect should be initated when this message is received. This might be related to #5.

mrsmn:
:+1:

voneiden:
Also seeing this happening.. on average perhaps once a month.

vanm:
@grantmd @paulhammond @evansolomon Any chance you guys can review this?  I've been logging this event and am seeing it regularly and slack-client isn't reconnecting w/o this change.

ngtjah:
Seeing this problem almost daily..

```
[Thu Nov 05 2015 20:41:02 GMT-0700 (MST)] DEBUG pong
[Thu Nov 05 2015 20:41:07 GMT-0700 (MST)] DEBUG ping
[Thu Nov 05 2015 20:41:08 GMT-0700 (MST)] DEBUG pong
[Thu Nov 05 2015 20:41:12 GMT-0700 (MST)] DEBUG ping
[Thu Nov 05 2015 20:41:12 GMT-0700 (MST)] DEBUG pong
[Thu Nov 05 2015 20:41:14 GMT-0700 (MST)] DEBUG Unknown message type: team_migration_started
[Thu Nov 05 2015 20:41:14 GMT-0700 (MST)] DEBUG { type: 'team_migration_started' }
```

pong ping stops and then I have 1-way communication

l12s:
Hey,

I'm going to give a proper update on this soon, but we're working on doing some significant client refactoring that'll address this issue and more. I'll be putting together a full update and change log in the next week or so that'll talk to what that will look like.

I appreciate your patience while we work on this!

l12s:
Ok, first off, I apologize for how long it's taken us to resolve these issues.

I just pushed the 2.0.0 release for this library, which is a ground up refactor with the aim of building a lib that both Slack and our external contributors can easily work with and improve.

[...]

PR #54, opened 2015-07-02
Handle team_migration_started message + tests

vanm:
Attempt to address #51 

As the issue suggests, it's hard to reproduce.  This fix calls reconnect (close websocket and login again) and assumes the `migration_in_progress ` error referenced in https://api.slack.com/events/team_migration_started will come through rtm.start response within _onLogin's `!data.ok` error handler.  In that case, it will use the normal reconnect with backoff and respect the `autoReconnect` option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,9 @@
     "grunt-contrib-watch": "~0.5.3",
     "grunt-release": "~0.6.0",
     "grunt-shell": "~0.5.0",
+    "mocha": "~1.13.0",
     "should": "~2.0.2",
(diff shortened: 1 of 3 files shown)


ulion:
reconnect is a bad function, which increase the sleep wait each time, and multiple calls to reconnect may cause more problem

ekmartin:
This would be really helpful.

vanm:
Rebased w/ master last week so this should be good to merge.  Unless there is a different preferred approach @l12s

PR #57, opened 2015-07-10
Wrap JSON.parse in try/catch blocks

structAnkit:


l12s:

It also provides much better handling for [JSON parse errors](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/clients/client.js#L165)

PR #58, opened 2015-07-10
Add support for reaction events.

sillygwailo:
The object from this looks like this (unique IDs redacted):

Reaction added to a file:

    {
      type: 'reaction_added',
      user: '[REDACTED]',
      item: {
        type: 'file',
        file: '[REDACTED]'
      },
      reaction: 'grin',
      event_ts: '[REDACTED]' 
    }

Reaction removed from a message:

    {
      type: 'reaction_removed',
      user: '[REDACTED]',
      item: {
        type: 'message',
        channel: '[REDACTED]',
        ts: '[REDACTED]'
      },
      reaction: '+1',
      event_ts: '[REDACTED]'
    }

Note the `ts` property in the `type` object.

It's up to the app to look at the item property, check the type and act accordingly. Script used to test this:

    var S = require('slack-client');
    
    var slackOptions = {
      token: "YourTokenHere",
      autoReconnect: true,
      autoMark: true
    }
    
    var Cl = new S(slackOptions.token, slackOptions.autoReconnect, slackOptions.autoMark);
    
    Cl.login();
    Cl.on('open', function() {
    
    });
    
    Cl.on('reaction_added', function(reaction) {
      console.log(reaction);
    });
    Cl.on('reaction_removed', function(reaction) {
      console.log(reaction);
    });

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -333,6 +333,16 @@ class Client extends EventEmitter
    onStarRemoved: (data) ->
      @emit 'star_removed', data
 
+   #
+   # Reactions handler callbacks and dispatches
+   #
+
+   onReactionAdded: (data) ->
+     @emit 'reaction_added', data
+
+   onReactionRemoved: (data) ->
+     @emit 'reaction_removed', data
+
   #
   # Message handler callback and dispatch
   #
@@ -492,6 +502,12 @@ class Client extends EventEmitter
       when 'star_removed'
           @emit 'star_removed', message
 
+      when 'reaction_added'
+          @emit 'reaction_added', message
+      
+      when 'reaction_removed'
+          @emit 'reaction_removed', message
+
       else
         if message.reply_to
           if message.type == 'pong'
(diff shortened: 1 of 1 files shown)


vanm:
Nice. Unrelated to your code, but something to keep in mind for the JSON examples is that the structure for reaction.item is totally different for `file` and `file_comment` types.  Specifically, you can't rely on `item.channel` or `item.ts` - in my experience, no channel information was included in `file` reaction item types

```
  "type": "file",
  "file": { ... },
    "reactions": [
       {
         "name": "thumbsup",
         "count": 1,
         "users": [ "U1" ]
       }
    ]

  "type": "file_comment",
  "file": { ... },
  "comment": { ... },
  "reactions": [
    {
      "name": "facepalm",
      "count": 1034,
      "users": [ "U1", "U2", "U3", "U4", "U5" ]
    }
  ]
```

https://api.slack.com/methods/reactions.list

sillygwailo:
That's right, the RTM API doesn't return a channel ID for the file type, though it will for the `message` type, and that's something the RTM API does. Your app would have to use the API https://api.slack.com/methods/files.info to get information about the file based on its the `item.file` property (which is the ID of the file), from which you'd get channel information.

sillygwailo:
#59 might be the pull request to watch instead of this one. That one incorporates the API requests to add/remove reactions as well as listen for them.

l12s:

That release now passes through all messages from the RTM API by default as well as providing [handlers for reactions](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/data-store/message-handlers/reactions.js) against objects in the in-memory data store. Hopefully that gives you what you need, but if not, please raise an issue or create a new PR.

PR #60, opened 2015-07-17
Add createChannel method

mikebryant:


l12s:

It also moves the web API into a separate client, with individual function calls for addressing each of the API endpoints. That includes a [`channels.create`](https://github.com/slackhq/node-slack-client/blob/master/lib/clients/web/facets/channels.js#L53) method

PR #64, opened 2015-08-07
Added optional callback to channel.postMessage

pveyes:
`channel.postMessage` is missing optional callback because it ends up hitting _apiCall. I just realized this after having race condition of multiple `postMessage` call.

### DIFF
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -86,7 +86,7 @@ class Channel
     m = new Message @_client, {text: text}
     @sendMessage m
 
-  postMessage: (data) ->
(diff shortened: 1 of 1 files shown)


l12s:
Ok, first off, I apologize for how long it's taken us to address this PR.

It also refactors the way that the web API is handled, providing individual function calls for each API endpoint. In the case of the [`chat.postMessage`](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/clients/web/facets/chat.js) call it also adds in an optional callback.

Issue #53, opened 2015-06-29
WebSocket closed silently and will not reconnect


compulim:
Sometimes Slack RTM WebSocket connection close silently after running for a few days (2-3 days in our 2-3 months experiences). It does not emit any errors, unhandled exceptions, or quitting the app.

The existing "reconnect" algorithm only reconnect on auth error. Therefore, the app does not know the WS connection is dropped. It simply receive no new messages until someone restarted the app.

Currently, our remedy looks like this:
```
var slack = new (require('slack-client'))('xoxb-...', true, true);

slack
    …
    .on('close', function () {
        console.log('Connection closed, retrying');
        slack.reconnect();
    })
    …
    .login();
```

I would suggest putting this logic back into `client.coffee` when `autoReconnect` is enabled. Or building a more robust watchdog.

IgorAntun:
[+1]

Oh, good (somehow) to know I'm not the only one with this. I have the exact same issue, usually every 2~3 days the app will just disconnect without sending any notice or exceptions or anything. Thanks for your 'remedy', I'll implement it until it gets fixed for good :)

ccoenen:
possible duplicate of https://github.com/slackhq/node-slack-client/issues/5 - which has been unanswered for over a year :-(

vanm:
In my testing, this proposed reconnect on `close` can result in multiple (duplicate) open websocket connections over time.  Seems like we need to dive in on the root cause of these "silent" websocket close events or make a more robust reconnect method.  Interested what other people find after trying this approach over a few days, too - for me, it always results in duplicate websockets after a few days.

shepelt:
@vanm the same behavior was observed for me as well.
I tried slack.login() on close event as well, which resulted in the same behavior (multiple websocket created)
I believe the reconnect mechanism is not implemented properly at the moment.

vanm:
I started debugging by using a forked version of this library that logs the error code + message from websocket close events.  See PR #61 

The websocket error code I'm seeing for these silent connection closed events is `1000`, which maps to "normal" in the ws npm.  I was hoping for a more descriptive error code (for example, there are "going away" or "protocol error" codes), but I think the "normal" error code backs up how widespread this is and why it doesn't trigger any error handlers.

shepelt:
@vanm I think the root cause of this issue lies in the fact that there is no coping mechanism for connection closed events.
According to the current implementation, node-slack client is not supposed to reconnect for any kind of closed event.

[...]

PR #66, opened 2015-08-10
Fix reconnect for websocket closed without breaking pong timeout reconnect

vanm:
This branch aims to support reconnecting if a websocket is terminated / closed cleanly by Slack without disrupting the current "reconnect on Pong timeout" behavior.

Currently, this Client library does not handle the case where a websocket is closed cleanly (Code 1000) by the Server, which is why people have started to call `reconnect` in the `close` event (see #53 and #5).  I have logged many instances of Slack websockets being closed cleanly without errors or pong timeouts.  

Other people noticed this too and started calling reconnect explicitly in this Client's `close` event.  Unfortunately, calling `reconnect` in `close` can result in duplicate websockets being created if it is a part of a Pong Timeout reconnect cycle (See #53).  Because the Pong timeout flow leaves the websocket open, it's `reconnect` calls `close` on an open websocket.  If you were to have a `reconnect` call within the Slack client close event in this case, it would result in two simultaneous `login()` calls.

My hope is that this address some of the issues laid out in #5.  Warning: this will mean anyone calling `reconnect()` in this Client's `close` event needs to re-evaluate their approach.  The "reconnecting" flag should cover them, but I think the library should be entirely responsible for handling `reconnect` based on the `autoReconnect` setting (and not require consumers to handle special cases in `close`).

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -47,6 +47,7 @@ class Client extends EventEmitter
     @_apiCall 'rtm.start', {agent: 'node-slack'}, @_onLogin
 
   _onLogin: (data) =>
+    @reconnecting = false
     if data
       if not data.ok
         @emit 'error', data.error
@@ -125,7 +126,11 @@ class Client extends EventEmitter
         @emit 'error', error
 
       @ws.on 'close', =>
-        @emit 'close'
+        @emit 'close' 
+        # If autoReconnect is enabled and the client is not already reconnecting due to pong timeout
+        # Then handle the normal WS termination/close by reconnecting
+        if @autoReconnect && !@reconnecting
+          @reconnect()
         @connected = false
         @socketUrl = null
 
@@ -149,6 +154,7 @@ class Client extends EventEmitter
       return true
 
   reconnect: ->
+    @reconnecting = true
     if @_pongTimeout
       clearInterval @_pongTimeout
       @_pongTimeout = null


compulim:
This is good enough for short term. But I think in the long term, we need to rewrite the connection logic.

To make the connection robust and easy to debug, the connection logic should be driven by a singleton worker (per instance of `Client`). It will be signaled by calling `login`, `connect`, `reconnect`, and `disconnect`.

In additional to existing "ping timeout" scenario, we also need to add timeouts to `rtm.start`, establishing WebSocket connection, and wait for `{"type":"hello"}`.

vanm:
Agreed - this is definitely a short term patch.  I'm not a fan of tracking state as an instance variable between method calls and that wouldn't required if slack.close wasn't tightly bound to ws.close / or we had more expressive flow as you outlined above.

ulion:
There is other issue: each reconnect it sleep longer, I don't think things will go to better in this re-connect way for a long running app.

vanm:
@ulion can you create a new Issue for your concern re: no limit on retries?  Agreed there should be a cap but trying to keep this branch contained to a reconnect patch.

ulion:
well, maybe it's not an issue, I do remember for some case it retry sleep longer each time, but I did checked code it did reset _connAttempts after ws open, so never mind, it's not a big problem.

vanm:
OK, cool.

vanm:
Thoughts @grantmd ?

compulim:
@vanm I am working on a full rewrite of connection logic in https://github.com/compulim/node-slack-client/tree/feat-robust-connection, with the mentioned pattern above. It is still in the work and not fully compatible with current implementation yet.

Nesurion:
@vanm I used your version in our bot for a few days now and it looked promising for a few days, but then we had two open sockets. ie. the bot would react with PONG PONG on a single ping message.
```
2015-08-14 19:52:10.021 +0000 UTC ERROR Last pong is too old: 14.938
2015-08-14 19:52:10.023 +0000 UTC INFO Reconnecting in 1000ms
2015-08-14 19:52:11.023 +0000 UTC INFO Attempting reconnect
2015-08-14 19:52:11.024 +0000 UTC Connecting...
2015-08-14 19:52:11.171 +0000 UTC INFO Slack client closed, waiting for reconnect
2015-08-14 19:52:15.947 +0000 UTC INFO Logged in as *** of ********, but not yet connected
2015-08-14 19:52:27.269 +0000 UTC INFO Slack client now connected
```
I don't see whats gone wrong here, but I hope the logs help. Ping if I can provide you with any more info or logs.

vanm:
@Nesurion Hmm, I don't see duplicate "Slack client now connected" log messages in your output.  The "Slack client closed..." right after Connecting is suspect.  I can step through the code again later this week to see if anything stands out, but if anyone else has ideas, please chime in.

Hunchmun:
Is there any news on this?

vanm:
I haven't seen any suspect duplicate connections.  @Nesurion did you run into any further issues?  I haven't had time to create a test for all of this library's reconnect behavior yet.  My advice is still to run this fork in a test environment and see if it results in an improvement for you.  In my experience this works fine to reconnect and I haven't seen any dups.

labeneko:
:+1:

labeneko:
I tested https://github.com/vanm/node-slack-client/tree/ws-and-team-migration-fixes for defects and it works fine for me.
I can't reproduce the duplicate problem.

l12s:
@vanm - we chatted about the update, so I'll save the full comment, but apologies for the fact we never got to this. The plan is for 2.0.0 to be a better experience for us as maintainers and everyone who's been generous enough with their time to commit too.

vanm:
Thanks @l12s. :+1:

PR #67, opened 2015-08-12
Add count and callback to fetchHistory

mpd106:
Adding the extra `count` parameter to fetch a maximum number of results, and adding a callback with similar behaviour to that of the `client`. Also added reasonably rudimentary tests for existing and new behaviour.

### DIFF
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -103,19 +103,22 @@ class Channel
     message.channel = @id
     @_client._send(message)
 
-  fetchHistory: (latest, oldest) ->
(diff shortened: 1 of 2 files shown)


l12s:

It also moves the web API into a separate client, with individual function calls for addressing each of the API endpoints. The fetch history functions for channels, groups etc now also support all parameters the web API does, for example [`channels.history`](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/clients/web/facets/channels.js#L73)

PR #78, opened 2015-10-13
Pass additional fetch history params: inclusive and count

sdotz:
Take additional documented params for fetch history
https://api.slack.com/methods/channels.history

### DIFF
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -103,13 +103,15 @@ class Channel
     message.channel = @id
     @_client._send(message)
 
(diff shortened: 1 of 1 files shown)


l12s:

It also move the web API into a separate client, with individual function calls for addressing each of the API endpoints. In the case of [`channels.history`](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/clients/web/facets/channels.js#L73) that function also supports passing in the inclusive and counts params.

PR #79, opened 2015-10-14
Ability to query users by email id

ybv:
I could see this being useful in a lot of external bot integrations.


ybv:
@l12s / @vanm ping!

PR #85, opened 2015-11-06
Allows for custom tokens in place of the bot's

justin97530:
This is useful in methods such as "chat.delete" where the user may require a token with administrative access.

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -537,7 +537,7 @@ class Client extends EventEmitter
       return message
 
   _apiCall: (method, params, callback) ->
-    params['token'] = @token
+    if typeof params.token != "string" then params['token'] = @token
 
     post_data = querystring.stringify(params)
 


justin97530:
Hey, could you guys consider this pull request?

kesne:
It might be advantageous to make a PR against the `2.0.0-beta` branch.

l12s:
So, I'm not super familiar with the way that the new bot scopes work, so I'm going to ask one of the folks on our platform team to answer this PR.

My initial reaction is that there should be a safer way to do this than requiring users to give you tokens against their accounts.

DEGoodmanWilson:
I'm going to close this, because new features like this should be focused on the 3.x branches. However, this is a good and interesting point you raise that deserves further attention. Would you be willing to open an issue for discussing it further? (I am _very_ familiar with bot tokens, and how they are issued along side a more standard bearer token.)

justin97530:
Sure, here it is: https://github.com/slackhq/node-slack-sdk/issues/237

DEGoodmanWilson:
🙇

PR #90, opened 2015-12-08
add check to avoir client crash when archiving

mvaragnat:
When archiving a channel in which the bot is present, channel._onMark refers to a null object and crash. This bugfix adds a check to prevent it

### DIFF
--- a/src/channel.coffee
+++ b/src/channel.coffee
@@ -133,7 +133,8 @@ class Channel
     @_client._apiCall method, params, @_onMark
 
   _onMark: (data) =>
-    @_client.logger.debug data
+    if @is_archived then return null
+    else @_client.logger.debug data
     # TODO: Update @unread_count based on ts
 
   leave: ->


l12s:
What was the actual crash that happened here? I'm pretty sure that `>= 2.0.0` won't have it anymore, but it would be great to know so I can check.

mvaragnat:
The crash can be reproduced by inviting the bot to a public channel, then archiving the channel. This causes a exception, apparently _client.logger is undefined in that case

l12s:
Thanks for the update. The new version decouples model objects from the clients, so this shouldn't be possible in the latest.

Given that, I'm going to archive this PR as I'd suggest the best way to address this is to update to `>= 2.0.0`. Apologies that it took us so long to respond to this PR originally.

mvaragnat:
No problem. Thanks !

Is update to 2.0.0 from the v1 client breaking ?

l12s:
Yes - essentially a ground up rewrite to get it to a place where it's easier to test, maintain and develop. We're going to try and follow semver pretty tightly going forwards, too, so you can follow the major / minor / patch bumps for details on this too.

https://github.com/slackhq/node-slack-client/blob/master/CHANGELOG.md v2.0.0 entry has a TLDR on what's changed, but please open issues if you run into anything that's unclear and I'll either fix or add an item to the "Migrating ..." section in the README

Issue #33, opened 2015-03-11
Usage behind proxy


nilsga:
I'm trying to use this library behind a corporate proxy. How can this be configured, if possible?

evansolomon:
I have no specific experience but would guess it works like any other requests from behind the proxy. One difference may be that the client uses a web socket, not sure if that's the sort of thing that has to be explicitly whitelisted.

Are you having trouble with it?  Any specific error messages?  Can you connect to other web socket servers (e.g. https://www.websocket.org/echo.html)?

nilsga:
I have been investigating this some more, and I have come to the following conclusion. To be able to use this library through a proxy, it needs to:

* Use the `http` module instead of `https` module, and connect to the API-server through a proxy, as described here: http://stackoverflow.com/questions/16459751/calling-a-https-web-service-in-node-js-behind-proxy
* Somehow achieve the same with the `ws` module, which also uses `https` module instead of `http` module when using the `wss://` protocol.

evansolomon:
In the StackOverflow example, they're making a request to an entirely different host.  Instead of https://x.com, they're requesting http://proxy:8080/x.com.  So I assume anything is going to require some configuration by the user since the library obviously can't know some other address to use.

From a quick look at the `Client` class, it seems like two things might be a good start.

1. Changing the `host` property of a client instance
2. Allowing some way to prefix/customize the `socketUrl`

nilsga:
The "entirely different host" is the proxy server, and the format of the request is apparently how a request is proxied through a http proxy at the http level. So the supporting http/https libraries used by this module, and the websocket module does not transparently support proxy configuration. That would have to be done by this module, or depend on modules that can handle it, I suppose.

evansolomon:
> The "entirely different host" is the proxy server

Yea, I got that.  Sorry if my comment about it was unclear.

As someone who never uses this sort of proxy, I don't really know enough to say exactly what the best way to do that is (my 2 steps above were a very rough guess of where one might start).  A PR for this feature would be cool :)

nilsga:
Sorry, didn't mean sound rude. I just wanted to clarify the usage of proxy through the node http/https module. I don't think it's a quick solution to this either. It's more of a general NodeJS issue I guess.

jtatum:
The requests library has a bunch of code for proxy support and might be a good place to find inspiration. It's a little surprising the functionality isn't built into node. https://github.com/request/request

evansolomon:
[...]

PR #91, opened 2015-12-08
Add support for https proxy

GrandPubba:
Related to issue #33 this patch enables proxy support if a proxy URL is specified in the constructor to the Client class.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 *.js
+.idea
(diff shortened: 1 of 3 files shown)


l12s:
Have you tested establishing an RTM via the websocket with this?

GrandPubba:
No thanks for pointing that out.  I'll update the pull request when I have the change made and tested.

GrandPubba:
I have added support for configuring the WebSocket to proxy.  I've tested it locally and it is working.

GrandPubba:
Excellent thank you.  How do we go about getting a release to npmjs.org so I can finish the required patch for hubot-slack?

PR #94, opened 2015-12-22
Including teamDomainChange and teamRename events

brunosiqueira:


l12s:
Do you need to reconnect in the case of a team domain change or rename?

Other than that, the new client should provide the behavior you're looking for as it'll pass the event through.

brunosiqueira:
After team domain,name changed there's no need to reconnect.
Thank you!

Bruno Siqueira

2016-02-16 0:02 GMT+00:00 Leah Jones <notifications@github.com>:

l12s:
Ok, great. Closing this PR.

PR #95, opened 2015-12-30
add addReaction method to message

nguyenvanduocit:


l12s:

That refactor moves the web API interface out to its own client, with separate functions for each of the API endpoints. That includes an endpoint for [`reactions.add`](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/clients/web/facets/reactions.js#L31) which should allow you to do what you're looking for with this PR.

PR #59, opened 2015-07-11
Reactions to messages

hildjj:
React and unreact to messages, list the reactions for a message.
TODO: list reactions on a User.

As discussed in feedback+id525916.

PR #96, opened 2016-01-03
Pass through reaction_added messages

mbland:
Just sends `reaction_added` messages straight through without any special treatment. Used to implement behavior needed by 18F/hubot-slack-github-issues.

I understand that 2.0 is on the way soon, and #59 is also still open, so no hard feelings if this gets closed without merging.

cc: @afeld @ccostino

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -500,6 +500,9 @@ class Client extends EventEmitter
         @reconnect()
         @emit 'team_migration_started', message
 
+      when 'reaction_added'
+        @emit 'reaction_added', new Message(@, message)
+
       else
         if message.reply_to
           if message.type == 'pong'
--- a/test/client.coffee
+++ b/test/client.coffee
@@ -4,6 +4,7 @@
 should = require 'should'
 sinon = require 'sinon'
 Client = require '../src/client'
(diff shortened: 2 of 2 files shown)


l12s:

That release now passes through all messages from the [RTM API by default](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/clients/rtm/client.js#L356) as well as providing [handlers for reactions](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/data-store/message-handlers/reactions.js) against objects in the in-memory data store. Hopefully that gives you what you need, but if not, please raise an issue or create a new PR.

mbland:
Thanks, @l12s! Any idea when this upgraded client will get integrated into slackhq/hubot-slack?

l12s:
I've added an issue to track this: https://github.com/slackhq/node-slack-client/issues/141

TLDR is I'd like the `>= 2.0.x` release to be in use for at least 2-4 weeks to shake out any bugs and then I'll talk to the folks who are running other OSS projects (botkit, hubot etc.) about what a transition plan should look like.

PR #101, opened 2016-01-07
Update ws to 1.0.1 due to security vulnerability

timmyc:
Quick dependency version change due to a recent security advisory in `ws` https://nodesecurity.io/advisories/67

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "coffee-script": "~1.9.3",
-    "ws": "0.8.1",
+    "ws": "1.0.1",
     "log": "1.4.0",
     "https-proxy-agent": "1.0.0"
   },


l12s:
Do you know if this affects clients that aren't running ws server-side? 1.0.1 only supports node >0.12 and above, whereas the client is (currently) specced for >0.10, which gives me a little hesitation

l12s:

I meant to include the jump to `ws >= 1.0.1` as part of the 2.0.0 release, but it slipped my mind, so I'll leave this PR open as a prompt to do so. The 2.0.0 release targets `node >= 0.12`, so my concern above isn't an issue anymore.

l12s:
https://github.com/slackhq/node-slack-client/commit/88b504e73f25c3254e820488a769beaea0f2f320 fixes this

PR #112, opened 2016-01-20
Enables the Slack V3 API to linkify link content

stijndevoogd:
This will enable the Slack v3 API to linkify any @username and #channel links in messages sent by hubot. 

Usage: "@#{username}"

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -547,7 +547,7 @@ class Client extends EventEmitter
     options =
       hostname: @host,
       method: 'POST',
-      path: '/api/' + method,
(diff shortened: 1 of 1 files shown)


l12s:

That refactors the web API out to a separate client, with function calls for each of the methods it supports. Those methods also have the full set of options you can pass [documented in the docstring](https://github.com/slackhq/node-slack-client/blob/03743068caa535f15a636da164953f402d1899c5/lib/clients/web/facets/chat.js#L46). In the case of `link_names` you can pass it in to both the `postMessage` and `update` functions to get the behavior you want.

PR #116, opened 2016-01-24
Fixed dm.list method

xomyaq:
_Moved PR._
`BaseAPIClient.prototype.makeAPICall` `optData` param actually not optional. When you call it with two parameters, e.g. via `dm.list`, optional callback becomes second parameter, that causes error in `getData` and so on.


l12s:
Fixed by updating the makeAPICall function to correctly handle opt params here: https://github.com/slackhq/node-slack-client/commit/8608ee3844233225d4c123ee5ff09619844d23e2

PR #121, opened 2016-01-25
2.0.0: don't erase members when you rename a channel

acemtp:
We only init members with an empty array if members is not existing.

BTW, I don't understand why you init members in _setProperties. It should have a `members = []` in the ctor of ChannelGroup.


l12s:
I can't remember why I had this there. I think it was back when DMs were using the same object, so there could be cases where it would be empty.

I've pushed a separate PR that ditches that logic, so closing this one.

PR #122, opened 2016-01-25
2.0.0: removed toString()

acemtp:
JSON.stringify call toString() on the object if the function exists so when we try to json a model, it displays an empty string.

So I removed toString() to make JSON.stringify working


l12s:
Sounds good. I'll be updating the model logic to actually get JSON representations working properly at some point soon.

PR #124, opened 2016-01-28
2.0.0: handleNewOrUpdatedUser don't update

peernohell:
If `handleNewOrUpdatedUser` is called with an existing user we must update the user and not create a new one.
Without this patch if a user change his name dataStore loses the "presence" information (presence is not provide by user_change event).

### DIFF
--- a/lib/data-store/message-handlers/helpers.js
+++ b/lib/data-store/message-handlers/helpers.js
@@ -15,8 +15,13 @@ var noopMessage = function noopMessage(dataStore, message) {
  * {@link https://api.slack.com/events/user_change|user_change}
  */
 var handleNewOrUpdatedUser = function handleNewOrUpdatedUser(dataStore, message) {
-  var user = new models.User(message.user);
-  dataStore.setUser(user);
+  var user = dataStore.getUserById(message.user.id);
+  if(user) {
+    user.update(message.user);
+  } else {
+    user = new models.User(message.user);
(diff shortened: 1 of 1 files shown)


lib/data-store/message-handlers/helpers.js
```diff
@@ -15,8 +15,13 @@ var noopMessage = function noopMessage(dataStore, message) {
  * {@link https://api.slack.com/events/user_change|user_change}
  */
 var handleNewOrUpdatedUser = function handleNewOrUpdatedUser(dataStore, message) {
-  var user = new models.User(message.user);
-  dataStore.setUser(user);
+  var user = dataStore.getUserById(message.user.id);
+  if(user) {
```
l12s: Nitpick, but `if (user) {`

lib/data-store/message-handlers/helpers.js
```diff
@@ -15,8 +15,13 @@ var noopMessage = function noopMessage(dataStore, message) {
...
+    user.update(message.user);
+  } else {
+    user = new models.User(message.user);
+    dataStore.setUser(user);
```
l12s: Can you move the set outside the if block so it executes in either case? I'm going to be updating the way data stores function at some point so that a redis / memcache store could also be used.

peernohell:
it's fixed.
thanks for your comments it's explain to me we you keep using setters on store when I thought update was enough.

PR #127, opened 2016-02-03
Update example-web-client.js

misterdai:
Callback argument was `channels` but data being returned is team information.
Altered to avoid confusion where someone might follow the example and expect channel information.


l12s:
Sorry, I didn't notice this PR! Totally my bad - given the somewhat painful merge / rebase given I've now brought `2.0.0` to `master` I'm going to create a separate commit with this change in - thanks for catching it!

PR #133, opened 2016-02-08
pass data.error on CLIENT_EVENTS.FAILED_AUTHENTICATION to allow detecting account_inactive response

larsklevan:
If a bad token is passed to the `RtmClient` Slack will respond with something like the following:

```json
{
  "ok": false,
  "error": "account_inactive"
}
```

Right now the Slack error is not passed back to the `failed_auth` listener. With this change `failed_auth` will be called with`err` of `"account_inactive"`, `"invalid_auth"`, etc if Slack responds with those errors.

### DIFF
--- a/lib/clients/rtm/client.js
+++ b/lib/clients/rtm/client.js
@@ -218,8 +218,9 @@ RTMClient.prototype._onStart = function _onStart(err, data) {
   this._reconnecting = false;
 
   if (err || !data.url) {
-    this.logger.log('verbose', 'failed to connect to the RTM API: ' + err);
-    this.emit(CLIENT_EVENTS.FAILED_AUTHENTICATION, err);
+    var error = err || data.error;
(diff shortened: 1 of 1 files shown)


l12s:
Thanks, this is actually pretty badly written by me. I'm going to update this so that it throws client events that make more sense. I'll update that so you get the actual Slack error passed back through when appropriate.

Will link to that commit once I've written it.

l12s:
See https://github.com/slackhq/node-slack-client/commit/65228f30662b964abae6c87ca9101e64ea8c0c11

l12s:
`2.0.0-beta.8` has the fix for this + the other issue you reported to me directly. Let me know if that covers what you need and I'll close this PR.

Thanks,

Leah

larsklevan:
https://github.com/slackhq/node-slack-client/commit/65228f30662b964abae6c87ca9101e64ea8c0c11 addresses my issue. Thanks for the quick fix!

PR #135, opened 2016-02-10
sync is_member on a channel_left event

larsklevan:
When a `channel_left` event is received for a channel set `is_member` to `false`.

This is significant for a bot client so they respond correctly if they are `/kick`ed from a channel.

### DIFF
--- a/lib/data-store/message-handlers/base-channel.js
+++ b/lib/data-store/message-handlers/base-channel.js
@@ -70,10 +70,10 @@ var handleRename = function handleRename(dataStore, message) {
 var handleLeave = function handleLeave(activeUserId, activeTeamId, dataStore, message) {
   var index;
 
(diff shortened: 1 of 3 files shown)


lib/data-store/message-handlers/base-channel.js
```diff
@@ -70,10 +70,10 @@ var handleRename = function handleRename(dataStore, message) {
 var handleLeave = function handleLeave(activeUserId, activeTeamId, dataStore, message) {
   var index;
 
-  var baseChanel = dataStore.getChannelGroupOrDMById(message.channel);
-  if (baseChanel) {
-    index = baseChanel.members.indexOf(activeUserId);
-    baseChanel.members.splice(index, 1);
+  var baseChannel = dataStore.getChannelGroupOrDMById(message.channel);
```
larsklevan: just fixing a typo here

PR #150, opened 2016-02-29
Changed dm faucet name to im to match with Slack API method documentation

spideynr:
Took me a while to find why `webClient.im.open` didn't work, after working with the Slack API docs.

### DIFF
--- a/lib/clients/web/facets/dm.js
+++ b/lib/clients/web/facets/dm.js
@@ -11,7 +11,7 @@
  */
 
 function DmFacet(makeAPICall) {
-  this.name = 'dm';
+  this.name = 'im';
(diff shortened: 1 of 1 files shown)


spideynr:
I tried renaming all functions to IM from DM as well but figured it would maybe cause some mayhem in development and when i reverted everything it came back to just changing the `this.name` of the faucet

l12s:
Hey, sorry for the slow reply.

This is actually a deliberate decision - DM is interchangeable with IM in the documentation, and DM is the preferred term. So, I decided to use that instead of IM for the client. I get that the inconsistency is a bit confusing though and it's something we'll try and address with some better documentation.

For now though, I'm going to close this PR.

PR #158, opened 2016-03-04
Reaction listeners

anskaal:
Thank you for creating this client! :+1: 
Here are a few lines of code for enabling support for reaction listeners :)

### DIFF
--- a/examples/example-rtm-client.js
+++ b/examples/example-rtm-client.js
@@ -14,3 +14,11 @@ rtm.start();
 rtm.on('message', function handleRtmMessage(message) {
   console.log(message);
 });
+
+rtm.on('reaction_added', function handleRtmReactionAdded(reaction) {
+  console.log(reaction);
+});
+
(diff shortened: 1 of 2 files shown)


l12s:
Hi there - the current version of the client should emit `reaction_xxx` messages.

We emit an event https://github.com/slackhq/node-slack-client/blob/master/lib/clients/rtm/client.js#L424 for every message type (bar the `RTM_CLIENT_INTERNAL_EVENT_TYPES` ones). So, you should be able to do:

```
var RTM_EVENTS = require('@slack/client').RTM_EVENTS;
rtm.on(RTM_EVENTS.REACTION_ADDED, function() {
  ...
});
```

Our docs really aren't great right now though, so if you wanted to add the updates you have to the `example-rtm-client.js` file, that'd be great.

anskaal:
You are right @l12s :) Thanks

l12s:
Great - thanks!

PR #164, opened 2016-03-07
Updates the RTM message send functionality.

### LINKED ISSUE #163: Returning a reference of a sent message
So in the pre-v2 days you could do something like this:

```
let sentMessage = channel.send(`Hello!`);

setTimeout(() => {
  sentMessage.updateMessage('I will be deleted in 2 seconds.');
}, 1000);

setTimeout(() => {
  sentMessage.deleteMessage();
}, 3000);
```

Nowadays the callback for `rtm.sendMessage` only returns an error object and no other arguments. 

I had a look through the source and found that although you can update messages via `channel.updateMessage`, you need a reference to the message you sent, in particular its timestamp value. Plus it seems that you can only retrieve messages by using `channel.getMessageByTs`, again requiring that pesky timestamp.

Is there a way in the current system to emulate the above behavior? If not is it perhaps an idea to add a reference to the sent message in `rtm.sendMessage`'s optional callback like so:

```
rtm.sendMessage(`Hello!`, channel.id, (err, msg) => {
  channel.updateMessage('Updated!', msg.id);
});
```

Happy to implement this myself is someone can point me in the right direction.

l12s:
This:
- adds promise support to the RTM send / sendMessage functions
- updates the way message responses are handled, so that the message success callback is triggered when the API returns a message object with a `reply_to` matching the id of the sent message

Fixes #163

### DIFF
--- a/lib/clients/rtm/client.js
+++ b/lib/clients/rtm/client.js
@@ -3,11 +3,15 @@
  * See [the RTM client events](../events/client) for details of the client event lifecycle.
  */
 
+var Promise = require('bluebird');
 var bind = require('lodash').bind;
 var cloneDeep = require('lodash').cloneDeep;
 var contains = require('lodash').contains;
+var hasKey = require('lodash').has;
 var inherits = require('inherits');
+var isFunction = require('lodash').isFunction;
 var isUndefined = require('lodash').isUndefined;
+var noop = require('lodash').noop;
 
 var RTM_API_EVENTS = require('../events/rtm').EVENTS;
 var RTM_CLIENT_INTERNAL_EVENT_TYPES = [
@@ -23,6 +27,7 @@ var UNRECOVERABLE_RTM_START_ERRS = [
 var CLIENT_EVENTS = require('../events/client').RTM;
 var BaseAPIClient = require('../client');
 var RtmFacet = require('../web/facets').RtmFacet;
+var SlackRTMSendError = require('../errors').SlackRTMSendError;
 var makeMessageEventWithSubtype = require('../events/utils').makeMessageEventWithSubtype;
 var wsSocketFn = require('../transports/ws');
 
@@ -65,6 +70,20 @@ function RTMClient(token, opts) {
    */
   this.ws = undefined;
 
+  /**
+   *
+   * @type {number}
+   * @private
+   */
+  this._messageId = 1;
+
+  /**
+   *
+   * @type {{}}
+   * @private
+   */
+  this._msgResponseHandlers = {};
+
   if (!isUndefined(clientOpts.dataStore)) {
     this.registerDataStore(clientOpts.dataStore);
   }
@@ -104,14 +123,6 @@ RTMClient.prototype.activeUserId = undefined;
  * @type {string}
  */
 RTMClient.prototype.activeTeamId = undefined;
-
-
-/**
- * Maps message id to message object for messages sent to but not replied to by the remote server.
- * @type {Object}
- * @private
- */
-RTMClient.prototype._pendingMessages = {};
 
 
 /**
@@ -198,9 +209,10 @@ RTMClient.prototype.login = function login(opts) {
 
 /**
  * Generates the next message id to use.
+ *
+ * NOTE: This id must be unique per RTM connection.
  */
 RTMClient.prototype.nextMessageId = function nextMessageId() {
-  this._messageId = this._messageId || 1;
   return this._messageId++;
 };
 
@@ -322,7 +334,7 @@ RTMClient.prototype._pingServer = function _pingServer() {
   var pongInterval;
 
   if (this.connected) {
-    this.send({ type: 'ping' });
+    this.send({ type: 'ping' }, noop);
 
     // If the last pong was more than MAX_PONG_INTERVAL, force a reconnect
     pongInterval = Date.now() - this._lastPong;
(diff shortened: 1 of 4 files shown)


lib/clients/rtm/client.js
```diff
@@ -505,33 +515,106 @@ RTMClient.prototype.sendMessage = function sendMessage(text, channelId, optCb) {
...
     });
   } else {
-    err = 'ws not connected, unable to send message';
-    this.logger('debug', err);
-    if (!isUndefined(optCb)) {
-      optCb(new Error(err));
+    ret = new Promise(function wsSendPromiseInner(fulfill, reject) {
+      _this._registerMsgHandler(msgId, { fulfill: fulfill, reject: reject });
+      _this.ws.send(jsonMessage, undefined, function handleWsMsgResponseCb(wsRespErr) {
+        if (!isUndefined(wsRespErr)) {
```
PaulAsjes: I'm a little confused here, surely we'd want to call the callback from `sendMessage` with the response if there's no ws error? 

I tested this out and found that the callback was not invoked when a message is sent successfully.

l12s: This is the callback for the websocket's send function. The reason we don't call the callback for success here is that we're treating message success as being when the server responds to us with a message object containing a `reply_to` key and a `ts` key.

l12s: Can you give me an example of the calling code where you're not seeing the success cb fire? It was working for me in some simple tests, so not sure what's happening there...

PaulAsjes: Okay I see where I got turned around, I thought this was where we get the response object from the server containing the `reply_to` and `ts` keys.

Here's the code that I'm running: 

```
slack.sendMessage('Hello!', channel.id, (err, msg) => {
    console.log('ret:', err, msg);
  });
```

That log never fires despite the bot posting the message on Slack.

l12s: What's the channel id value there? I'm wondering how the response handler will work if it's given information that couldn't work, e.g. a channel id that doesn't exist.

l12s: Ok, weirdly, it seems like it will default to posting to the general channel if you pass in an invalid or non-existent channel id, so my thought there doesn't seem to make sense.

lib/clients/rtm/client.js
```diff
@@ -403,13 +415,15 @@ RTMClient.prototype._handleWsMessageViaEventHandler = function _handleWsMessageV
   messageType, message) {
   var replyTo;
 
+  if (!isUndefined(this.dataStore)) {
+    this.dataStore.handleRtmMessage(this.activeUserId, this.activeTeamId, messageType, message);
+  }
+
```
PaulAsjes: I might have found a clue as to what's going on. 

I added a log here:
`console.log('messageType:', messageType, message);`

And got this:
`messageType: undefined { ok: true, reply_to: 2, ts: '1457315835.000510', text: 'Hello!' }`

The message is the one sent by the bot, but that undefined value is intriguing, it should be of the type `RTM_API_EVENTS.MESSAGE` I believe. I also logged the `msgId` value in `_send`, and it corresponds to the one in the returned object above.

l12s: Hmm, thanks. I think it may be a misunderstanding on my part about how the `reply_to` ack is handled. Looking into that now.

l12s: Yep, that'd be it. Thanks for catching this! I'm not sure where the message with message type of `message` I tested against came from...

PaulAsjes: Yup, the `message` object here is missing a `type` parameter oddly enough.

Everesty: So the api docs were right, though I find that a bad thing in this case. I feel like that'd be a cause of a high number of issues for anyone trying to make their own client to use with the RTM api (most people would probably throw an error if they got a message without a type that wasn't an error). That, or I'm crazy.

I have math to do so I won't be able to test this for a solid hour, but I'll be sure to throw my best worst-cases at it.

PaulAsjes: @Everesty is right, this is technically Working As Intended. A workaround could be to check [here](https://github.com/slackhq/node-slack-client/blob/update_rtm_send/lib/clients/rtm/client.js#L380) whether `message` is undefined and if `ok` is true. If so, use the timestamp provided to grab the message via `getMessageByTs` (how you would gain access to that particular method I don't know) and use that instead.

l12s: I can work around the response not having a type. The thing that's confusing me right now is that I get a message with type `message` back from the RTM API with the contents of the message that I pushed, but only on reconnect.

I'm going to chat to the folks who work on our websocket code tomorrow to try and understand what the correct behavior here should be.

l12s:
Ok, I have a question out to the folks who work on our websocket servers to clarify what the expected behavior here should be.

Depending on how this should work, I have a couple thoughts on how to make the behavior work cleanly with this PR. I'll likely not be able to write a fix until mid-end of this week though, so will likely be somewhat delayed getting this out as a `2.2.0`

l12s:
I actually got a super quick reply, so I've added another commit that contains the logic for how I'd propose to address this.

PaulAsjes:
Did a quick test this morning and it seems to be working as expected. Although I did notice that when restarting my node process the last `RTM_EVENTS.MESSAGE` event would immediately fire, with the last known message as the payload.

I'll have another look this evening to see if I can figure out why the above is happening.

l12s:
That's actually an undocumented feature of the API. It returns the last known message when the connection first opens. That message will have a `reply_to` id set, although the id may not make any sense in the context of the local client, as it could be from a prior node process or from the web client if the user also has that open.

So, couple of options:
- the client can simply ignore events with type `message` and a `reply_to` key (the simplest option)
- the client can strip the `reply_to` and treat the last message as a regular message
- the client can check to see if it has a message handler registered for that `reply_to` id and then fire it if it does (ultimately the correct solution)

The other issue that comes up with this is that the client needs to (and doesn't currently) deal with message reply callbacks when the connection is dropped. I have some thoughts on how to do that gracefully, but they're kind of incomplete right now.

PaulAsjes:
It seems a little odd to me that the API would return the last known message on connection open, as it was sent before the local client started and thus can probably be ignored. However it's certainly not a huge issue and it might be useful in certain scenarios.

I think a combination of options 2 and 3 is probably best. The `reply_to` key is ultimately useless in the local client so should be omitted. All this is definitely out of scope for this particular PR so I say :+1: and thanks for turning this around so quickly!

l12s:
I believe that it's a convenience for cases where the client loses connection while a message is pending.

I'm going to implement option 2 and leave a TODO + create an issue to figure out the logic around managing handlers when the client disconnects / reconnects.

PR #165, opened 2016-03-07
Prevent pong crashing

mvaragnat:
Thanks Leah for this awesome client. 

I ran into an issue when starting several hundreds of teams at once. Sometimes, in a difficult-to-reproduce way, the pong from the RTM connection happened while the Websocket was not yet opened or was in the process of closing (if the token had been revoked). I am not 100% sure of the diagnostic, the error log was

/app/node_modules/ws/lib/WebSocket.js:180 
throw new Error('not opened'); 
^ 
Error: not opened 
at WebSocket.pong (/app/node_modules/ws/lib/WebSocket.js:180:11) 
at RTMClient.handleWsPing (/app/node_modules/@slack/client/lib/clients/rtm/client.js:465:13)

The proposed patch does NOT solve the underlying issue (which I don't understand) but simply sets the flag dontFailWhenClose on the WebSocket to avoid the app to crash


l12s:
Thanks, I'm likely going to be a bit slow merging this as I'm slammed for time to work on the client this week. Will try and pick this up and have a think about what might be happening with the pong connections mid-week.

l12s:
So, I spent some time thinking about this and I don't think that function actually makes any sense anymore.

It was something I translated directly from the old coffescript client, which may have been using an old enough version of `ws` when it started that the underlying ping / pong control frame back and forth wasn't automatic. At this point, it seems like all the standard node websocket libraries deal with that automatically, e.g. https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L830, so I can't figure out what value having a handler for the underlying websocket `ping` event gives us.

So, I'm going to remove that event handler for now. Note that the logic for handling the Ping / Pong exchange at the RTM level, as discussed in https://api.slack.com/rtm will still be there.

https://github.com/slackhq/node-slack-client/commit/7f658e9971fd63b565c0aacc7dd32c7c1778fc16 does that work. I'll push out a 2.3.0-beta.1 release shortly. Would be great if you could try that out and see if it fixes things for you.

l12s:
Ok, you can `npm install @slack/client@2.3.0-beta.2` to get the version with that function removed.

I'm going to close this PR - could you update https://github.com/slackhq/node-slack-client/issues/176 if you have issues with that beta release?

PR #195, opened 2016-04-14
Updates the API facets to add new endpoints, fix a couple of bad args…

l12s:
… in others and create new facets.


l12s:
@DEGoodmanWilson - let's chat through this tomorrow / Friday?

PR #199, opened 2016-04-19
add an rtm client example with datastore usage

ggauravr:
It took me a while to figure out the usage of datastore. So adding a simple example on how to use it with rtm client that would help others get started!


l12s:
Thanks for contributing!

ggauravr:
Awesome. Thanks!

PR #205, opened 2016-05-01
Add missing facets

ekmartin:
Seems like some of the facets added in 3.0.0 are missing from the web client itself.
Also adds a test that checks that all facets have been registered to the client.


l12s:
Thanks for catching that!

ekmartin:
No problem, quick response!

PR #206, opened 2016-05-03
New updateMessage feature

PaulAsjes:
Version 1.x used to have the [`updateMessage` method](https://github.com/slackhq/node-slack-client/blob/1.x/src/message.coffee#L60) which doesn't seem to have made the transition to later versions. I included the Chat facet in the RTM client to fix that.

I'm unsure how to test that the `chat.update` API call was made here without a valid token (`sendMessage` tests seem to have the same issue). Any thoughts?


lib/clients/rtm/client.js
```diff
@@ -617,6 +628,35 @@ RTMClient.prototype.sendMessage = function sendMessage(text, channelId, optCb) {
...
+ * @param message {object} message The message object to be updated,
+ * see /lib/clients/web/facets/chat.js for more details
+ * @param {Function=} optCb Optional callback
+ */
+RTMClient.prototype.updateMessage = function updateMessage(message, optCb) {
+  var err;
+  var ret;
+
+  if (this.connected) {
+    ret = this._chat.update(message.ts, message.channel, message.text, message.opts, optCb);
```
DEGoodmanWilson: Could you say a little more about this design decision? Connection to the MS isn't necessary for a call to chat.update to succeed, being a web API call.

test/clients/rtm/client.js
```diff
@@ -140,6 +140,27 @@ describe('RTM API Client', function () {
...
+      var rtm = new RtmAPIClient('fake-token');
+      rtm.updateMessage({}, function (err, res) {
+        expect(err).to.not.equal(null);
+        expect(res).to.equal(null);
+        done();
+      });
+    });
+
+    it('should call a catch cb with an err when the RTM client is not connected', function (done) {
+      var rtm = new RtmAPIClient('fake-token');
```
DEGoodmanWilson: Probably the thing to do—at least what I do with the client I develop on my own—is simply require a valid token in the environment. And then provide one to a test team you've set up for the purposes of running tests.

lib/clients/rtm/client.js
PaulAsjes: I'm a little unclear on what you mean by MS, but my goal here was to expose the already existing `update` function of the [`ChatFacet`](https://github.com/slackhq/node-slack-sdk/blob/master/lib/clients/web/facets/chat.js#L116).

test/clients/rtm/client.js
PaulAsjes: The only issue with this approach is that it's very hard to test "out of the box". Ideally someone should be able to clone the repo and run tests without having to set up a team. Is it possible to generate some sort of fake test token which returns the correct payload but otherwise does nothing?

lib/clients/rtm/client.js
DEGoodmanWilson: Sorry, I forget that's internal terminology. MS = Message Server = the thing you connect to with the RTM API. My question is on the `if (this.connected)` condition—you can call `chat.update` without a live RTM connection.

test/clients/rtm/client.js
DEGoodmanWilson: Unfortunately, at this time, it is not. The only way to test this functionality is with a live team. I agree it's far from ideal :( But this might be a good feature to consider for the platform to build!

lib/clients/rtm/client.js
PaulAsjes: Ah I didn't realise! I'll remove that and have a rethink about the testing before updating this PR.

DEGoodmanWilson:
@PaulAsjes Looks like there is a merge conflict with your latest commit, can you fix?

PaulAsjes:
Phew, lot of commits that didn't merge nicely. That'll teach me to be inactive.

lib/clients/web/facets/team.js
```diff
@@ -21,8 +21,6 @@ function TeamFacet(makeAPICall) {
  * @see {@link https://api.slack.com/methods/team.accessLogs|team.accessLogs}
  *
  * @param {Object=} opts
-
-
```
DEGoodmanWilson: There are a lot of whitespace changes in this PR…would you mind adjusting that? Just to keep things clean?

DEGoodmanWilson:
👍

PR #210, opened 2016-05-11
update DmFacet to use updated makeAPIcall args

larsklevan:
For calls in `DmFacet` like `dm.open` the callback is never being called.

```javascript
SlackClient = require '@slack/client'

slack = new SlackClient.WebClient("mytoken")
slack.dm.open('USERID', function(err, response) {
  console.log('this will never be logged');
});
```

The equivalent call with `im.open` works.

Update `DmFacet` to use the updated `makeAPICall` signature from https://github.com/slackhq/node-slack-client/pull/195 with separated `requiredArgs` and `opts`.

### DIFF
--- a/lib/clients/web/facets/dm.js
+++ b/lib/clients/web/facets/dm.js
@@ -24,11 +24,11 @@ function DmFacet(makeAPICall) {
  * @param {function} optCb Optional callback, if not using promises.
  */
 DmFacet.prototype.close = function close(channel, optCb) {
-  var args = {
+  var requiredArgs = {
     channel: channel
   };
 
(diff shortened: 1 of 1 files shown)


l12s:
Oof, this is a big oversight on my part - sorry about that :(

The codegen approach I'm using shouldn't have generated the DM facet in this case - it should be a synonym for the IM facet, which is updated per your suggestion. I'm going to close this PR and remove the DM facet and update it to be a link to the IM facet instead.

l12s:
https://github.com/slackhq/node-slack-sdk/compare/fix_dm_facet?expand=1

PR #217, opened 2016-05-22
Fix retry-after header name and add fallback value

foiseworth:
When diagnosing an unrelated issue I noticed that our in-house bot was repeatedly trying requests after receiving a 429 response.

This was caused by the header ```Retry-After``` being instead ```retry-after``` and the previous code then passing NaN to setTimeout which caused it to be called immediately.

I fixed it by renaming the header and then adding a fallback value if the header is not present. I also added a log message so that it is more obvious what is happening.


l12s:
Thanks!

PR #230, opened 2016-06-05
Fix and improve consistency of datastore examples

fvgs:
Remove the `autoMark` option from the Data stores section of README.md since it no longer seems to be a valid option. Likewise, remove the `autoReconnect: true` option since it defaults to true. The user can learn about the available options from the source code where they are documented.

Improve consistency between example-rtm-client-datastore.js and README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,11 +123,7 @@ var rtm = new RtmClient(token, {
   // Sets the level of logging we require
   logLevel: 'error',
   // Initialise a data store for our client, this will load additional helper functions for the storing and retrieval of data
-  dataStore: new MemoryDataStore(),
-  // Boolean indicating whether Slack should automatically reconnect after an error response
-  autoReconnect: true,
-  // Boolean indicating whether each message should be marked as read or not after it is processed
-  autoMark: true
+  dataStore: new MemoryDataStore()
 });
 
 rtm.start();
--- a/examples/example-rtm-client-datastore.js
+++ b/examples/example-rtm-client-datastore.js
@@ -7,8 +7,8 @@ var MemoryDataStore = require('@slack/client').MemoryDataStore;
 var token = process.env.SLACK_API_TOKEN || '';
 
 var rtm = new RtmClient(token, {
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -123,11 +123,7 @@ var rtm = new RtmClient(token, {
   // Sets the level of logging we require
   logLevel: 'error',
   // Initialise a data store for our client, this will load additional helper functions for the storing and retrieval of data
-  dataStore: new MemoryDataStore(),
-  // Boolean indicating whether Slack should automatically reconnect after an error response
-  autoReconnect: true,
```
DEGoodmanWilson: Why did you remove this option from the readme? This is still a valid option.

examples/example-rtm-client-datastore.js
```diff
@@ -7,8 +7,8 @@ var MemoryDataStore = require('@slack/client').MemoryDataStore;
 var token = process.env.SLACK_API_TOKEN || '';
 
 var rtm = new RtmClient(token, {
-  logLevel: 'info', // check this out for more on logger: https://github.com/winstonjs/winston
-  dataStore: new MemoryDataStore({}) // pass a new MemoryDataStore instance to cache information
```
DEGoodmanWilson: Why these changes? This seems outside the scope of your PR's discussion (removing invalid options from documentation)

README.md
fvargas: The option is certainly valid, however it defaults to `true`. Therefore, having `autoReconnect: true` is redundant.

I could see value in demonstrating the existence of the option since this is an example. However, there are several other options not in the example which a user would need to find in the source code documentation. Thus, since the option is non-essential and redundant, I think it is most intuitive to consistently use only those options which are essential.

examples/example-rtm-client-datastore.js
fvargas: This was done to make the example consistent with the code in the readme. Consistent examples give the user a better idea of what sensible defaults to use. The purpose of the PR is also to improve consistency.

PR #233, opened 2016-06-10
Upgrade lodash

crookedneighbor:
Upgrades lodash to version 4

* Replaces instances where a function called `this` as the last argument with `bind`

    ```js
    forEach(variable, function () {}, this);
    ``` 
    with 
    ```js
    forEach(variable, bind(function () {}, this));
    ``` 

* Changed `contains` to `includes`
* Changed `zipObject` to `fromPairs`
* Changed find shorthand:

    ```js
    find(collection, 'property', value);
    ``` 
    to
    ```js
    find(collection, ['property', value]);
    ```

All tests pass after these changes and I didn't _see_ anything else that looked concerning, but it's possible I missed something. 

I ran the code in the examples directory and it worked.

### DIFF
--- a/lib/models/model.js
+++ b/lib/models/model.js
@@ -3,6 +3,7 @@
  */
 
 var forEach = require('lodash').forEach;
+var bind = require('lodash').bind;
 var has = require('lodash').has;
 var isArray = require('lodash').isArray;
 var isEmpty = require('lodash').isEmpty;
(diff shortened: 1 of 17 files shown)


lib/models/model.js
```diff
@@ -152,7 +153,7 @@ Model.prototype._setArrayProperty = function _setArrayProperty(key, val) {
 
 
 Model.prototype.toJSON = function toJSON() {
-  var objRepresentation = mapValues(this._properties, function mapModelPropertiesToJSON(val, key) {
+  var objRepresentation = mapValues(this._properties, bind(function mapModelPropsToJSON(val, key) {
```
crookedneighbor: Had to shorten this function name because eslint wasn't happening about how long it was.

johnagan:
LGTM 👍

PR #244, opened 2016-07-07
Support auto-reconnecting websocket failures

mhahn:
Support auto-reconnecting on websocket failures.

Fixes: https://github.com/slackhq/node-slack-sdk/issues/240

### DIFF
--- a/lib/clients/rtm/client.js
+++ b/lib/clients/rtm/client.js
@@ -576,6 +576,14 @@ RTMClient.prototype._handleMostRecentMsgReply = function _handleMostRecentMsgRep
 RTMClient.prototype.handleWsError = function handleWsError(err) {
   this.logger('debug', err);
   this.emit(CLIENT_EVENTS.WS_ERROR, err);
+  this.connected = false;
+  if (this.autoReconnect) {
+    if (!this._connecting) {
+      this.reconnect();
+    }
+  } else {
+    this.disconnect('websocket error with auto-reconnect false on the RTM client');
+  }
 };
 
 


DEGoodmanWilson:
I'm going to look into this PR today, as improvements to the reconnect logic are important! Before I accept this, however, I would like for you to include tests around this new functionality.

l12s:
Are we guaranteed that getting an error of the WS means that it's actually either closed or needs to be reopened?

DEGoodmanWilson:
Depending on the source of this particular error, it _might_ mean the token used to open the connection is no longer valid, that's one of the things I want to investigate.

mhahn:
I've been getting a couple of these every few days and the error is usually something along the lines of "socket hang up". i am not sure if it always means the connection is closed though, i'll check on that.

mhahn:
when the token is no longer valid, the `DISCONNECT` event is emitted with the code `account_inactive`

DEGoodmanWilson:
Good to know.

mhahn:
Some of the specific errors I've seen:

```
Error: write EPROTO 140405736589120:error:140790E5:SSL routines:ssl23_write:ssl handshake failure:../deps/openssl/openssl/ssl/s23_lib.c:177:

    at exports._errnoException (util.js:890:11)
    at WriteWrap.afterWrite (net.js:767:14)
```

```
Error: socket hang up
    at TLSSocket.onHangUp (_tls_wrap.js:1097:19)
    at TLSSocket.g (events.js:273:16)
    at emitNone (events.js:85:20)
    at TLSSocket.emit (events.js:179:7)
    at endReadableNT (_stream_readable.js:913:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9)
```

```
Error: socket hang up
    at createHangUpError (_http_client.js:250:15)
    at TLSSocket.socketOnEnd (_http_client.js:342:23)
    at emitNone (events.js:85:20)
    at TLSSocket.emit (events.js:179:7)
    at endReadableNT (_stream_readable.js:913:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9)
}
```

after reading througth the ws code, i think @l12s has a point, we're not guaranteed that the connection has been closed and it looks like most times an `error` is emitted, the `close` event is also emitted during cleanup: https://github.com/websockets/ws/blob/e591ba61022ffc7a6c0c865c8759309c2b2a35c3/lib/WebSocket.js#L712.

i have more monitoring and logging setup around these errors so i'll keep and eye on it and see if `WS_ERROR` is actually resulting in a closed connection, or if it is already being reconnected because of `WS_CLOSE` being emitted on cleanup.

DEGoodmanWilson:
Where do we stand on this, y'all? Is it looking like the current logic is more or less the right logic? I still feel that not _every_ disconnection deserves a reconnect attempt (specifically when tokens are invalidated because of uninstall)—is this the right place to be thinking about crafting that logic improvement?

aoberoi:
@mhahn ping! are there any clear next steps here? i'm not totally up to speed, but it seems like in order to do this right, we need to keep track of whether an error event corresponding to a case that we determine should cause a reconnect occurs, wait until the corresponding close event (when the websocket resource is in fact released) and then begin reconnection. so it sounds like the PR as-is shouldn't be merged.

also, i want to note that the next major version of `ws` exists. it might present a more favorable API for us to implement this specific feature, so it might be worthwhile to investigate.

aoberoi:
also, dat CLA.

CharlieHess:
Given
>Are we guaranteed that getting an error of the WS means that it's actually either closed or needs to be reopened?

I think we're in agreement that the current behavior is 👌.

PR #247, opened 2016-07-12
Changed update message documentation to ES5 syntax for consistency

PaulAsjes:
Realised that I left some ES6 code in the `updateMessage` documentation, so fixed that and added a link.


DEGoodmanWilson:
I really should add CI checks for that. Thanks for noticing.

Issue #292, opened 2016-11-30
Cannot access headers (and thus X-OAuth-Scopes)


benjie:
* [x] I've read and understood the [Contributing guidelines](./CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](./CODE_OF_CONDUCT.md). (Also this link doesn't work in GitHub issues preview.)
* [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description

When doing an API fetch, headers aren't available and thus we can't determine what OAuth scopes are available to the current access token.

### Workaround

Bypass the SDK and perform a simple HTTPS request with `?token=...`; then read the headers from this response.

```
$ curl https://slack.com/api/files.list?token=abcd -I
HTTP/1.1 200 OK
X-OAuth-Scopes: files:read, chat:write:bot
X-Accepted-OAuth-Scopes: files:read
```

#### Reproducible in:

Node.js 6.8.1
@slack/client version: 3.6.1

#### Steps to reproduce:

1. `const slack = new SlackWebClient(accessToken);`
2. `slack.team.info((err, details) => {...})`
3. No way to determine headers as described in [Working with Scopes](https://api.slack.com/docs/oauth-scopes)

### Expected result:

Response data (JSON) and headers.

### Actual result:

Response data only.

DEGoodmanWilson:
This is quite good, thanks for this! We should look into adding returning that information to the callbacks for sure. Shouldn't be hard, I think!

DEGoodmanWilson:
@benjie Would you have a look at #296 and see if this helps resolve your issue?

benjie:
Looks like it should solve it; I also like the improvements to error handling, the previous behaviour was not idiomatic. Thanks! 👍

PR #296, opened 2016-12-08
Add more error reporting information to the Web Client callback responses

### LINKED ISSUE #291: postMessage API inconsistency
#### Description
Per looking at [this example](https://slackapi.github.io/node-slack-sdk/basic_usage#posting-a-message-with-web-api) in the documentation of the Web API, I expected my callback to have been called with an `Error` as the first argument if there's an issue.

### Expected result:
The `err` callback arg to have been populated with an `Error`.

### Actual result:
The `response` callback arg is called with an "error data structure".

Although I had an error handler in place, after a little bit of trial an error, I determined that the success handler was instead being called with the following: `{ ok: false, error: 'channel_not_found' }`.

For what it's worth, I triggered the above error scenario (trying to publish to a channel that doesn't exist) to double-check that my logging was working as expected.

Specifically, I'm wrapping `webClient.chat.postMessage` in a promise and, upon success, I'm acknowledging a message on a queue as processed. However, like mentioned above, I was erroneously acknowledging because I'd expected the error argument to have been provided.

I guess my question is this -- I feel like, from an API perspective, I should be handling ***one*** scenario or the other. In other words, if the error is always going to be contained within the `response` argument as an error data structure, then the callback only needs to have one param provided. Conversely, the library could invoke the callback with an `Error` if the response isn't successful. But, as of right now, I'm having to handle **both** possibilities:

```JavaScript
    // ... other code
      webClient.chat.postMessage(channel, text, {
        as_user: true,
        attachments,
      }, (err, response) => {
        // NOTE: The following two lines specifically.
        const { error } = response;
        if (err || error) return reject(err || error);
        return resolve(response);
      });
    });
```

I guess I just more or less wanted confirm that this is as-intended from a design perspective, because it definitely caught me off-guard and sent me on a little 5-minute debugging detour! 😄

DEGoodmanWilson:
* [X] I've read and understood the [Contributing guidelines](./CONTRIBUTING.md) and have done my best effort to follow them.
* [X] I've read and agree to the [Code of Conduct](./CODE_OF_CONDUCT.md).
* [X] I've been mindful about doing atomic commits, adding documentation to my changes, not refactoring too much.
* [X] I've a descriptive title and added any useful information for the reviewer. Where appropriate, I've attached a screenshot and/or screencast (gif preferrably).
* [X] I've written tests to cover the new code and functionality included in this PR.
* [X] I've read, agree to, and signed the [Contributor License Agreement (CLA)](https://docs.google.com/a/slack-corp.com/forms/d/1q_w8rlJG_x_xJOoSUMNl7R35rkpA7N6pUkKhfHHMD9c/viewform).

#### PR Summary
Adds the scoping information from the headers to callback response, and puts API errors into the callback error

#### Related Issues
Fixes #291 and #292 

#### Test strategy
TBD

### DIFF
--- a/lib/clients/transports/call-transport.js
+++ b/lib/clients/transports/call-transport.js
@@ -102,12 +102,17 @@ var handleHttpResponse = function handleHttpResponse(body, headers, client, apiC
   }
 
   if (!jsonResponse.ok) {
-      jsonError = new Error(jsonResponse.error);
-      client.logger('error', jsonResponse.error);
+    jsonError = new Error(jsonResponse.error);
+    client.logger('error', jsonResponse.error);
   }
 
-  jsonResponse.scopes = (headers['X-OAuth-Scopes'] === undefined) ? [] : headers['X-OAuth-Scopes'].trim().split(/\s*,\s*/);
-  jsonResponse.acceptedScopes = (headers['X-Accepted-OAuth-Scopes'] === undefined) ? [] : headers['X-OAuth-Scopes'].trim().split(/\s*,\s*/);
+  jsonResponse.scopes = (headers['X-OAuth-Scopes'] === undefined) ?
+    [] :
+    headers['X-OAuth-Scopes'].trim().split(/\s*,\s*/);
+  jsonResponse.acceptedScopes = (headers['X-Accepted-OAuth-Scopes'] === undefined) ?
+    [] :
+    headers['X-OAuth-Scopes'].trim().split(/\s*,\s*/);
+
 
   try {
     apiCb(jsonError, jsonResponse);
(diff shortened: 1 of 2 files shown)


lib/clients/transports/call-transport.js
```diff
@@ -85,24 +85,37 @@ var handleHttpErr = function handleHttpErr(retryOp, apiCb, statusCode) {
...
+    jsonError = new Error(jsonResponse.error);
+    client.logger('error', jsonResponse.error);
+  }
+
+  jsonResponse.scopes = (headers['X-OAuth-Scopes'] === undefined) ?
+    [] :
+    headers['X-OAuth-Scopes'].trim().split(/\s*,\s*/);
+  jsonResponse.acceptedScopes = (headers['X-Accepted-OAuth-Scopes'] === undefined) ?
+    [] :
+    headers['X-Accepted-OAuth-Scopes'].trim().split(/\s*,\s*/);
```
benjie: Headers should be accessed in a case-insensitive way. If this is a node http.IncomingMessage then the headers will have been lowercased:

> Key-value pairs of header names and values. Header names are lower-cased. 
> -- https://nodejs.org/api/http.html#http_message_headers

DEGoodmanWilson: It's not—it is from `request`, but thank you for pointing this out! I'll look up what the best practices for handling headers passed by `request` are.

DEGoodmanWilson:
@benjie Thanks for your comments, I've updated the header keys to be lowercase.

PR #302, opened 2016-12-20
Telemetry

DEGoodmanWilson:
Allow Slack to better understand how this SDK is being used.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -35,13 +35,21 @@ var getVersionString = function getVersionString() {
   var i;
 
   for (i = 0; i < versions.length; ++i) {
-    str = str + versions[i][0] + '/' + versions[i][1] + ' ';
+    if (Array.isArray(versions[i])) {
+      str = str + versions[i][0] + '/' + versions[i][1] + ' ';
+    } else {
+      str = str + versions[i];
+    }
   }
   return str.trim();
 };
 
 var appendToVersionString = function appendToVersionString(name, version) {
-  versions.push([name, version]);
+  if (version) {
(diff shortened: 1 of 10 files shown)


lib/helpers.js
```diff
@@ -21,5 +23,36 @@ var getLogger = function getLogger(optLogLevel, optTransport) {
...
+  [os.platform(), os.release()],
+  ['node', process.version.replace('v', '')]
+];
+
+var getVersionString = function getVersionString() {
+  var str = '';
+  var i;
+
+  for (i = 0; i < versions.length; ++i) {
+    if (Array.isArray(versions[i])) {
```
aoberoi: you built this array as a literal, why are you checking for this?

aoberoi: oh i see, its because the `appendToVersionString()` function adds user-generated.

lib/helpers.js
```diff
@@ -21,5 +23,24 @@ var getLogger = function getLogger(optLogLevel, optTransport) {
...
+// TODO we can do this better
+var versionStr = module.exports.repository.replace('/', ':') + '/' + module.exports.version + ' '
+  + os.platform() + '/' + os.release() + ' '
+  + 'node/' + process.version.replace('v', '');
+
+var getVersionString = function getVersionString() {
+  return versionStr;
+};
+
+var appendToVersionString = function appendToVersionString(name, version) {
```
aoberoi: can we put some JSDoc on this function so that its surfaced by our reference docs? the intention is for this to be used by "higher-level" users of the SDK, right?

PR #326, opened 2017-03-08
Clarify webhook usage

frrakn:
`webhook.send` not returning `err` does not mean the message successfully sent - just means the request successfully went through.


aoberoi:
@frrakn this might be a stupid question, but can you give me an example of a case where there is no `err` but the message was not sent?

frrakn:
This was a while back, so difficult to remember. As I vaguely recall, `err` comes up when theres a network / communication error, whereas if you successfully make a request that results in a 4XX or 5XX it will not return an error, but will return a bad status code.

Feel free to close, this information isn't particularly fresh in my mind.

aoberoi:
LGTM

Issue #245, opened 2016-07-07
Confusing implications of "UNABLE_TO_RTM_START"


mhahn:
I think https://github.com/slackhq/node-slack-sdk/blob/master/lib/clients/rtm/client.js#L348 should be some event other than `UNABLE_TO_RTM_START`.

here: https://github.com/slackhq/node-slack-sdk/blob/master/lib/clients/rtm/client.js#L258, an `UNABLE_TO_RTM_START` message is published, but that doesn't mean anything definitively. If it happens to be an error it can't recover from, `DISCONNECT` is later published. If `autoReconnect` is enabled, it will attempt to reconnect. Which might publish `UNABLE_TO_RTM_START` because we've exceeded the max connection attempts.

I think in the case of: https://github.com/slackhq/node-slack-sdk/blob/master/lib/clients/rtm/client.js#L348, we should raise either a distinct message or call `DISCONNECT` so that at least we're consistent.

My goal is to know when an RTM connection is hosed, ie. `DISCONNECT` but right now `UNABLE_TO_RTM_START` can also sometimes indicate that the connection is lost.

mhahn:
Sure.

**Scenario 1: Connecting with a token that is no longer valid (ie. `account_inactive` is returned)**
- emit `CLIENT_EVENTS.CONNECTING`
- start fails with `account_inactive` which is listed within https://github.com/slackhq/node-slack-sdk/blob/master/lib/clients/rtm/client.js#L24
- emit `CLIENT_EVENTS.UNABLE_TO_RTM_START`
- emit `CLIENT_EVENTS.DISCONNECT`

this token is invalid and something should be done (ie. deactivating the team in your system to prevent future connections)

**Scenario 2: Network failure, auto-reconnect not enabled**
- emit `CLIENT_EVENTS.CONNECTING`
- start fails with some error not listed in https://github.com/slackhq/node-slack-sdk/blob/master/lib/clients/rtm/client.js#L24
- emit `CLIENT_EVENTS.UNABLE_TO_RTM_START`

you wish you would have enabled auto reconnecting

**Scenario 3: Network failure, auto-reconnect enabled**
- emit `CLIENT_EVENTS.CONNECTING`
- start fails with some error not listed in https://github.com/slackhq/node-slack-sdk/blob/master/lib/clients/rtm/client.js#L24
- emit `CLIENT_EVENTS.UNABLE_TO_RTM_START`
[...]

Issue #348, opened 2017-05-17
Ping / pong for rtm client is not implemented correctly


CharlieHess:
* [x] I've read and understood the [Contributing guidelines](./CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](./CODE_OF_CONDUCT.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

#### Description
The current logic for ping / pong can lead to spurious disconnects. There are a few reasons for this:

1. We save `_lastPong` when we get a pong message response, but don't check it until the _next_ time we `_pingServer`. Since we `_pingServer` on a 5 second interval, this means the gap can never be lower than 5 seconds. Basically we're imposing a 5 second handicap. :joy:
1. We don't let other messages count as pongs. This can cause starvation if the socket is super busy processing other work **cough** IBM **cough**.
1. On some platforms, like [macOS](https://github.com/electron/electron/issues/5101), timers are artificially throttled when the application is backgrounded. This, in combination with 1, can result in a lot of unnecessary reconnects.

#### Steps to reproduce:

1. Set a `debug` `logLevel`
2. Drop `logger` statements in `_pingServer` and `_handlePong`
3. Subtract the intervals

aoberoi:
fixed in #349

Issue #170, opened 2016-03-08
Add some useful retry policies, aliased with reasonable names


l12s:

l12s:
#224 partially addresses this

DEGoodmanWilson:
Could you provide a little more detail on what you were hoping to achieve?

aoberoi:
fixed in #349

PR #349, opened 2017-05-17
Make rtm connection more reliable

CharlieHess:

## PR Summary
This PR tries to address unreliable RTM connections, in a few different ways:

* Double the max pong interval.
  * For comparison, the Slack webapp uses a 60 second timeout, so we've still got a lot of room here.
* Calculate the pong latency differently.
  * Noted in #348, we're imposing a 5 second handicap on ourselves.
* Allow other socket messages to stand in for a pong in some cases.
* Move the ping-pong startup to align with `RTM_CONNECTION_OPENED`.
  * Previously it was started in `WS_OPEN` but that seemed inconsistent.
* Change the default retry policy to reconnect forever.
  * The previous implementation would retry five times in five minutes, then throw `UNABLE_TO_RTM_START`, then repeat the process `MAX_CONNECTION_ATTEMPTS` (10 by default). This was a result of us keeping around two different approaches for retries and led to confusion in #245.
  * @l12s noted that we should just rely on the policy specified in `retryConfig` and remove `MAX_CONNECTION_ATTEMPTS`, which is what we're doing now.

## Related Issues
See #348, #245, and #170.

## Test strategy
There are now tests around the entire retry policy:
```
Retry policy
      ✓ should call the rtm.start error callback if attempts exceed retries
      ✓ should disconnect if the error is unrecoverable
      ✓ should disconnect if autoReconnect is set to false
      ✓ should use max attempts and backoff options in place of a retryConfig
      ✓ should support retrying forever with exponential backoff 
```


docs/_reference/RTMClient.md
```diff
@@ -26,8 +26,9 @@ permalink: /reference/RTMClient
     * [.handleWsMessage(wsMsg)](#RTMClient+handleWsMessage)
     * [._handleMessageAck(replyTo, message)](#RTMClient+_handleMessageAck)
     * [.handleWsError(err)](#RTMClient+handleWsError)
-    * [.handleWsClose()](#RTMClient+handleWsClose)
```
aoberoi: were there arguments that were previously undocumented, or actually no arguments?

CharlieHess: previously undocumented.

docs/_reference/RTMClient.md
```diff
@@ -140,14 +139,16 @@ Disconnects from the RTM API.
...
-| optCode | 
+| Param | Type |
+| --- | --- |
+| optReason | <code>Error</code> | 
+| optCode | <code>Number</code> | 
 
 <a name="RTMClient+reconnect"></a>
 
 ### rtmClient.reconnect()
+Attempts to reconnect to the websocket by retrying [rtm.start](rtm.start).
```
aoberoi: maybe not be so specific, since it could be an `rtm.connect()` instead of `rtm.start()`, right?

lib/clients/rtm/client.js
```diff
@@ -47,8 +47,6 @@ var wsSocketFn = require('../transports/ws');
  * @param {boolean} opts.autoReconnect Whether or not to automatically reconnect when the connection
  *     closes.
  * @param {boolean} opts.useRtmConnect True to use rtm.connect rather than rtm.start
- * @param {number} opts.maxReconnectionAttempts
```
aoberoi: one concern here: removing these options from the API is backwards-incompatible. can we give some thought to how we might be able to turn these options into a parameterized retry policy on the fly?

aoberoi: we can also mark them `@deprecated`

lib/clients/rtm/client.js
```diff
@@ -359,48 +355,19 @@ RTMClient.prototype.disconnect = function disconnect(optErr, optCode) {
 
 
 /**
- * Attempts to reconnect to the websocket.
+ * Attempts to reconnect to the websocket by retrying {@link rtm.start}.
```
aoberoi: again, might me too specific

docs/_reference/RTMClient.md
aoberoi: got it 👍

lib/clients/rtm/client.js
CharlieHess: good call, I think we can turn them into a retry policy. Important to note that the current retry behavior is to use `FIVE_RETRIES_IN_FIVE_MINUTES` _first_, and only once that has failed do we count it as a "connection attempt" and apply the backoff defined here. In other words, we're already way off-base in terms of applying these variables.

CharlieHess:
It was a bit of a PITA but I split the RTM tests into separate categories (`Connection logic`, `Retry policy`, and `Message handling`) and added the following tests for the retry policy:

```
Retry policy
      ✓ should call the rtm.start error callback if attempts exceed retries
      ✓ should disconnect if the error is unrecoverable
      ✓ should disconnect if autoReconnect is set to false
      ✓ should use max attempts and backoff options in place of a retryConfig
      ✓ should support retrying forever with exponential backoff 
```

This fixes #245 and maintains backwards compatibility for folks who were using `maxReconnectionAttempts` or `reconnectionBackoff`.

lib/clients/rtm/client.js
```diff
@@ -35,25 +34,28 @@ var ChatFacet = require('../web/facets').ChatFacet;
...
+ * @param {Function} opts.socketFn      A function to call, passing in a websocket URL, that should
+ * return a websocket instance connected to that URL
+ * @param {Object} opts.dataStore       A store to cache Slack info, e.g. channels, users etc. in.
+ * Pass null or false to use no store
+ * @param {Boolean} opts.autoReconnect  Whether or not to automatically reconnect when the connection
+ * closes. Defaults to true
+ * @param {Boolean} opts.useRtmConnect  True to use rtm.connect rather than rtm.start
+ * @param {Object} opts.retryConfig     The retry policy to use, defaults to forever with exponential
+ * backoff {@see https://github.com/SEAPUNK/node-retry}
+ * @param {Number} opts.maxReconnectionAttempts DEPRECATED: Use retryConfig instead
```
aoberoi: yay!

lib/clients/rtm/client.js
```diff
@@ -433,21 +414,21 @@ RTMClient.prototype.handleWsMessage = function handleWsMessage(wsMsg) {
...
-    // TODO(leah): Emit an event here?
-    this.logger('debug', 'unable to parse message: ' + err);
+    this.logger('error', 'Unable to parse message: ' + err);
     return;
   }
 
   if (includes(RTM_CLIENT_INTERNAL_EVENT_TYPES, message.type)) {
     this._handleWsMessageInternal(message.type, message);
   } else {
+    this._maybeKeepAlive(message);
```
aoberoi: @CharlieHess help me understand this, please. `RTM_CLIENT_INTERNAL_EVENT_TYPES` includes `'pong'`, so pong messages will *not* keep the socket connection alive?

aoberoi: nevermind, i think i see how this works now.

CharlieHess: so the idea is, pong messages will still keep it alive, but if we _don't_ receive a pong message because we are being flooded with so many other legitimate messages, we'll treat those messages like a pong and not close the socket

test/clients/rtm/connection-logic.js
```diff
@@ -0,0 +1,162 @@
...
+      };
+
+      var secondConnection = function (wss, rtm) {
+        expect(rtm.reconnect.calledOnce).to.equal(true);
+      };
+
+      testReconnectionLogic([firstConnection, secondConnection], done);
+    });
+
+    // This is overly complex for what it's trying to test (that a state var is getting toggled),
```
aoberoi: props for fleshing it out though!

aoberoi:
words cannot express how excited i am for this change.

@CharlieHess, this one's for you:

tenor

PR #356, opened 2017-05-24
Support a "retry now" method

CharlieHess:

## 🚧 Note 
Merge #349 first then rebase.

## PR Summary
This PR builds on the work in #349. Since we're using a "retry forever with exponential backoff" policy, we can hit the cap of 30 minutes relatively quickly. If we then regain network or wake from sleep, we want to be able to recreate the socket connection immediately rather than waiting for the next retry. This PR adds a `forceRetry` parameter to `rtm.reconnect` to support this.

Then, in my client-side code, I can do something like:

``` js
// If we have network when we resume from sleep, retry immediately
async handleWakeFromSleep() {
  const online = await isOnline({ timeout: 30000 });
  if (online) {
    this.rtm.reconnect(true);
  }
}
```

## Related Issues
No issues filed.

## Test strategy
One new test was added:
```
 ✓ should support reconnecting immediately
```

### DIFF
--- a/lib/clients/client.js
+++ b/lib/clients/client.js
@@ -101,10 +101,11 @@ BaseAPIClient.prototype.registerDataStore = function registerDataStore(dataStore
  * Empties the request queue and returns it to a clean slate.
  */
 BaseAPIClient.prototype.clearQueue = function clearQueue() {
-  this.logger('info', 'Clearing the request queue', this.requestQueue.concurrency);
+  this.logger('info', 'Clearing the request queue');
+  this.requestQueue.kill();
 
-  // Initially tried to accomplish this with `requestQueue.kill`, but it leaves the queue blocked.
-  // See https://github.com/caolan/async/issues/1422.
+  // Initially tried to accomplish this with just `requestQueue.kill`, but it leaves the queue
+  // blocked. See https://github.com/caolan/async/issues/1422.
   this.requestQueue = async.queue(
     bind(this._callTransport, this),
     this.requestQueue.concurrency || 3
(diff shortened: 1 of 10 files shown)


lib/clients/client.js
```diff
@@ -102,6 +98,22 @@ BaseAPIClient.prototype.registerDataStore = function registerDataStore(dataStore
...
 /**
+ * Empties the request queue and returns it to a clean slate.
+ */
+BaseAPIClient.prototype.clearQueue = function clearQueue() {
+  this.logger('info', 'Clearing the request queue');
+  this.requestQueue.kill();
+
+  // Initially tried to accomplish this with just `requestQueue.kill`, but it leaves the queue
+  // blocked. See https://github.com/caolan/async/issues/1422.
+  this.requestQueue = async.queue(
```
aoberoi: there's some code duplication here with L61-L63. but honestly, that can be cleaned up with a later pass.

lib/clients/transports/call-transport.js
```diff
@@ -102,15 +101,15 @@ var handleHttpResponse = function handleHttpResponse(body, headers, client, apiC
   if (!jsonResponse.ok) {
     jsonError = new Error(jsonResponse.error);
     client.logger('error', 'Response not OK: ', jsonResponse.error);
-  }
```
aoberoi: what if the scopes are relevant to the error that just occurred, and the client wanted this data in the `jsonResponse` object?

aoberoi:
@CharlieHess anything i can do to help this progress?

CharlieHess:
@aoberoi after further consideration, we don't need this method. Just need to be sure to call `disconnect` before you `reconnect`.

PR #394, opened 2017-09-05
docs(README): remove inaccurate comment

AriLFrankel:
#### PR Summary
Remove an inacurate comment which makes reference to a section that
does not exist


aoberoi:
thanks for the correction... we probably should write a topic about sensitive data, for now this is def an improvement 👍

PR #432, opened 2017-12-13
Optionally allow a user ID to be passed into users.identity for WTA

bertrandom:
###  Summary

users.identity can now be passed a user ID when used with Workspace Token Apps. This allows an optional user ID to be passed into the method, while supporting the old function signature where this parameter used to be a callback.

### DIFF
--- a/lib/clients/web/facets/users.js
+++ b/lib/clients/web/facets/users.js
@@ -40,19 +40,24 @@ UsersFacet.prototype.getPresence = function getPresence(user, optCb) {
  * Get a user's identity.
  * @see {@link https://api.slack.com/methods/users.identity|users.identity}
  *
+ * @param {Object=} options
+ * @param {?} opts.user - When calling this method with a workspace token, set this to
+ * the user ID of the user to retrieve the identity of
  * @param {function=} optCb Optional callback, if not using promises.
  */
-UsersFacet.prototype.identity = function identity(user, optCb) {
+UsersFacet.prototype.identity = function identity(options, optCb) {
   var cb = null;
   var opts = {};
 
   cb = optCb;
 
-  if (user) {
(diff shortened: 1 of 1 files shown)


aoberoi:
@bertrandom this is great! i have a couple questions for ya.

1. the user argument isn't documented on [the reference](https://api.slack.com/methods/users.identity). will it be? if so, when? if not, then i'm not too keen on adding "hidden" arguments into the SDK's API. if those arguments change, we will break users, and hidden arguments are subject to change with no announcement.

2. we've had a number of issues trying to "upgrade" a facet from having only explicit position arguments to later accepting an object for optional arguments. i'd like to future proof where we can. this would be API gymnastics, but how do you feel about an implementation that supports all these call sites?
  *  `web.users.identity()`
  *  `web.users.identity(callback)`
  *  ~`web.users.identity(user)`~
  *  `web.users.identity(user, callback)`
  *  `web.users.identity(user, options)`
  *  `web.users.identity(user, options, callback)`

aoberoi:
**Update**: i'm dumb.

`user` would be an optional argument, not a required one (which we translate into positional arguments in the facet).

we actually want to support these call sites then:
  *  `web.users.identity()`
  *  `web.users.identity(callback)`
  *  `web.users.identity(options)`
  *  `web.users.identity(options, callback)`

bertrandom:
@aoberoi Let me try to get that parameter unhidden before we change this. And you make a good point, using options instead of user will help us future-proof if we add more params later, I'll update the PR.

bertrandom:
@aoberoi Okay, I modified this to pass an options object, would you mind reviewing it one last time and then merging it in when you get a chance? Thanks.

aoberoi:
thanks so much @bertrandom!

PR #450, opened 2018-01-11
Fixes a bug where args were being overridden by previous requests when using requestOptionsTransport

bertrandom:
###  Summary

There's a bug when using the requestOptionsTransport where the options object starts storing arguments from previous requests. This makes a clone of the object so that even if it's mutated, a new one is used for each request.

### DIFF
--- a/lib/clients/transports/request.js
+++ b/lib/clients/transports/request.js
@@ -6,6 +6,7 @@ var HttpsProxyAgent = require('https-proxy-agent');
 var has = require('lodash').has;
 var partial = require('lodash').partial;
 var defaults = require('lodash').defaults;
+var cloneDeep = require('lodash').cloneDeep;
 var request = require('request');
 
 var handleRequestTranportRes = function handleRequestTranportRes(cb, err, response, body) {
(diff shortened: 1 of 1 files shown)


lib/clients/transports/request.js
```diff
@@ -64,7 +65,8 @@ var proxiedRequestTransport = function proxiedRequestTransport(proxyURL) {
 
 var requestOptionsTransport = function requestOptionsTransport(options) {
   return function _requestOptionsTransport(args, cb) {
-    var requestArgs = defaults(options, getRequestTransportArgs(args));
+    var instanceOptions = cloneDeep(options);
```
aoberoi: could this line be outside of the inner `_requestOptionsTransport` function? or in other words, what's the point in cloning the `options` each time a request is fired, as opposed to just cloning it once per transport "constructed"?

bertrandom: I think the problem is that the way defaults works is by putting instanceOptions as the base of requestArgs and then layering the args on top of it. This means that after the args are layered on top, instanceOptions has now been overridden with the args specified by the caller.

This is problematic for the next request because then instanceOptions now also includes the args from the previous request. By ensuring that the original options object doesn't mutate and cloning it for each request, we avoid this issue.

aoberoi: ah! i was assuming `defaults()` is a pure function. silly lodash developers. 👍

PR #487, opened 2018-03-11
allow unspecified arguments, fixes #484

### LINKED ISSUE #484: allow unspecified arguments in method types
### Description

right now, with all strictness checks on using TS 2.6.2, you will get the following error when using an argument that isn't explicitly specified: `Object literal may only specify known properties, and '{argument}' does not exist in type '{method type}'`.

clavin:
###  Summary

Fixes #484.

This PR adds the ability to use arguments not specified in the `MethodArguments` types (auxiliary arguments). For example:

```ts
web.chat.postEphemeral({
    user: 'not me',
    channel: 'somewhere',
    text: 'wait, what did that say?',

    extra_argument: 'one that doesnt exist, in fact'
});
```

Note that this preserves static errors for *missing* arguments, so if `channel` in the above example was missing, an error will still be thrown (when using types, of course).

### Requirements

### DIFF
--- a/src/methods.ts
+++ b/src/methods.ts
@@ -8,13 +8,17 @@ import { WebAPICallOptions, WebAPIResultCallback, WebAPICallResult } from './Web
  */
 export default interface Method<MethodArguments extends WebAPICallOptions> {
   // TODO: can we create a relationship between MethodArguments and a MethodResult type?
-  (options?: MethodArguments): Promise<WebAPICallResult>;
-  (options: MethodArguments, callback: WebAPIResultCallback): void;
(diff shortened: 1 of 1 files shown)


aoberoi:
thanks so much @clavin, i'm happy with this fix. i was thinking about doing roughly the same thing. the tradeoff here is that argument names that have typos (`chanel` instead of `channel`) will not trigger an error. that's less than ideal, but i think once all the methods have an interface for their arguments, we can try to remove `AuxilaryArguments` (but we might want to preserve it for direct calls to `.apiCall()` so that can always be seen as an "escape hatch" if you know what you're doing).

Issue #307, opened 2017-01-03
fix files.upload from Buffer with formData options


jsarafajr:
Getting `Error: no_file_data` while uploading Buffer object with files.upload.

```js
// doesn't work (see trace log below)
const image = fs.readFileSync('image.png');
await bot.files.upload('image.png', {
  file: image
});

// text buffer doesn't work too
await bot.files.upload('text.txt', {
  file: new Buffer('hello world');
});

// uploads image as Plain Text (incorrect)
await bot.files.upload('image.png', {
  content: image
});

// works well with streams
const imageStream = fs.createReadStream('image.png')
await bot.files.upload('image.png', {
  file: imageStream
});
```

Stacktrace: 
```bash
2016-12-27T13:58:29.415Z - error:  Error: no_file_data
     at handleHttpResponse (/bot/node_modules/@slack/client/lib/clients/transports/call-transport.js:105:17)
     at handleTransportResponse (/bot/node_modules/@slack/client/lib/clients/transports/call-transport.js:155:19)
     at apply (/bot/node_modules/lodash/lodash.js:499:17)
     at wrapper (/bot/node_modules/lodash/lodash.js:5356:16)
     at Request.handleRequestTranportRes (/bot/node_modules/@slack/client/lib/clients/transports/request.js:20:5)
     at apply (/bot/node_modules/lodash/lodash.js:499:17)
     at Request.wrapper [as _callback] (/bot/node_modules/lodash/lodash.js:5356:16)
     at Request.self.callback (/bot/node_modules/request/request.js:186:22)
...
```

linuxfreakus:
I too am interested in how to make this happen.  I've tried wrapping the Buffer as a Readable stream... no dice.  It works if I write my buffer to a temp file and then use `fs.createReadStream` but that is really messy, slow, and not fun.  I just want to send my data that is already available in memory without having to jump through hoops.

Everesty:
Took me quite a bit of searching to look for the issue, but it lies in the fact that the module [`request`](https://www.npmjs.org/package/request) uses, [`form-data`](https://www.npmjs.org/package/form-data), only recognizes **certain stream objects** as files when evaluating what is a file and what is a multi-part data parameter:

> Form-Data can recognize and fetch all the required information from common types of streams **(fs.readStream, http.response and mikeal's request)**, for some other types of streams you'd need to provide "file"-related information manually:

[...]

PR #500, opened 2018-03-14
fix file upload - consider metadata options by creating form data

KharitonOff:
###  Summary
fix file upload issue, now you can upload buffer stream as expected:
```
let file = fs.readFileSync('../file.png')
slackClient.files.upload({ channels, file, filename: 'file.png' })
```
as well as proposed fix in #307:
```
let file = fs.readFileSync('../file.png')
slackClient.files.upload({ channels, file: {value: file, options:{filename: 'file.png'}}, filename: 'file.png' })
```
related to #307


src/WebClient.spec.js
```diff
@@ -390,7 +400,7 @@ describe('WebClient', function () {
           // verify that any requests after maxRequestConcurrency were delayed by the responseDelay
           const queuedResponses = responses.slice(3);
           const minDiff = concurrentResponses[concurrentResponses.length - 1].diff + responseDelay;
-          queuedResponses.forEach(r => assert.isAbove(r.diff, minDiff));
+          queuedResponses.forEach(r => assert.isAtLeast(r.diff, minDiff));
```
aoberoi: was this actually causing a failure? i think your change is more technically correct, but i'm wondering if you actually got a test run where the timers were so perfect that `r.diff` and `minDiff` were actually equal.

src/WebClient.spec.js
```diff
@@ -403,13 +413,15 @@ describe('WebClient', function () {
...
+          const concurrentResponses = responses.slice(0, 1); // the first response
           concurrentResponses.forEach(r => assert.isBelow(r.diff, responseDelay));
 
           // verify that any requests after maxRequestConcurrency were delayed by the responseDelay
-          const queuedResponses = responses.slice(1);
+          const queuedResponses = responses.slice(1);// the second response
           const minDiff = concurrentResponses[concurrentResponses.length - 1].diff + responseDelay;
-          queuedResponses.forEach(r => assert.isAbove(r.diff, minDiff));
+          queuedResponses.forEach(r => {
+            assert.isAtLeast(r.diff, minDiff)
```
aoberoi: i think this could fit in a single line rather than a block.

src/WebClient.spec.js
```diff
@@ -201,18 +201,28 @@ describe('WebClient', function () {
...
+          });
+      });
+
+      it('should properly serialize when the file is an object with value as a Buffer and options containing filename', function () {
+        const imageBuffer = fs.readFileSync(path.resolve('test', 'fixtures', 'train.jpg'));
+
+        // intentially vague about the method name
+        return this.client.apiCall('upload', {
+          file: { value: imageBuffer, options: { filename: 'train.png' } },
+          filename: 'train.png',
```
aoberoi: i was trying to avoid this usage all together, because it seems redundant to have to specify the filename twice. can you see any advantage to doing it this way?

src/WebClient.spec.js
```diff
@@ -201,18 +201,28 @@ describe('WebClient', function () {
...
-            filename: 'train.png',
-          })
+          file: imageBuffer,
+          filename: 'train.png',
+        })
+          .then((parts) => {
+            assert.lengthOf(parts.files, 1);
+            const file = parts.files[0];
+            // options were not provided to the form builder
+            assert.include(file, { fieldname: 'file' });
```
aoberoi: still not asserting that the `filename` field exists, and removed the comment that would remind us to improve the tests so that it is asserted. why?

src/WebClient.spec.js
```diff
@@ -201,18 +201,28 @@ describe('WebClient', function () {
...
           .then((parts) => {
             assert.lengthOf(parts.files, 1);
             const file = parts.files[0];
-            // TODO: understand why this assertion is failing. already employed the buffer metadata workaround, should
-            // look into the details about whether that workaround is still required, or why else the `source.on` is not
-            // defined error would occur, or if Slack just doesn't need a filename for the part
-            // assert.include(file, { fieldname: 'file', filename: 'train.jpg' });
-            // NOTE: it seems the file and its filename are emitted as a field in addition to the token, not sure if
-            // this was happening in the old implementation.
             assert.include(file, { fieldname: 'file' });
```
aoberoi: still not asserting that the `filename` field exists.

aoberoi:
@KharitonOff thanks for your contribution! can you help clarify by answering my comments above?

src/WebClient.spec.js
clavin: I can confirm that I've run into the issue of getting a test run where the timers matched up and a false positive was thrown--just once, but it has happened.

src/WebClient.spec.js
KharitonOff: The main advantage I see is the current implementation state of all the folks who used the old style. I don't know whether there is a need for someone to provide further attributes in the  options field (e.g. contentType or knownLength)

src/WebClient.spec.js
KharitonOff: yes, I added this fix to the PR, because the build of my first commit broke accidentally

src/WebClient.spec.js
KharitonOff: sorry, I've overseen this assertion statement and felt that the fix solved the problem pointed in the comment. I'll bring your TODO and assertion back.

KharitonOff:
I've changed your test again in order to fix the build. I hope it is ok for you.
The last travis build failed because of timeout. This seems to be an occasional fail:
build for node v9 succeeds:
screen shot 2018-03-15 at 16 10 39
build for node v8 fails:
screen shot 2018-03-15 at 16 11 08
https://travis-ci.org/slackapi/node-slack-sdk/builds/353822357

test/typescript/test.ts
```diff
@@ -4,7 +4,7 @@ import { check, checkDirectory } from 'typings-tester';
 describe('typescript typings tests', () => {
   it('should allow WebClient to work without a token', () => {
     check([`${__dirname}/sources/webclient-no-token.ts`], `${__dirname}/tsconfig-strict.json`);
-  });
```
aoberoi: maybe the best place to override this value is in `test/mocha.opts`, so that it doesn't just apply to this one case. i don't see anything unique to this case that should make it run longer than the other. its my guess that Travis generally has unpredictable speed and we should make the ceiling high enough to accomodate.

src/WebClient.spec.js
aoberoi: i also want to make sure the implementation accounts for those who used the old style, but i prefer that the path we take to get there is not to redo the same hack, but to smooth out the API so that you don't need to redundantly specify the filename.

i wouldn't call it a hack if there was legitimately a use case or need for using the `{ value, options }` format for `file`. i don't see any benefit from exposing `contentType` and/or `knownLength` at this time, since that only has the potential to force to incorrect values and cause the upload to fail (but I'm happy to be corrected if there is a real use case).

maybe we can channel this effort towards investigating why the following code didn't behave as we want: https://github.com/slackapi/node-slack-sdk/blob/a206c3e1286be958c963795f9ce80a18cb85ce7d/src/WebClient.ts#L530-L537

AFAICT it should result in the same thing...

also, have you tried this implementation to upload actual files? i'm not 100% confident that the nock expectations are correct.

KharitonOff: Sorry, but I don't really understand your comment... this code snippet, you mentioned above works now as well. Off course I have tried it before I've opened this PR. And the problem with it, why it didn't work before was in the implementation of 
https://github.com/slackapi/node-slack-sdk/blob/a206c3e1286be958c963795f9ce80a18cb85ce7d/src/WebClient.ts#L556-L561

I've added this part in order to provide the attribute `options` to the form builder.
```
if (key === 'file' && value.value) {
    form.append(key, value.value, value.options);
    return form;
}
```
https://github.com/KharitonOff/node-slack-sdk/blob/45c77022a81cd2631a5a2bd3ee0dd29e401f51d7/src/WebClient.ts#L554-L563

This is actually the main goal of this PR and all other things we talked here about, were done with intention not to break the current state. As I mentioned in the PR description, you can upload files now with no need of additional/redundant information
```
let file = fs.readFileSync('../file.png')
slackClient.files.upload({ channels, file, filename: 'file.png' })
```

Should i remove the 'backup' solution to restrict the file upload to only one way?

test/typescript/test.ts
KharitonOff: Agree! I could not find this setting before.

src/WebClient.spec.js
aoberoi: @KharitonOff thanks so much for your clarification. i missed that diff amidst the rest of the changes but i now see where the issue was before.

yes, i'd prefer removing the backup solution so that we don't unnecessarily expand the API contract. if you can do that, i'm happy to land this PR!

aoberoi:
@KharitonOff its a small world. I just noticed that you work on CLA Assistant (which we use) and you helped review an implementation for a feature i requested, way back when (https://github.com/cla-assistant/cla-assistant/issues/246). 🙌

KharitonOff:
@aoberoi yes, indeed, glad to meet you again! It is nice to see that you are using CLA Assistant and it was very interesting to have a contributer role and sign your CLA 😄

src/WebClient.spec.js
```diff
@@ -201,34 +206,31 @@ describe('WebClient', function () {
...
+          file: imageBuffer,
+          filename: 'train.jpg',
+        })
           .then((parts) => {
             assert.lengthOf(parts.files, 1);
             const file = parts.files[0];
-            // TODO: understand why this assertion is failing. already employed the buffer metadata workaround, should
-            // look into the details about whether that workaround is still required, or why else the `source.on` is not
-            // defined error would occur, or if Slack just doesn't need a filename for the part
-            // assert.include(file, { fieldname: 'file', filename: 'train.jpg' });
```
aoberoi: note to self: should the assertion with filename work now?

src/WebClient.spec.js
```diff
@@ -241,19 +243,19 @@ describe('WebClient', function () {
...
             assert.lengthOf(parts.files, 1);
             const file = parts.files[0];
-            // TODO: understand why this assertion is failing. already employed the buffer metadata workaround, should
+            // TODO: understand why this assertion was failing. already employed the buffer metadata workaround, should
             // look into the details about whether that workaround is still required, or why else the `source.on` is not
             // defined error would occur, or if Slack just doesn't need a filename for the part
-            // assert.include(file, { fieldname: 'file', filename: 'train.jpg' });
+            assert.include(file, { fieldname: 'file', filename: 'train.jpg' });
 
             // NOTE: it seems the file and its filename are emitted as a field in addition to the token, not sure if
```
aoberoi: note to self: the following assertion is probably redundant now

aoberoi:
This is so great!! Extra thanks for cleaning up the style issues and some incorrect behavior in the existing tests.

I'm going to follow up with a PR that addresses the "note to self" comments above.

KharitonOff:
Cool! Thanks for merging it 👍
I hope you'll release it soon, our 🤖 is eagerly waiting for the fix 😃

PR #508, opened 2018-03-15
Make IncomingWebhook `send` promisable

clavin:
###  Summary

Saw #501, decided the addition would be easy because `IncomingWebhook#send` was already defined as if it was planned to be promisable.

I also set up the tests to be similar to that of `WebClient` for testing the result of a method that can callback or promise.

### DIFF
--- a/src/IncomingWebhook.ts
+++ b/src/IncomingWebhook.ts
@@ -49,7 +49,10 @@ export class IncomingWebhook {
    * @param message the message (a simple string, or an object describing the message)
    * @param callback
    */
-  public send(message: string | IncomingWebhookSendArguments, callback: IncomingWebhookResultCallback): void {
+  public send(message: string | IncomingWebhookSendArguments): Promise<void>;
+  public send(message: string | IncomingWebhookSendArguments, callback: IncomingWebhookResultCallback): void;
+  public send(message: string | IncomingWebhookSendArguments,
+              callback?: IncomingWebhookResultCallback): Promise<void> | void {
     // NOTE: no support for proxy
     // NOTE: no support for TLS config
     let payload: IncomingWebhookSendArguments = this.defaults;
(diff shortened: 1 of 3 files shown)


src/IncomingWebhook.ts
```diff
@@ -49,7 +49,10 @@ export class IncomingWebhook {
    * @param message the message (a simple string, or an object describing the message)
    * @param callback
    */
-  public send(message: string | IncomingWebhookSendArguments, callback: IncomingWebhookResultCallback): void {
+  public send(message: string | IncomingWebhookSendArguments): Promise<void>;
```
aoberoi: if its going to be a Promise of something, why not a promise of a `string` (the response text), or better yet, define a type for it (an object that might get more properties in the future).

src/IncomingWebhook.ts
```diff
@@ -67,6 +70,10 @@ export class IncomingWebhook {
       return;
     });
```
aoberoi: maybe a `catch()` to change errors into a `CodedError`?

src/IncomingWebhook.ts
clavin: While I agree that *something* should be returned, I somewhat feel that's out of the scope of the issue I was originally targeting.

I simply was aiming to expose the [already-existing implementation](https://github.com/slackapi/node-slack-sdk/blob/a206c3e1286be958c963795f9ce80a18cb85ce7d/src/IncomingWebhook.ts#L63-L68) (which already intentionally doesn't return anything) as a `Promise`, rather than change the implementation to be of more aid.

Anyhow, I'll gladly tack this on 👍✨

src/IncomingWebhook.ts
clavin: Oh, good *catch*! ([I apologize for that pun.](https://img.etsystatic.com/il/6637bd/749993407/il_570xN.749993407_65ss.jpg))

aoberoi:
taking a look at the codecov failure, it looks like the uncovered part of the diff is in typically hard to reach code paths - ones that are not covered for the same reason in `WebClient`. i think this warrants an exception to the checks so i'm overriding that block.

PR #519, opened 2018-03-22
Generalize handling of arguments with binary value types

### LINKED ISSUE #452: reactions.get Facet fails to send request when requesting "full: true" and "file: id"

aoberoi:
###  Summary

based on #513 - as this PR states, this should allow methods like `users.setPhoto` to also work properly.

fixes #452 

### DIFF
--- a/src/WebClient.spec.js
+++ b/src/WebClient.spec.js
@@ -215,48 +215,6 @@ describe('WebClient', function () {
           });
       });
 
-      it('should properly serialize when binary argument is a Buffer and filename is specified', function () {
-        const imageBuffer = fs.readFileSync(path.resolve('test', 'fixtures', 'train.jpg'));
-
-        // intentially vague about the method name and argument name
-        return this.client.apiCall('upload', {
-          someBinaryField: imageBuffer,
-          filename: 'train.jpg',
-        })
-          .then((parts) => {
-            assert.lengthOf(parts.files, 1);
-            const file = parts.files[0];
-            assert.include(file, { fieldname: 'someBinaryField' });
-            assert.notInclude(file, { filename: 'train.jpg' });
-
-            assert.lengthOf(parts.fields, 2);
-            assert.deepInclude(parts.fields, { fieldname: 'token', value: token });
-            assert.deepInclude(parts.fields, { fieldname: 'filename', value: 'train.jpg' });
-          });
-      });
-
-      // TODO: re-enable this once we decide if we want to generate a random name or if we want to log a warning
-      it.skip('should log a warning when binary argument is a Buffer and there is no filename', function () {
-        const imageBuffer = fs.readFileSync(path.resolve('test', 'fixtures', 'train.jpg'));
-        const output = [];
-        const stub = function (level, message) {
-          output.push([level, message]);
-        }
-        const debuggingClient = new WebClient(token, { logLevel: LogLevel.WARN, logger: stub });
-
-        // intentially vague about the method name
-        return debuggingClient.apiCall('upload', {
-          someBinaryField: imageBuffer,
-        })
-          .then(() => {
-            assert.isNotEmpty(output);
-            const anyLogLineIsLevelWarn = output.reduce((acc, line) => {
-              return acc || (line.indexOf('warn') !== -1)
-            }, false);
-            assert(anyLogLineIsLevelWarn);
-          });
-      });
-
       it('should properly serialize when the binary argument is a ReadableStream', function () {
         const imageStream = fs.createReadStream(path.resolve('test', 'fixtures', 'train.jpg'));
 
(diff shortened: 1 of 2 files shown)


src/WebClient.spec.js
```diff
@@ -201,23 +215,28 @@ describe('WebClient', function () {
...
+
+            assert.lengthOf(parts.fields, 2);
+            assert.deepInclude(parts.fields, { fieldname: 'token', value: token });
+            assert.deepInclude(parts.fields, { fieldname: 'filename', value: 'train.jpg' });
           });
       });
 
-      // Reactivate this test once we find out if the workaround in the test case before is necessary
-      it('should log a warning when file is a Buffer and there is no filename', function () {
+      // TODO: re-enable this once we decide if we want to generate a random name or if we want to log a warning
```
aoberoi: an open point we should resolve before merging this.

aoberoi:
blocker: i just tried using this implementation with a ReadableStream from `fs.createReadStream()` and it failed with the following error:

```
   { ok: false,
     error: 'no_file_data',
     acceptedScopes: [ 'files:write:user', 'post' ] } }
```

aoberoi:
okay, i've learned an interesting thing. the following is true for `files.upload` (i haven't checked against other methods, but i will):

if in the multipart body, the part that contains the binary data does not specify a `filename` attribute in the header, the upload will fail. if another part contains a `filename` that will override whatever was specified in the `filename` attribute of the first part.

aoberoi:
and the same is true for `users.setPhoto`. that is that if no `filename` attribute is specified, it will fail. in this case, there's no need for a `filename` attribute, since the filename is generated by Slack.

aoberoi:
side note: love it when you implement a new feature with more deletions than additions 💯

src/WebClient.ts
```diff
@@ -656,6 +651,8 @@ export interface WebAPIHTTPError extends CodedError {
  * Helpers
  */
 
+const defaultFilename = 'Untitled';
```
Roach: Should we add a note to the docs for this?

aoberoi: heh, if we had generated docs for the individual methods at all, then i'd say yes.

PR #522, opened 2018-03-23
WebClient.apiCall() will detect some types of bad arguments

### LINKED ISSUE #521: Fail with a meaningful error when WebClient.apiCall()'s first argument is not a key-value object
### Description

This is useful when a user calls `apiCall()` (or any of the named method aliases) with a flat list of arguments accidentally. This is more likely to happen when a user is migrating from v3 or earlier.

aoberoi:
###  Summary

fixes #521

### DIFF
--- a/src/WebClient.spec.js
+++ b/src/WebClient.spec.js
@@ -100,6 +100,37 @@ describe('WebClient', function () {
       });
     });
 
+    describe('when called with bad options', function () {
+      it('should reject its Promise with TypeError', function (done) {
+        const results = [
+          this.client.apiCall('method', 4),
+          this.client.apiCall('method', 'a string'),
+          this.client.apiCall('method', false),
+        ];
+        const caughtErrors = results.map(r => {
+          assert(isPromise(r));
+          return r
+            .then(() => {
+              // if any of these promises resolve, this test fails
+              assert(false);
+            })
+            .catch((error) => {
+              // each promise should reject with the right kind of error
+              assert.instanceOf(error, TypeError);
+            });
+        });
+        Promise.all(caughtErrors)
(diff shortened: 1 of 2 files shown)


src/WebClient.spec.js
```diff
@@ -100,6 +100,37 @@ describe('WebClient', function () {
...
+        const results = [
+          this.client.apiCall('method', 4),
+          this.client.apiCall('method', 'a string'),
+          this.client.apiCall('method', false),
+        ];
+        const caughtErrors = results.map(r => {
+          assert(isPromise(r));
+          return r
+            .then(() => {
+              // if any of these promises resolve, this test failes
```
Roach: Fancy spelling

aoberoi: hahaha

src/WebClient.ts
```diff
@@ -121,7 +121,11 @@ export class WebClient extends EventEmitter {
 
     // The following thunk is the actual implementation for this method. It is wrapped so that it can be adapted for
     // different executions below.
-    const implementation = () => {
+    const implementation = async () => {
+
+      if (typeof options === 'string' || typeof options === 'number' || typeof options === 'boolean') {
```
Roach: Could you check for `is an object` rather than `is not x, or x, or x`?

aoberoi: `typeof null === 'object'` so that would pass this check. that being said, it passes this check right now too. 🤔

aoberoi: im afraid of weird javascript quirks, and this is more explicit regardless. `null`s would still pass this check, but `null` never has been a valid argument value, so i'm not too afraid of that case. i stand by this condition.

PR #528, opened 2018-04-11
Allow mrkdwn_in, callback_id fields in message attachments

DominikPalo:
###  Summary

A message attachment supports also the `mrkdwn_in` property, which is used to enable formatting on attachment fields (see https://api.slack.com/docs/message-formatting).

I added also the `callback_id` field used with interactive messages (see https://api.slack.com/docs/interactive-message-field-guide)


src/methods.ts
```diff
@@ -91,6 +91,9 @@ export interface MessageAttachment {
     type: string;
     text?: string;
   }[];
+  attachment_type?: 'default';
```
aoberoi: while i see that this is an accepted field, there are no values other than 'default', so i don't think it adds any value to type it (i'm not sure there will ever be any other possible values, and each additional property has a visual cost when your editor is trying to give you hints)

src/methods.ts
```diff
@@ -91,6 +91,9 @@ export interface MessageAttachment {
...
+  callback_id?: string;
+  mrkdwn_in?: ('pretext' | 'text' | 'fields')[];
```
aoberoi: this works and its right, but i was a little confused by the syntax. would you mind making a named string-based enum type and then setting this type to an array of that?

src/methods.ts
DominikPalo: thanks, I understand, but our UX team uses [Walkiebot](https://walkiebot.co/) for designing flows and messages. We also plan to fetch those messages designed by Walkiebot automatically to our integrations (using the Walkiebot API), but there is a problem - Walkiebot generates JSON/JS of designed attachments with `attachmet_type` field included (see the attached screenshot), therefore the output is not compatible with this SDK and we have to manually remove this field from every exported message. I can contact Walkiebot and ask them to remove this field, but it seems, they strictly stick to Slack API specs, so maybe as a first step is good to remove this unnecessary field from Slack API documentation.
walkiebot

src/methods.ts
DominikPalo: I would prefer to use a union type instead of enum, because declare it as an enum type, e.g.:
```
export enum MrkdwnIn {
  Pretext: 'pretext',
  Text: 'text',
  Fields: 'fields'
}

MessageAttachment {
  //...
  mrkdwn_in?: MrkdwnIn[];
}
```
it won't be possible to define mesages in this "plain" way
```
const attachment: MessageAttachment = {
  mrkdwn_in: ['pretext']
}
```
and forces a user to use this style:
```
const attachment: MessageAttachment = {
  mrkdwn_in: ['MrkdownIn.pretext']
}
```

aoberoi: 😮 i guess i never really gave any thought to the intricate differences between a union type and an enum. i always thought that enums were a "more powerful" union type because TypeScript would generate a symbol with the actual map, but didn't realize the compiler would complain when you use the literal values like in your example. [i tried it out here](http://www.typescriptlang.org/play/#src=enum%20MrkdwnInOne%20%7B%0D%0A%20%20Pretext%20%3D%20'pretext'%2C%0D%0A%20%20Text%20%3D%20'text'%2C%0D%0A%20%20Fields%20%3D%20'fields'%0D%0A%7D%0D%0A%0D%0Ainterface%20MessageAttachmentOne%20%7B%0D%0A%20%20%2F%2F...%0D%0A%20%20mrkdwn_in%3F%3A%20MrkdwnInOne%5B%5D%3B%0D%0A%7D%0D%0A%0D%0Aconst%20myAttachment%3A%20MessageAttachmentOne%20%3D%20%7B%0D%0A%20%20%20%20mrkdwn_in%3A%20%5BMrkdwnInOne.Pretext%5D%0D%0A%7D%0D%0A%0D%0Atype%20MrkdwnInTwo%20%3D%20'pretext'%20%7C%20'text'%20%7C%20'fields'%3B%0D%0A%0D%0Ainterface%20MessageAttachmentTwo%20%7B%0D%0A%20%20%2F%2F...%0D%0A%20%20mrkdwn_in%3F%3A%20MrkdwnInTwo%5B%5D%3B%0D%0A%7D%0D%0A%0D%0Aconst%20anotherAttachment%3A%20MessageAttachmentTwo%20%3D%20%7B%0D%0A%20%20%20%20mrkdwn_in%3A%20%5B'text'%2C%20'pretext'%5D%0D%0A%7D%3Bt).

in light of this fact, i agree that the union type is a better solution. thanks!

src/methods.ts
DominikPalo: thanks, we use TypeScript for all our projects - but I thought before that our attachments were incompatible with the `MessageAttachment` because of missing the `attachment_type` field, but as I see now, my assumption is not true (thanks to the `AuxiliaryArguments`). Our incompatibility comes from missing types in the nested field `MessageAttachment.actions` - the current type definition for `actions` contains only two fields:
```
actions?: {
    type: string;
    text?: string;
  }[];
```
but the attachment actions in our project contains additional fields. So I will prepare a separate PR also with complete definitions for the attachment actions.

brianeletype:
Does anyone have any idea when the next release will go out that has this update in it?  

Thanks

aoberoi:
@brianeletype sorry for the wait, we'd like to release the next version ASAP too. right now i'm holding on the investigation of #531 since its been observed in the wild by a few developers and it seems pretty critical. if we think fixing that is going to take longer a day or two, then we'll release without fixing it, definitely before the end of the week.

brianeletype:
Very timely and perfectly reasonable response.  Thank you very much.  I've put  in a work around for the time being.  Keep up the good work, @aoberoi

PR #530, opened 2018-04-12
Make the validate argument in the channels.join method optional

DominikPalo:
###  Summary

The `validate` argument in the `channels.join` method is optional (see https://api.slack.com/methods/channels.join), therefore it should be marked as optional also in the associated type.


aoberoi:
thanks for finding this inconsistency!

PR #543, opened 2018-04-27
Updated MessageAttachment inteface to contain better type for actions

brianeletype:
###  Summary

MessageAttachment.actions is incomplete and incorrect based on the field guide (https://api.slack.com/docs/interactive-message-field-guide).

* [X] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.

### DIFF
--- a/src/methods.ts
+++ b/src/methods.ts
@@ -87,12 +87,40 @@ export interface MessageAttachment {
   footer?: string;
   footer_icon?: string; // footer must be present
   ts?: string;
-  actions?: { // note: this is a very minimal definition of link buttons, interactive buttons, and message menus
-    type: string;
-    text?: string;
-  }[];
+  actions?: AttachmentAction[];
   callback_id?: string;
   mrkdwn_in?: ('pretext' | 'text' | 'fields')[];
+}
+
+export interface AttachmentAction {
+  id?: string;
+  confirm?: Confirmation;
+  data_source?: string;
+  min_query_length: number;
+  name: string;
+  options?: OptionField[];
+  option_groups?: {
+    text: string
+    options: OptionField[];
+  }[];
+  selected_options: OptionField[];
+  style?: string;
+  text: string;
+  type: string;
+  value?: string;
+}
+
+export interface OptionField {
+  description?: string;
+  text: string;
+  value: string;
+}
+
+export interface Confirmation {
+  dismiss_text?: string;
+  ok_text?: string;
+  text: string;
+  title?: string;
 }
 
 export interface LinkUnfurls {


aoberoi:
looks great to me!

src/methods.ts
```diff
@@ -87,14 +87,42 @@ export interface MessageAttachment {
...
+  actions?: AttachmentAction[];
   callback_id?: string;
   mrkdwn_in?: ('pretext' | 'text' | 'fields')[];
 }
 
+export interface AttachmentAction {
+  id?: string;
+  confirm?: Confirmation;
+  data_source?: string;
+  min_query_length: number;
```
DominikPalo: This field is optional, so I think it should be defined as `min_query_length?` instead of `min_query_length`

src/methods.ts
```diff
@@ -87,14 +87,42 @@ export interface MessageAttachment {
...
+  name: string;
+  options?: OptionField[];
+  option_groups?: {
+    text: string
+    options: OptionField[];
+  }[];
+  selected_options: OptionField[];
```
DominikPalo: This field is optional, so I think it should be defined as `selected_options?` instead of `selected_options `

src/methods.ts
```diff
@@ -87,14 +87,42 @@ export interface MessageAttachment {
...
+  style?: string;
+  text: string;
+  type: string;
+  value?: string;
+}
```
DominikPalo: Please, add also the optional field `url?: string` (used with link buttons, see https://api.slack.com/docs/message-attachments#link_buttons)

src/methods.ts
```diff
@@ -87,14 +87,42 @@ export interface MessageAttachment {
...
   callback_id?: string;
   mrkdwn_in?: ('pretext' | 'text' | 'fields')[];
 }
 
+export interface AttachmentAction {
+  id?: string;
+  confirm?: Confirmation;
+  data_source?: string;
+  min_query_length: number;
+  name: string;
```
DominikPalo: This field is optional, so I think it should be defined as `name?` instead of `name` (it's required only for standard actions, but it's not used with link buttons - see https://api.slack.com/docs/message-attachments#link_buttons)

aoberoi:
@DominikPalo these are all great suggestions. thanks for your in-depth review. if anyone has a PR to address these comments, i'm happy to merge.

brianeletype:
I agree.  I'll make those handful of updates tonight.

DominikPalo:
@brianeletype sorry, I already made it https://github.com/slackapi/node-slack-sdk/pull/551 :)

brianeletype:
Even better!

Issue #509, opened 2018-03-15
Strong type api method responses


bilby91:
### Description

Hello, I'm continuing my migration for `botbuilder-slack` to the new 4.x release. Today I realized that the responses from method calls are not type dependent on the method invocation, response type for every api call is `Promise<WebAPICallResult>`. Is there any plan to support strong typed responses ? I can work on this with my previous typings.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [X] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [X] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [X] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [X] I've searched for any related issues and avoided creating a duplicate issue.

aoberoi:
@bilby91 I’m glad to see your migration continue!

We’re having a discussion about a few different ways to get richer types in another issue. Specifically, I mentioned exactly what you’re asking for in point (4) in [this comment](https://github.com/slackapi/node-slack-sdk/issues/496#comment-372803268). I’d really appreciate your participation in that discussion so we can potentially totally come up with the steps to get there and divide the work.

neekolas:
I was running into this issue as well. I got fed up trying to shut the Typescript compiler up every time I made an API call and wound up crawling the [Object Type](https://api.slack.com/types) definition pages, using the examples to generate a JSON Schema, and using the schema to build Typescript runtime validators and interfaces.

I am doing the same thing for the events API, which works really nicely, but the web responses have been giving me trouble. Isssue is that there aren't enough samples for me to infer which fields are reliably going to be there and which ones aren't. Here's an example of the generated schema:
_Not all of these fields should really be required_
```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "$ref": "#/definitions/File",
  "definitions": {
    "File": {
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "channels": { "type": "array", "items": { "type": "string" } },
        "comments_count": { "type": "integer" },
        "created": { "type": "integer" },
        "display_as_bot": { "type": "boolean" },
        "edit_link": { "type": "string" },
        "editable": { "type": "boolean" },
[...]

PR #563, opened 2018-05-16
Add first pass at response typings

neekolas:
###  Summary

Adds auto-generated response types for all API methods. Typings are generated based on the public JSON schema, so it is only as complete as that schema is. Addresses #509. Does not make any adjustments to runtime code.

Not entirely sure how strict you want to be on the cases where there is incomplete data. Master currently shows a typescript error any time you try and access an API response property directly, which I took as a hint and erred on the side of strictness. I cannot find any cases where Typescript throws an error or warning on my branch that doesn't also get a warning or error in master.

It wouldn't let me commit without resolving a few unrelated tslint errors. I went through your .tslint.json and you guys have `no-redundant-jsdoc` turned on - so I think I was fixing a lint error and not creating one - but it's entirely possible there's something mismatched in my local environment to yours that is triggering it for me.

### DIFF
--- a/src/methods.ts
+++ b/src/methods.ts
@@ -1,5 +1,9 @@
 import { Stream } from 'stream';
-import { WebAPICallOptions, WebAPIResultCallback, WebAPICallResult } from './WebClient';
+import {
+  WebAPICallOptions,
+  WebAPIResultCallback,
+  WebAPICallResult,
+} from './WebClient';
 
 // NOTE: could create a named type alias like data types like `SlackUserID: string`
 // NOTE: not clear if these interfaces should be exported at the top-level
@@ -7,10 +11,16 @@ import { WebAPICallOptions, WebAPIResultCallback, WebAPICallResult } from './Web
 /**
  * Generic method definition
  */
-export default interface Method<MethodArguments extends WebAPICallOptions> {
+export default interface Method<
+  MethodArguments extends WebAPICallOptions,
+  ResponseType extends WebAPICallResult
+> {
   // TODO: can we create a relationship between MethodArguments and a MethodResult type?
-  (options?: MethodArguments & AuxiliaryArguments): Promise<WebAPICallResult>;
-  (options: MethodArguments & AuxiliaryArguments, callback: WebAPIResultCallback): void;
+  (options?: MethodArguments & AuxiliaryArguments): Promise<ResponseType>;
+  (
+    options: MethodArguments & AuxiliaryArguments,
+    callback: WebAPIResultCallback,
+  ): void;
 }
 
 /*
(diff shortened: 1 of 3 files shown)


src/methods.ts
```diff
@@ -1,16 +1,26 @@
...
+  MethodArguments extends WebAPICallOptions,
+  ResponseType extends WebAPICallResult
+> {
   // TODO: can we create a relationship between MethodArguments and a MethodResult type?
-  (options?: MethodArguments & AuxiliaryArguments): Promise<WebAPICallResult>;
-  (options: MethodArguments & AuxiliaryArguments, callback: WebAPIResultCallback): void;
+  (options?: MethodArguments & AuxiliaryArguments): Promise<ResponseType>;
+  (
+    options: MethodArguments & AuxiliaryArguments,
+    callback: WebAPIResultCallback,
```
aoberoi: maybe `WebAPIResultCallback` needs a generic type constraint of `ResponseType`?

src/responses.ts
```diff
@@ -0,0 +1,2430 @@
+export interface Im {
```
aoberoi: could some of these interfaces be unified with interfaces defined in `src/methods.ts`? i know that these define the API outputs, and those the API inputs. unifying them could come with the potential win that similar structures only need to be updated in one place. keeping them separate comes with the win of decoupling. just food for thought at this point.

aoberoi:
whoa, thanks for this contribution! i'm so glad you're willing to help further the goals of #509.

i think this approach is very nice. i like that method types and response types don't directly depend on one another, and that the WebClient actually pairs them up. 

you said this output was generated, would you mind sharing the script? i think it would be useful to keep in the repo to "sync" with the OpenAPI spec every once in a while. was there much manual manipulation of the output after the script was run? i'm really interested in understanding the process.

reviewing a change this big will definitely take some time. i'd like to build the package and try using it locally to see what the typescript language service has to offer with all these awesome types to work with! i don't expect to have time to do this in the next week or so. i encourage other community members to check out this patch and leave some feedback.

since i see this as a change that might take some iteration before landing, and since potentially more people may like to get involved, perhaps it makes sense to retarget this PR onto a new feature branch (e.g. `feat-webclient-response-types`). that way this PR could land on that branch and others could make additional PRs to continue to add polish. @neekolas wdyt?

---

i saw that you tried to undo the style changes introduced by prettier, thanks. you didn't get all of them, and parsing out the real change is adding time for me to review 🤷‍♂️. the majority of the style changes i'd probably want to revert before this lands on master. some were great though, and i think we can contribute some changes to our `tslint` rules to enforce them! happy to have that discussion in a separate PR.

aoberoi:
@clavin i'd love to hear your thoughts, if you have a chance.

9renpoto:
@neekolas Hi. If you do not have a problem, could you please rebase it ? 👍 

refs: https://github.com/neekolas/node-slack-sdk/compare/master...9renpoto:master

Issue #451, opened 2018-01-18
make webclient rate-limit retries configurable


rodrigo4244:
### Description

I read issue #305 that explained that giving a `retryConfig` object for `WebClient` it would be possible to disable the retry operation but it actually isn't (I explain properly on steps to reproduce).
On your docs is: https://slackapi.github.io/node-slack-sdk/web_api#changing-the-retry-configuration.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Reproducible in:

`@slack/client` version: 3.15.0

node version: 8.9.3

OS version(s): Ubuntu 16.04 LTS

#### Steps to reproduce:

1. Run the following script until you get `Rate limited` errors. (make sure you're using a token from a workspace which have a good amount of channels, otherwise it could take long)
```js
const { WebClient } = require('@slack/client')

const SLACK_USER_TOKEN='<token here>'

const slack = new WebClient(SLACK_USER_TOKEN, { retryConfig: { retries: 0 }, maxRequestConcurrency: Infinity })

for (let i = 1; i <= 1000; i++) {
  slack.channels.list()
    .catch(e => console.log('this never gets called'))
}
```

#### Expected result:

I expected that `@slack/client` would throw an error so I can throttle API calls myself.

#### Actual result:

`@slack/client` is automatically retrying.

#### Attachments:

gunar:
I got it to stop retrying with `{ retryConfig: { retries: 0, forever: false } }`, but the promise won't reject.

gunar:
It seems `handleRateLimitResponse` is a special case and doesn't really use the library `retry`.

aoberoi:
thanks for point that out @gunar. you're right, and the docs don't specify whether the retry configuration should apply for rate-limited requests. the key information is buried in the code comments:

[...]

PR #570, opened 2018-05-31
Move rate limiting/retrying to error catching, add tests for rate limit handling

ggruiz:
###  Summary

* Rate limiting/retrying being part of the `apiCall()`'s `.then()` area stopped it from actually running, so I moved it to the `.catch()` portion to make it work.
* Added relevant tests.

### DIFF
--- a/src/WebClient.spec.js
+++ b/src/WebClient.spec.js
@@ -6,6 +6,7 @@ const { assert } = require('chai');
 const { WebClient } = require('./WebClient');
 const { LogLevel } = require('./logger');
 const { addAppMetadata } = require('./util');
+const rapidRetryPolicy = require('./retry-policies').rapidRetryPolicy;
 const CaptureStdout = require('capture-stdout');
 const isPromise = require('p-is-promise');
 const nock = require('nock');
@@ -13,7 +14,6 @@ const Busboy = require('busboy');
 const sinon = require('sinon');
 
 const token = 'xoxa-faketoken';
-const fastRetriesForTest = { minTimeout: 0, maxTimeout: 1 };
 
 describe('WebClient', function () {
 
@@ -72,7 +72,7 @@ describe('WebClient', function () {
 
   describe('apiCall()', function () {
     beforeEach(function () {
-      this.client = new WebClient(token, { retryConfig: fastRetriesForTest });
+      this.client = new WebClient(token, { retryConfig: rapidRetryPolicy });
     });
 
     describe('when making a successful call', function () {
@@ -339,7 +339,7 @@ describe('WebClient', function () {
           .post(/api/)
           .reply(200, { ok: true });
         // NOTE: appMetaData is only evalued on client construction, so we cannot use the client already created
-        const client = new WebClient(token, { retryConfig: fastRetriesForTest });
+        const client = new WebClient(token, { retryConfig: rapidRetryPolicy });
         return client.apiCall('method')
           .then(() => {
             scope.done();
@@ -349,7 +349,7 @@ describe('WebClient', function () {
   });
 
   describe('apiCall() - without a token', function () {
-    const client = new WebClient(undefined, { retryConfig: fastRetriesForTest });
+    const client = new WebClient(undefined, { retryConfig: rapidRetryPolicy });
 
     const scope = nock('https://slack.com')
       // NOTE: this could create false negatives if the serialization order changes (it shouldn't matter)
@@ -365,7 +365,7 @@ describe('WebClient', function () {
 
   describe('named method aliases (facets)', function () {
     beforeEach(function () {
-      this.client = new WebClient(token, { retryConfig: fastRetriesForTest });
+      this.client = new WebClient(token, { retryConfig: rapidRetryPolicy });
     });
     it('should properly mount methods as functions', function () {
       // This test doesn't exhaustively check all the method aliases, it just tries a couple.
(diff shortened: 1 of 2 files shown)


src/WebClient.spec.js
```diff
@@ -500,15 +501,95 @@ describe('WebClient', function () {
   });
 
   describe('has rate limit handling', function () {
-    it('should expose retry headers in the response');
+    it('should expose retry headers in the response', function() {
+      const scope = nock('https://slack.com')
+        .post(/api/)
+        .reply(200, { ok: true }, {
```
aoberoi: hmmm... not quite. we would only ever see `retry-after` when the response has a status code of 429. i'm not 100% sure what the body would contain, but `{ ok: true }` seems wrong to me, too.

src/WebClient.spec.js
```diff
@@ -500,15 +501,95 @@ describe('WebClient', function () {
...
 
     describe('when a request fails due to rate-limiting', function () {
       // NOTE: is this retrying configurable with the retry policy? is it subject to the request concurrency?
-      it('should automatically retry the request after the specified timeout');
-      it('should pause the remaining requests in queue');
-      it('should emit a rate_limited event on the client');
+      it('should automatically retry the request after the specified timeout', function() {
+        const scope = nock('https://slack.com')
+          .post(/api/)
+          .reply(429, { ok: true }, {
```
aoberoi: probably not `{ ok: true }`

src/WebClient.spec.js
```diff
@@ -500,15 +501,95 @@ describe('WebClient', function () {
...
+            'retry-after': 1
+          })
+          .post(/api/)
+          .reply(200, { ok: true });
+        const client = new WebClient(token, { retryConfig: { retries: 1 } });
+        return client.apiCall('method')
+          .then((resp) => {
+            assert.propertyVal(resp, 'ok', true);;
```
aoberoi: how can we verify that the request actually occurred after the specified timeout (1 second)?

src/WebClient.spec.js
```diff
@@ -500,15 +501,95 @@ describe('WebClient', function () {
...
+            'retry-after': 1
+          })
+          .post(/api/)
+          .reply(429, { ok: true }, {
+            'retry-after': 0
+          })
+          .post(/api/)
+          .reply(200, { ok: true });
+        const client = new WebClient(token, { retryConfig: { retries: 3 } });
+        return client.apiCall('method')
```
aoberoi: i don't think this reflects the intention of the test case. the idea is: given a WebClient with a request concurrency of 2 (just to keep things simple), you fire three requests concurrently (without waiting for the promises to resolve). the first request gets rate-limited, for a specified amount of time, let's say 1 second. the second request would have already been triggered (the concurrency allows for 2 requests to be in flight at the same time) and let's say it completes very quickly (~10ms), but the third request should not even start until after the 1 second timeout from the first request has elapsed. does this make sense?

src/WebClient.spec.js
```diff
@@ -500,15 +501,95 @@ describe('WebClient', function () {
...
+        })
+        .post(/api/)
+        .reply(200, { ok: true });
+      const client = new WebClient(token, { retryConfig: { retries: 0 } });
+      return client.apiCall('method')
+        .then(() => {
+          assert.fail(0, 1, 'Rate Limit retry not turned off');
+        })
+        .catch((error) => {
+          assert.isFalse(scope.isDone());
```
aoberoi: it makes sense that you're setting up responses that are never meant to trigger in the scope, and therefore you check `scope.isDone()` rather than calling `scope.done()`. do you think we should also be cleaning up these test cases in a teardown method with `scope.cleanAll()`?

src/WebClient.ts
```diff
@@ -144,46 +144,37 @@ export class WebClient extends EventEmitter {
...
               throw requestErrorWithOriginal(error);
             } else if (error.name === 'ReadError') {
               throw readErrorWithOriginal(error);
             } else if (error.name === 'HTTPError') {
+              // Special case: retry if 429
+              const result = this.buildResult(error.response);
+              if (error.statusCode === 429) {
+                const retryAfterMs = result.retryAfter !== undefined ? result.retryAfter : (60 * 1000);
+                this.emit('rate_limited', retryAfterMs / 1000);
+                if (this.retryConfig.retries !== 0) {
```
aoberoi: ah-hah. i think you found an issue that actually resides within the tests. the case `should allow rate limit triggered retries to be turned off` was noted assuming that's functionality we intend to support. but some time after that note, it was made clear that the `retryConfig` has nothing to do with the rate-limit control built into this object. see: https://github.com/slackapi/node-slack-sdk/issues/451

specifically, i think we need to remove this condition and think about that problem more holistically. right now you check a very specific `retries` property of the `retryConfig`, but its actually a complex object that's meant to be interpreted by a more sophisticated library like `retry`. long story short: we need to solve for that test case in a different way, and we should probably remove its implementation for now.

we can start work on resolving #451, but i think that should happen in a separate PR.

src/WebClient.spec.js
ggruiz: my bad; i was supposed to change that to 429, but forgot about it. Thanks!

src/WebClient.spec.js
ggruiz: Will change this and every instance thereof

src/WebClient.spec.js
ggruiz: note: time specific, not existence

src/WebClient.spec.js
ggruiz: i get it, will change to reflect this

PR #573, opened 2018-06-04
catch errors on platform

ggruiz:
###  Summary

This commit is solely to catch platform errors upon receiving a 200 request that is actually not ok. Test is also included.

### DIFF
--- a/src/WebClient.spec.js
+++ b/src/WebClient.spec.js
@@ -191,15 +191,16 @@ describe('WebClient', function () {
       });
     });
 
-    it('should return platform error if 200 request is not ok', function () {
+    it.only('should fail with platform errors when the API response is an error', function () {
       const scope = nock('https://slack.com')
         .post(/api/)
-        .reply(200, { ok: false });
+        .reply(200, { ok: false, error: 'bad error' });
       const client = new WebClient(token);
       return client.apiCall('method')
         .catch((error) => {
           assert.propertyVal(error, 'code', 'slackclient_platform_error');
-          assert.propertyVal(error.data, 'ok', false);
+          assert.nestedPropertyVal(error, 'data.ok', false);
+          assert.nestedPropertyVal(error, 'data.error', 'bad error');
           scope.done();
         });
     });
--- a/src/WebClient.spec.js
+++ b/src/WebClient.spec.js
@@ -191,7 +191,7 @@ describe('WebClient', function () {
       });
     });
 
-    it.only('should fail with platform errors when the API response is an error', function () {
+    it('should fail with platform errors when the API response is an error', function () {
       const scope = nock('https://slack.com')
         .post(/api/)
         .reply(200, { ok: false, error: 'bad error' });
--- a/src/WebClient.spec.js
+++ b/src/WebClient.spec.js
@@ -191,6 +191,19 @@ describe('WebClient', function () {
       });
     });
 
+    it('should return platform error if 200 request is not ok', function () {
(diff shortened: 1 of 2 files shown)


src/WebClient.spec.js
```diff
@@ -191,6 +191,19 @@ describe('WebClient', function () {
       });
     });
 
+    it('should return platform error if 200 request is not ok', function () {
```
aoberoi: can we rephrase the description of this test case a bit? how about `should fail with platform errors when the API response is an error`?

src/WebClient.spec.js
```diff
@@ -191,6 +191,19 @@ describe('WebClient', function () {
...
+      const scope = nock('https://slack.com')
+        .post(/api/)
+        .reply(200, { ok: false });
+      const client = new WebClient(token);
+      return client.apiCall('method')
+        .catch((error) => {
+          assert.propertyVal(error, 'code', 'slackclient_platform_error');
+          assert.propertyVal(error.data, 'ok', false);
```
aoberoi: this works since you have indirectly checked the truthiness of `error` in the preceding line. but i think this ca be improved as `assert.nestedPropertyVal(error, 'data.ok', false);` since then the order of the assertions wouldn't be significant.

src/WebClient.spec.js
```diff
@@ -191,6 +191,19 @@ describe('WebClient', function () {
       });
     });
 
+    it('should return platform error if 200 request is not ok', function () {
+      const scope = nock('https://slack.com')
+        .post(/api/)
+        .reply(200, { ok: false });
```
aoberoi: this response isn't sufficient to test the code. there should be an `error` property in the response, the way the actual API would provide one.

src/WebClient.spec.js
```diff
@@ -191,6 +191,19 @@ describe('WebClient', function () {
...
+      const client = new WebClient(token);
+      return client.apiCall('method')
+        .catch((error) => {
+          assert.propertyVal(error, 'code', 'slackclient_platform_error');
+          assert.propertyVal(error.data, 'ok', false);
+          scope.done();
```
aoberoi: we're missing an assertion on `error.data.error`, which should be equal to the `error` property in the response (as commented above).

src/WebClient.spec.js
ggruiz: rephrased

src/WebClient.spec.js
ggruiz: taken care of in latest commit; changed line 202 to `assert.nestedPropertyVal`

src/WebClient.spec.js
ggruiz: taken care of in latest commit; added the property `error: 'bad error'` to line 197's test response object

src/WebClient.spec.js
ggruiz: taken care of in latest commit; along with `error: 'bad error'` to line 197, also added `assert.nestedPropertyVal(error, 'data.error', 'bad error')` to inner function of the `catch`

src/WebClient.spec.js
```diff
@@ -191,15 +191,16 @@ describe('WebClient', function () {
       });
     });
 
-    it('should return platform error if 200 request is not ok', function () {
+    it.only('should fail with platform errors when the API response is an error', function () {
```
aoberoi: ah, lets get rid of this `.only()`

ggruiz: oops

PR #588, opened 2018-06-29
Fix Dialog.selected_options type

harveyr:
###  Summary

Fixes the type of `Dialog.selected_options`. It now matches `AttachmentAction.selected_options` and what's described in [these docs](https://api.slack.com/dialogs#select_default_values).

### DIFF
--- a/src/methods.ts
+++ b/src/methods.ts
@@ -60,7 +60,7 @@ export interface Dialog {
     subtype?: 'email' | 'number' | 'tel' | 'url';
     // type `select`:
     data_source?: 'users' | 'channels' | 'conversations' | 'external';
-    selected_options?: string;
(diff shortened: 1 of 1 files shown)


src/methods.ts
```diff
@@ -60,7 +60,7 @@ export interface Dialog {
     subtype?: 'email' | 'number' | 'tel' | 'url';
     // type `select`:
     data_source?: 'users' | 'channels' | 'conversations' | 'external';
-    selected_options?: string;
+    selected_options?: OptionField[];
```
aoberoi: `OptionField[]` isn't actually right, since it defines the `text` property which cannot be used here. What we actually want is `SelectOption[]`

harveyr: 👍 fixed!

Issue #496, opened 2018-03-13
Extra strict method arguments


clavin:
### Description

On the topic of [method arguments interfaces](https://github.com/slackapi/node-slack-sdk/pull/487#issuecomment-372481734), thoughts on possibly making those interfaces *extra* strong by attempting to ensure that certain combinations of arguments are valid?

For example, one could say the following call to `chat.postMessage` is *technically* incorrect:

```ts
web.chat.postMessage({
    text: 'words, words, words',
    as_user: true,
    icon_url: 'https://lorempixel.com/48/50'
});
```

Because passing `as_user` as `true` means `icon_url` is ignored. With the right type signature, the above example can throw an error at static analysis.

This can be demonstrated in a smaller case. Say you have an object named `shape` and it either has the property `size` or `radius`, but not both. This this "exclusive or" operation [comes at the cost of verbosity](https://github.com/Microsoft/TypeScript/issues/14094) in TypeScript:

```ts
type Shape = { size: number; radius?: never } | { radius: number; size?: never };

const shape1: Shape = { // error!
    size: 12,
    radius: 13
};

const shape2: Shape = {}; // error!

const shape3: Shape = { // ok
    size: 12
};

const shape4: Shape = { // ok
    radius: 13
};
```

Going back to the example from before, the implementation would look something like this:

```ts
type ChatPostMessageArguments =
    {
        text: string;
        // ...
    } & (
        {
            as_user?: boolean;

            username?: never;
            icon_emoji?: never;
            icon_url?: never;
        } | (
            {
                username?: string;

                as_user?: never;
            } & (
                {
                    icon_emoji?: string;

                    icon_url?: never;
                } | {
                    icon_url?: string;

                    icon_emoji?: never;
                }
            )
        )
    );
```

The solution is mighty verbose, but the result is that method arguments can be statically checked to conform to the right shape:

```
Argument of type '{ text: string; as_user: boolean; icon_url: string; }' is
not assignable to parameter of type 'ChatPostMessageArguments'.
```

---

Obviously, this is a lot of work to do, but I feel like it might just be worth it (or, at least, a lot of fun).

Even more obvious, doing this would end up with bugs where correct combinations turn out to throw errors.

To mitigate this, the definition of `Method` at the top of `method.ts` could be altered so an optional type argument could be supplied to methods to be automatically unioned for method arguments, allowing callers to assert that their arguments are right by supplying `any` as the argument:

```ts
[...]

PR #596, opened 2018-07-25
Auto pagination

### LINKED ISSUE #343: Adopt pagination standard with a default page size for methods which support it

#### Reproducible in:
[hubot-slack](/slackapi/hubot-slack) version: latest
OS version(s): any
Device(s): any

#### Steps to reproduce:

1. Run users.list against an organization with thousands of users
2. See users.list throw a 500 http error
3. See users.list timeout without handling the error

>hubot-slack loads the entire users list into memory on load which seems insane to me when our organization has thousands of users and my bot is only in one or two channels with maybe 20 users. A user list by channel that loads on channel join would be way better.

### LINKED ISSUE #451: make webclient rate-limit retries configurable

aoberoi:
###  Summary

The platform is expanding the set of methods which support the shiniest of our pagination specifications, cursor-based pagination!

This change allows apps who previously made API calls to methods (that support cursor pagination) without any explicit pagination arguments to transparently traverse pages. This solves two major problems:
1. Previously, if the list was too long to return in a reasonable amount of time, these requests may result in server errors (500 response code).
2. Some methods started to return partial lists (with default `limit` values) in situations where the client didn't provide pagination arguments. This would create a subtle bug, but now these apps should get the entire list.

This change also helps apps prepare for a future where pagination becomes mandatory in certain situations.

Fixes #343 

**Please review and provide feedback**

At this point, my biggest concern is that this is a _lot_ of new code, for what amounts to functionality we still strongly recommend developers do not use. More specifically, we still want developers to paginate requests themselves, and not to depend on the automatic pagination. Our reason is that this will result in more performant and optimized code. So, planting all this code is very intentionally a band-aid. My proposal for how we'd address this concern is to merge this now, and in  the next major version to remove automatic pagination; moving that code into an add-on that can be used in conjunction with the baseline package. That add-on would be a shim that's only for backwards compatibility, but not encouraged.

Another concern is now the `cursorPaginationEnabledMethods` in `methods.ts` value is starting to store metadata about API responses, but outside of the holistic design we're hoping to arrive in #496 and #509.

## TODO

- [x] Sync metadata with the latest round of cursor-pagination supporting methods
- [x] Implement tests and get acceptable coverage
- [x] ~Add configuration for rate-limit retires (simple boolean) #451~ This can be done in a separate, follow-up PR, which will make it easier to review.
- [x] Handle general httpError conditions (like 500 responses)
- [x] Audit error types and codes
- [x] Turn default page size into a configuration option
- [ ] Document the new feature(s)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -11,7 +11,6 @@ export function noop(): void { } // tslint:disable-line:no-empty
 
 /**
  * Replaces occurences of '/' with ':' in a string, since '/' is meaningful inside User-Agent strings as a separator.
- * @param s
  */
 function replaceSlashes(s: string): string {
   return s.replace('/', ':');
@@ -43,11 +42,11 @@ export function getUserAgent(): string {
 
 // TODO: make initialValue optional (overloads or conditional types?)
 export async function awaitAndReduce<T, U>(iterable: AsyncIterable<T>,
-                                           callbackfn: (previousValue: U, currentValue: T) => Promise<U>,
+                                           callbackfn: (previousValue: U, currentValue: T) => U,
                                            initialValue: U): Promise<U> {
   let accumulator = initialValue;
   for await (const value of iterable) {
-    accumulator = await callbackfn(accumulator, value);
+    accumulator = callbackfn(accumulator, value);
   }
   return accumulator;
 }
--- a/src/util.ts
+++ b/src/util.ts
@@ -42,6 +42,17 @@ export function getUserAgent(): string {
   return ((appIdentifier.length > 0) ? `${appIdentifier} ` : '') + baseUserAgent;
 }
 
+// TODO: make initialValue optional (overloads or conditional types?)
+export async function awaitAndReduce<T, U>(iterable: AsyncIterable<T>,
+                                           callbackfn: (previousValue: U, currentValue: T) => Promise<U>,
+                                           initialValue: U): Promise<U> {
+  let accumulator = initialValue;
+  for await (const value of iterable) {
+  accumulator = await callbackfn(accumulator, value);
+}
+  return accumulator;
+}
+
 /**
  * The following is a polyfill of Node >= 8.2.0's util.callbackify method. The source is copied (with some
  * modification) from:
--- a/src/util.ts
+++ b/src/util.ts
@@ -25,7 +25,6 @@ const appMetadata: { [key: string]: string } = {};
 
 /**
  * Appends the app metadata into the User-Agent value
- * @param appMetadata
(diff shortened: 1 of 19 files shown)


src/util.ts
```diff
@@ -42,6 +40,34 @@ export function getUserAgent(): string {
...
+/**
+ * Reduce an asynchronous iterable into a single value.
+ * @param iterable the async iterable to be reduced
+ * @param callbackfn a function that implements one step of the reduction
+ * @param initialValue the initial value for the accumulator
+ */
+export async function awaitAndReduce<T, U>(iterable: AsyncIterable<T>,
+                                           callbackfn: (previousValue: U, currentValue: T) => U,
+                                           initialValue: U): Promise<U> {
+  // TODO: make initialValue optional (overloads or conditional types?)
```
shanedewael: 🤔

aoberoi: meh, i think i'll leave that comment in there.

right now we don't need a version of this function that doesn't take an `initialValue` because i'm manually passing `{} as WebAPICallResult` into it. but, since this lives in `util.ts`, it would be more reusable if it didn't need an initial value, the way that [Array.prototype.reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) also doesn't require an initial value.

PR #605, opened 2018-08-07
loosens type constraint on @types/node

### LINKED ISSUE #582: Make slack-sdk compatible with node 10
### Description

We're trying to use slack-sdk with projects which uses newer version of `@types/node` package (> 10.0.x). Slack is depending on ^9.4.7 which causes dependency collision and prevents us from successfully building package. Is it possible to make slack-sdk `@types/node` version agnostic, for example using `@types/node: '*'`?

- [ ] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

--

aoberoi:
###  Summary

When a user of this package wants to install a newer version of `@types/node` into their own application than the version specified in this package, bad things can happen (see #582). By loosening this constraint, it allows npm to resolve to a single version that works for the whole project. The risk this PR introduces is that future versions of node may specify new API that doesn't exist in older versions, so we have to be careful to only depend on node types which are identical across versions, or we have to offer implement a polyfill with its own type described (such as with `src/util.ts` `callbackify()`).

Fixes #582

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@types/got": "^7.1.7",
     "@types/is-stream": "^1.1.0",
     "@types/loglevel": "^1.5.3",
-    "@types/node": "^9.4.7",
+    "@types/node": ">=6.0.0",
     "@types/p-cancelable": "^0.3.0",
     "@types/p-queue": "^2.3.1",
     "@types/p-retry": "^1.0.1",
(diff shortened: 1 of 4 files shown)


aoberoi:
i don't know why codecov isn't reporting status on the latest commit. the coverage should be no diff still (we didn't add any lines of executable code). i'm overriding that status check to merge.

PR #609, opened 2018-08-09
implement actor enabled methods

### LINKED ISSUE #608: Working on behalf of users in workspace apps
### Description

Slack's workspace apps [now support invoking methods on behalf of a specified user](https://api.slack.com/docs/working-for-users).

The `WebClient` class needs to implement support for this new style of calling methods. The following is a proposal for changes to the API for that purpose.

1. Define a new `ActorEnabled` interface which defines an optional property `actor?: string`.
2. Intersect all method argument types for methods that support this feature with `ActorEnabled`.
3. In the implementation of `WebClient#apiCall()`, detect the presence of an `options.actor` property, and when present use the value in request header `X-Slack-User`.

### Open questions:

*  Can we think of a better name than `actor`? One constraint is that we don't want to use a name that might otherwise be used, or used in the future, as a method argument name. It also should communicate that the app is acting on behalf of this user. `actor` is mostly okay, but it isn't perfect because truly the app is the actor, and its acting **as** the user. So perhaps `actingAs`?

aoberoi:
###  Summary

Implements an additional method argument `actor`, which is used to fill the `X-Slack-User` request header on API calls.

Fixes #608

### DIFF
--- a/docs/_pages/web_client.md
+++ b/docs/_pages/web_client.md
@@ -9,6 +9,7 @@ headings:
     - title: Adding attachments to a message
     - title: Uploading a file
     - title: Getting a list of channels
+    - title: Calling methods on behalf of users
     - title: Using a callback instead of a Promise
     - title: Changing the retry configuration
     - title: Changing the request concurrency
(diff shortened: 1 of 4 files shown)


aoberoi:
updating this PR to use the preferred `on_behalf_of` option name, and adding some docs.

Issue #610, opened 2018-08-09
RTM State Machine never emits `failure` event


clavin:
### Description

Due to some [fuzzy Finity logic](https://github.com/slackapi/node-slack-sdk/blob/842c7a78ebcfc230b341f81719181f80f5817926/src/RTMClient.ts#L134), the `failure` event is never emitted within the RTM state machine when failing to reconnect.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Reproducible in:

`@slack/client` version: `v4.3.1`

node version: `v10.7.0`

OS version(s): macOS 10.14 Beta

#### Steps to reproduce:

```js
const { RTMClient, LogLevel } = require('@slack/client');
const rtm = new RTMClient('fake token', {
  logLevel: LogLevel.DEBUG
});

rtm.connect();
// !!! debug logs never show a transition to "disconnected"
```

#### Expected result:

The RTM client should transition to the `disconnected` state on failure to connect.

#### Actual result:

The `failure` event is never emitted from the submachine to the parent state machine, and thus the state never reaches `disconnected`

#### Attachments:

["Failure is not an option" -Gene Kranz](https://www.museumofflightstore.org/media/catalog/product/cache/1/image/9df78eab33525d08d6e5fb8d27136e95/S/T/STAT206247.jpg)

aoberoi:
Thanks. I was able to reproduce this one (I don't see the `transitioning...` log statement as we would expect).

PR #611, opened 2018-08-09
Add Promises to RTMClient `start` and `disconnect`

### LINKED ISSUE #198: Return a promise from RtmClient.start
``` js
const rtm = new RtmClient(token);
rtm.start()
  .then(data => {
    console.log('Authenticated!', data);
  });
```

Could be equivalent to:

``` js
new Promise((resolve, reject) => {
  rtm.on(RTM.AUTHENTICATED, resolve);
  rtm.on(RTM.UNABLE_TO_RTM_START, reject);
  rtm.start();
});
```

or even wait until the websocket connection is started too, not sure.

clavin:
###  Summary

Fixes #198 and #610:

- Fixes the RTM `connecting` submachine to transition to a `failed` state, which emits a `failure` event on the parent machine, allowing for the `disconnected` state to be reached from a failure to reconnect (tl;dr: it works 👍)
- Changes the return types of `RTMClient#start` to `Promise<WebAPICallResult>` and `RTMClient#disconnect` to `Promise<void>`

### Requirements

### DIFF
--- a/src/RTMClient.ts
+++ b/src/RTMClient.ts
@@ -235,9 +235,13 @@ export class RTMClient extends EventEmitter {
       .global()
         .onStateEnter((state, context) => {
           this.logger.debug(`transitioning to state: ${state}`);
-          // Emits events: `disconnected`, `connecting`, `connected`, 'disconnecting', 'reconnecting'
-          // event payload is anything related to the event, i.e. an error on disconnect
-          this.emit(state, context.eventPayload);
+          if (state === 'disconnected') {
+            // Emits a `disconnected` event with a possible error object (might be undefined)
+            this.emit(state, context.eventPayload);
+          } else {
+            // Emits events: `connecting`, `connected`, `disconnecting`, `reconnecting`
+            this.emit(state);
+          }
         })
     .getConfig();
 
--- a/src/RTMClient.ts
+++ b/src/RTMClient.ts
@@ -131,10 +131,7 @@ export class RTMClient extends EventEmitter {
 
                   return this.autoReconnect && isRecoverable;
                 })
-                .ignore().withAction(() => {
-                  // dispatch 'failure' on parent machine to transition out of this submachine's states
-                  this.stateMachine.handle('failure');
-                })
+                .transitionTo('failed')
           .state('authenticated')
             .onEnter((_state, context) => {
               this.authenticated = true;
@@ -145,6 +142,11 @@ export class RTMClient extends EventEmitter {
             })
             .on('websocket open').transitionTo('handshaking')
           .state('handshaking') // a state in which to wait until the 'server hello' event
+          .state('failed')
+            .onEnter((_state, context) => {
+              // dispatch 'failure' on parent machine to transition out of this submachine's states
+              this.stateMachine.handle('failure', context.error);
+            })
           .global()
             .onStateEnter((state) => {
               this.logger.debug(`transitioning to state: connecting:${state}`);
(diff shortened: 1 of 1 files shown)


src/RTMClient.ts
```diff
@@ -365,18 +367,40 @@ export class RTMClient extends EventEmitter {
...
+          reject(err);
+        } else {
+          resolve();
+        }
+      });
 
-    // delegate behavior to state machine
-    this.stateMachine.handle('explicit disconnect');
+      // delegate behavior to state machine
+      this.stateMachine.handle('explicit disconnect');
```
aoberoi: any reason to kick this event off inside the returned `Promise`'s constructor, as opposed to the line before the return statement?

src/RTMClient.ts
```diff
@@ -231,10 +233,11 @@ export class RTMClient extends EventEmitter {
           .onSuccess().transitionTo('connecting')
         .onExit(() => this.teardownWebsocket())
       .global()
-        .onStateEnter((state) => {
+        .onStateEnter((state, context) => {
           this.logger.debug(`transitioning to state: ${state}`);
           // Emits events: `disconnected`, `connecting`, `connected`, 'disconnecting', 'reconnecting'
-          this.emit(state);
+          // event payload is anything related to the event, i.e. an error on disconnect
+          this.emit(state, context.eventPayload);
```
aoberoi: it seems like this could have effects greater than passing through the error on disconnect. i'd like to be a little more explicit about what this argument will contain for each kind of transition.

src/RTMClient.ts
clavin: If the web socket immediately errors on closing (that is, on the same tick of the event loop as `handle` is called) then the event handler won't be registered by the time the state machine transitions, and the promise will never reject.

Alternatively, the web socket could also immediately close (e.g. when mocked for testing) and the same problem would happen, though the promise wouldn't *resolve*.

I'm just being cautious that the reactive bit of code is set up before I set off the chain reaction. Don't want any race conditions happening 😅

src/RTMClient.ts
clavin: Just looking through the code, the only other case I could see where something else passes through would be the [event payloads for web socket events](https://github.com/slackapi/node-slack-sdk/blob/b207d885cae73aef97301de8b1b65f7b1195ffb7/src/RTMClient.ts#L554-L555), specifically the [non-existent `open` event's payload](https://github.com/websockets/ws/blob/e4d032c383dd2931de7dfbe81a0a8185cdffbf52/doc/ws.md#event-open) and the web socket closing code and reason from the [`close` event](https://github.com/websockets/ws/blob/e4d032c383dd2931de7dfbe81a0a8185cdffbf52/doc/ws.md#event-close).

I agree, though, that this could end up leaking something else through. Is there a better course of action? Like guarding that only `eventPayload instanceof Error` passes?

aoberoi: it seems like the intention is to _only_ emit the `context.eventPayload` when transitioning to `"disconected". if that's true, i think it would be clearer to do something like:

```javascript
if (state === 'disconnected') {
  // when arriving at this state due to failure, the eventPayload will contain an error.
  // otherwise, like after a force disconnect, this value will be undefined
  const errorOrUndefined = context.eventPayload;
  this.emit(state, errorOrUndefined);
} else {
  this.emit(state);
}
```

its more explicit, and gives us a chance to document the intention.

aoberoi:
thanks!

PR #620, opened 2018-08-22
 Replace dependency got with axios

### LINKED ISSUE #536: update @types/got to v8
### Description

Now that the following issue is fixed, it seems reasonable (and it would clean up some quirks in the code) https://github.com/DefinitelyTyped/DefinitelyTyped/issues/23454

aoberoi:
###  Summary

There's two main reasons for this implementation detail change:

1) `got` has moved onto the next major version and drops support for node versions this package must continue to support. they have no plans to continue to ship features and bugfixes on the v8 release line.
2) `axios` has a significantly reduced bundle size compared to `got` v8.x.

This swap also removes the need for a couple `ts-ignore` pragmas. But, since `axios`'s typings are much looser than `got`'s, it also means the checks are less valuable. Still, it seems like the right tradeoff.

Fixes #536 

### DIFF
--- a/src/WebClient.ts
+++ b/src/WebClient.ts
@@ -121,6 +121,8 @@ export class WebClient extends EventEmitter {
       validateStatus: () => true, // all HTTP status codes should result in a resolved promise (as opposed to only 2xx)
       maxRedirects: 0,
     });
+    // serializeApiCallOptions will always determine the appropriate content-type
+    delete this.axios.defaults.headers.post['Content-Type'];
 
     this.logger.debug('initialized');
   }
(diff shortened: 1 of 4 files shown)


src/WebClient.ts
```diff
@@ -249,15 +242,15 @@ export class WebClient extends EventEmitter {
                     await delay(retrySec * 1000);
                     // resume the request queue and throw a non-abort error to signal a retry
                     this.requestQueue.start();
-                    // TODO: turn this into a WebAPIPlatformError
                     throw Error('A rate limit was exceeded.');
                   } else {
+                    // TODO: turn this into some CodedError
```
shanedewael: ??

Issue #617, opened 2018-08-15
RFC: Short-lived tokens for workspace apps


aoberoi:
The Slack platform will soon allow developers to use short-lived access tokens, which is a security feature first [discussed at Spec](https://www.youtube.com/watch?v=9JLegLCu6_k&list=PLWlXaxtQ7fUbvBMja82vIinN3LG9XotOU&index=5).

This issue will capture the work associated with supporting short-lived tokens in this package. At this point we'd like to share the planned API changes, and we're hoping for your feedback and comments.

### Goals:

1. Give developers a low friction option to use short-lived tokens by automatically handling token refresh without any intervention.

   A. These developers should be allowed to opt-in (but are not required) to observe the token refresh process in order to persist the latest access token to their own storage when appropriate.

2. Allow developers who have specialized needs to use short-lived tokens and implement the token refresh process on their own.

3. Allow developers who are not using short-lived tokens to be unaffected by the code changes, and for their usage to continue as expected.

#### Constructor

`new WebClient(token?: string, options?: WebClientOptions)`

The `WebClientOptions` interface will gain three new optional properties:

*  `refreshToken?: string` - an `xoxr-...` token provided by the initial token exchange with Slack. it also might be loaded from the environment or a database when the application is launched.
*  `clientId?: string` - the client ID provided by Slack on app creation
*  `clientSecret?: string` - the client secret provided by Slack on app creation

When all three of these options are specified, the client will function in automatically-refreshing mode. That means it will attempt to handle failures to API calls due to token expiration (see token refresh below). If any of these options is specified, all must be, otherwise the constructor will throw an error.

The `token` is still optional, but when it is not provided and the client is in automatically-refreshing mode, it will treat the first API call the same as one which failed due to token expiration and initiate token refresh before attempting the call.

The client does *not* handle initial token exchange to retrieve a refresh token. Developers can initialize an unauthenticated `WebClient` instance (without a `token`) to call the `client.oauth.access()` method and use the values in the result to initialize another, authenticated, `WebClient` instance. Or, they may get the values from an existing OAuth 2.0 library/framework.

#### Events

`token_refreshed`

This is a new event type that will be emitted after a successful token refresh occurs. By the time this event is triggered, the client has already been updated to use the refreshed values. A value is emitted with this event type, and its of a new interface type `TokenRefreshedEvent` with properties as follows:

*  `access_token: string` - the token from the refresh
*  `expires_in: string` - the number of seconds the token will remain valid
[...]

PR #624, opened 2018-08-26
WIP Add token refresh

aoberoi:
###  Summary

This PR implements the changes described in #617.

This is still a work in progress with at least the following TODO items:

- [x] try running with some actual short lived tokens
- [x] lots of opportunities for refactors and cleanups
- [x] get linter to pass
- [x] get test coverage up to acceptable levels

I wanted to share this even at this slightly less mature stage for visibility into the progress.

### DIFF
--- a/src/WebClient.ts
+++ b/src/WebClient.ts
@@ -218,7 +218,7 @@ export class WebClient extends EventEmitter {
       // build headers
       const headers = {};
       if (options !== undefined && optionsAreUserPerspectiveEnabled(options)) {
-        headers['x-Slack-User'] = options.on_behalf_of;
+        headers['X-Slack-User'] = options.on_behalf_of;
         delete options.on_behalf_of;
       }
 
--- a/src/WebClient.ts
+++ b/src/WebClient.ts
@@ -211,6 +211,12 @@ export class WebClient extends EventEmitter {
 
       // check for expired access token, and refresh token accordingly
       // an exception is made when this call includes its own token in the options
+      // TODO: what if the token refresh is in progress when this gets called? as is, it would start refreshing again
+      // one option is to make this condition fail so that it doesn't refresh, allow the API call to be made, and then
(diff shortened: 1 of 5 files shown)


aoberoi:
Some notes from this implementation:

Idea: `isTokenRefreshing` as a Promise.

What if the token refresh is in progress when the optimistic token refresh conditions are checked? With this implementation when the API call fails, the code will check the request time and determine that the token should be refreshed (again) and the API call will be tried again.

But another option would be to prevent the first attempt to the API call in the first place by waiting on the token refresh in progress to finish (if we represent `isTokenRefreshing` not as a boolean, but as a Promise that this API call can await (a mutex of sorts, or maybe a barrier, i can't remember the right terminology). The failure logic doesn't change, but it would be an optimization.

Idea: Only use optimistic checks to perform token refresh

What's the point in doing a token refresh right before a retry if the retry will do an optimistic check anyway? The refresh could just occur in one place.

src/WebClient.ts
```diff
@@ -26,16 +26,61 @@ const pkg = require('../package.json'); // tslint:disable-line:no-require-import
...
 
+  /**
+   * The backing store for the current access token.
+   */
+  private _accessToken?: string;
+
+  /**
+   * The time (in milliseconds) when the current access token will expire
+   */
+  private accessTokenExpiresAt?: number;
```
Roach: Should be `accessTokenExpiresIn`, it's a countdown TTL, not an absolute timestamp of the time of expiration.

aoberoi: the response contains `expires_in`, you're right. but the client only stores that value after it adds it to the current time, so this is still correct. it wouldn't be useful to store just the `expires_in` because we would have to compute the future time that the expiration will occur on every request. this way, we only do that computation once and store it. my biggest concern was also clearing this value at the appropriate time, which is handled by the setter.

Roach: Ah. I missed that. Carry on!

Issue #628, opened 2018-08-31
Add an option to WebClient initialization to specify static headers


aoberoi:
### Description

There's some (rare) situations where you want send custom headers with each request to Slack's APIs. Maybe your app is behind a gateway that requires certain headers. Maybe you're sending requests to a testing server and it requires certain headers. We would be able to support these use cases if there was a way to initialize the `WebClient` with a set of custom headers.

Example:

```
const slackWeb = new WebClient(token, { headers: { 'X-My-Custom-Header': 'foo' } });
```

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

ethan0920:
Hi @aoberoi, 

I have been working on this. And I was wondering if it would be useful to include a way to change the headers with a method on the `WebClient`. Or retrieve them.  Thanks!

Ethan

clavin:
I feel like it might just make  the most sense to simply expose a field/property on `WebClient` called `customHeaders` or similar.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ethan0920:
@clavin I agree with you on that :+1:

aoberoi:
IMHO mutability (allowing to change the value after initialization) makes the implementation much more difficult. The WebClient needs to know how to deal with a queue of many requests being executed (or paused) at the same time. What happens when you change the custom headers but there’s already a few requests in queue? This is why the code for token refresh is so complex. 

I don’t see any use cases at the moment for updating the values for the custom headers. So let’s start with the simplest thing possible, just setting them on initialization. If a use case for changing the headers comes up later, we can always come back and add that. I’d like to avoid unnecessary complexity.

aoberoi:
@ethan0920 thanks so much for jumping onto this one! its always exciting to land a new community contribution. i've left some comments for you on the PR. let us know if you need any clarifications or help 😄

PR #629, opened 2018-09-03
Static headers on WebClient

ethan0920:
###  Summary

Add the ability to add static headers on an instance of `WebClient` that will be included in every request to Slack's API.

See #628 

### DIFF
--- a/src/WebClient.spec.js
+++ b/src/WebClient.spec.js
@@ -20,9 +20,6 @@ const token = 'xoxa-faketoken';
 const refreshToken = 'xoxr-refreshtoken';
 const clientId = 'CLIENTID';
 const clientSecret = 'CLIENTSECRET';
-const staticHeaders = {
-  'X-ABC': 'value'
-};
 
 describe('WebClient', function () {
 
@@ -37,12 +34,6 @@ describe('WebClient', function () {
     it('should build a client without a token', function () {
       const client = new WebClient();
       assert.instanceOf(client, WebClient);
-    });
-
-    it('should build a client with static headers', function () {
-      const client = new WebClient(token, { headers: staticHeaders});
-      assert.instanceOf(client, WebClient);
-      assert.equal(client.staticHeaders, staticHeaders);
     });
   });
 
@@ -87,7 +78,7 @@ describe('WebClient', function () {
 
   describe('apiCall()', function () {
     beforeEach(function () {
-      this.client = new WebClient(token, { retryConfig: rapidRetryPolicy, headers: staticHeaders });
+      this.client = new WebClient(token, { retryConfig: rapidRetryPolicy });
     });
 
     describe('when making a successful call', function () {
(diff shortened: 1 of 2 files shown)


src/WebClient.ts
```diff
@@ -122,6 +122,11 @@ export class WebClient extends EventEmitter {
    */
   private logger: Logger;
 
+  /**
+   * Static headers on every request.
+   */
+  private staticHeaders?: object;
```
ethan0920: I'm curious if there is any purpose of even making it a property on the WebClient? I only use is in the constructor when the axios instance is created.

aoberoi: IMHO i don't think we should keep this property. i think you've got the right idea about mixing the object provided in the options into the axios instance inside the constructor, and that's all we need right now.

src/WebClient.spec.js
```diff
@@ -35,6 +38,12 @@ describe('WebClient', function () {
       const client = new WebClient();
       assert.instanceOf(client, WebClient);
     });
+
+    it('should build a client with static headers', function () {
```
aoberoi: i don't think this is a particularly useful test, since you declared the `staticHeaders` property private. and, if we remove that property as discussed in the comment you added, this would no longer pass. i think we should remove this test case.

src/WebClient.spec.js
```diff
@@ -78,7 +87,7 @@ describe('WebClient', function () {
 
   describe('apiCall()', function () {
     beforeEach(function () {
-      this.client = new WebClient(token, { retryConfig: rapidRetryPolicy });
+      this.client = new WebClient(token, { retryConfig: rapidRetryPolicy, headers: staticHeaders });
```
aoberoi: i don't think we should initialize with the static headers for all test cases inside this `describe` block. instead, i think we should add another `describe` block inside this one (as done for other specific features), maybe `describe('when using custom static headers', function () { /* ... */ });` Within that block, we should have at least one test case where a client is constructed with custom static headers, and the request is checked for the existence of those headers (using `nock`). We probably also don't need `staticHeaders` to be declared globally in this file. Let me know if you disagree or if you'd like any further help.

ethan0920:
Thank you @aoberoi for the reviews, I will make these changes shortly :smile:

ethan0920:
Hey @aoberoi,

I am having some troubles implementing the test for this. If you could give a brief explanation how I could tackle this (if you have any ideas). I do have to go to work for a a few hours and then I will be back to work on these changes further.

Ethan

src/WebClient.spec.js
```diff
@@ -494,6 +485,23 @@ describe('WebClient', function () {
     });
   });
 
+  describe('apiCall() - when using static headers', function () {
+    it('should include static headers on api request', function () {
+      const client = new WebClient(token, { headers: { 'X-XYZ': 'value' } });
+      const scope = nock('https://slack.com')
+        .post(/api/)
+        .reply(200, { ok: true })
+        .on('request', function (req, interceptor, body) {
```
aoberoi: this is a pretty clever way to check for headers, cool!

i think it would be a little more straight-froward if you were to use nock's `reqheaders` option, which was designed for this purpose. check it out here: https://github.com/nock/nock#specifying-request-headers

aoberoi:
@ethan0920 no problem! thanks for being so diligent. one small comment left for you.

src/WebClient.spec.js
ethan0920: Ah, thank you! Didn't notice that when looking over Nock :+1:

ethan0920:
Okay, I have done the changes. Sorry for the delay, been a little busy :+1:

aoberoi:
this looks great, thanks for the contributions and your patience to stick with this. 👍

PR #643, opened 2018-09-25
Migrate from 'channels.list' to 'conversations.list'

smaeda-ks:
###  Summary

Because your document says:
https://api.slack.com/methods/channels.list
> Don't use this method. Use conversations.list instead.

The sample code should be migrated as well.


aoberoi:
@smaeda-ks thanks so much for this contribution. you caught a very important best practice that we didn't previously mention in the documentation. your code example and explanation is very thorough and complete.

i have one recommendation: i'd prefer to keep pagination out of this code example so that we can present the simplest way first, and then lead the reader to understanding pagination separately (as you already did with the link to that heading). we can instead mention that the response will be limited to 100 results, and for the full list they can read about `[pagination](#pagination)`. i think this keeps the example clear so the reader doesn't need to learn about too many things at once.

we could also change the example code in the pagination section to also use `conversations.list` so that a reader coming from this section can more simply follow along.

if you need any help, let us know. i think what you already changed is really good, so i'm happy to let you finish the changes yourself if you'd like to.

Issue #649, opened 2018-10-02
Remove mentions of es2018.asynciterable from TypeScript docs


aoberoi:
### Description

#644 merged a docs change that mentioned for typescript 3.1 or later, you should use `"es2018.asynciterable"` in the `lib` configuration. that change was scheduled for typescript 3.1, but was later moved to 3.2, so this information is no longer accurate. we should remove that section until the change lands.

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

melisdogan:
I'll do this.

PR #650, opened 2018-10-02
Removed mention of es2018.asynciterable from TypeScript documentation

melisdogan:
###  Summary

Removed mention of es2018.asynciterable, as per #649 

* [x ] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x ] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

### DIFF
--- a/docs/_pages/typescript.md
+++ b/docs/_pages/typescript.md
@@ -46,8 +46,7 @@ For versions of TypeScript >= 3.1:
     "lib": [
       "es5",
       "es2015",
-      "es2016.array.include",
-      "es2018.asynciterable"
+      "es2016.array.include"
     ]
   }
 }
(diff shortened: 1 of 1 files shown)


aoberoi:
thanks @melisdogan! i think i was a little unclear in the issue, but even users of TS 3.1 or greater need to use `"esnext.asynciterable"` in `lib`. sorry about the confusion.

if we were to add that lib, then the whole config would look exactly like the config above it. so i think we could just remove the second config and the line about "For versions of TypeScript < 3.1:". what do you think?

aoberoi:
awesome 🥇

Issue #660, opened 2018-10-30
Add thread_ts to FilesUploadArguments


aoberoi:
### Description

The `thread_ts` argument allows apps to add files to a message in a thread, rather than a top-level channel message. The argument isn’t yet included in `FilesUploadArguments`, but should be. It would also be useful to comment that in order or use that argument it’s also required to use `channels`. 

#659

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

PR #661, opened 2018-10-30
Add thread_ts to files.upload type definition

tcd156:
###  Summary

`thread_ts` is a possible value to supply to `files.upload`, but it currently does not exist in the type definition. This PR adds `thread_ts` to the type definition, as well as a comment to specify that you must supply `channels` in order for the `thread_ts` value to work properly.

#660 

### DIFF
--- a/src/methods.ts
+++ b/src/methods.ts
@@ -464,6 +464,7 @@ export type FilesUploadArguments = TokenOverridable & {
   filetype?: string;
   initial_comment?: string;
   title?: string;
+  thread_ts?: string; // if specified, `channels` must be set
 };
 export type FilesCommentsAddArguments = TokenOverridable & {
   comment: string;


docs/_pages/web_client.md
```diff
@@ -139,6 +139,10 @@ web.files.upload({
   file: fs.createReadStream(`./${fileName}`),
   // Or you can use the content property (but not both)
   // content: 'plain string content that will be editable in Slack'
+  // Choose specific channels you want the file to be posted to
+  // channels: 'comma separated list of channels'
```
aoberoi: Thanks for being so comprehensive!

I don't feel like these belong here, since (`file` OR `content`) are required, while `channels` OR (`channels` and `thread_ts`) are optional. This example is meant to show the simplest of scenarios for developers who just want a quick recipe to copy and paste. If we wanted to be exhaustive, we'd have to write quite a bit more to describe all the arguments.

I'm open to feedback through. Do you think most developers would expect to post files within a conversation? Going further, do you think the majority of those would want to understand how it would be done for within a thread?

tcd156: That totally makes sense. I guess I didn't immediately understand the use case for just uploading files to Slack without having the intention of posting them to a channel. Is that a common use case?

I do agree it felt odd here - but since there wasn't any other documentation directly in this repo, I felt like I should add it somewhere. I will say - our use case was to use Hubot, and uploading files into threads was such a perfect scenario for us because we would be absolutely destroying our chat with images otherwise. Having hubot answer in threads reduces the clutter by orders of magnitude.

IMO, posting images into threads seems like a pretty useful feature, so even if it isn't front and center, I think it should absolutely be documented in the API. But this is coming from our use case, so I'm not sure if that's the exception.

tcd156:
@aoberoi let me know about the docs, I can get rid of them so we can get this closed.

docs/_pages/web_client.md
aoberoi: Thanks for adding your point of view. I'm convinced now that `channels` deserves a mention. I wouldn't say that it's common to upload a file without this, just that it's simpler. But I think you're right that its likely they want to see the uploaded file in a conversation.

How about the example keeps the comment with the `channels` argument (as you added), but instead of putting an explanation in the string value, we use something that looks like a real channel ID (e.g. `C123456`).

I'm not as convinced about `thread_ts` being needed in the example, but I think it could be mentioned in the comment. Maybe something along the lines of `Specify channel(s) where the uploaded file should be shared. This is optional, except when also specifying a thread_ts for putting the message in a thread.` I don't love that sentence, but I think it includes the important pieces.

aoberoi:
@tcd156 just responded, thanks for your patience.

docs/_pages/web_client.md
tcd156: What about something like `Specify channel(s) to upload the file to. Optional, unless also specifying a thread_ts value`. 

I think that may be clear enough in that - if you want to upload to a channel, it doesn't matter to you, but if you want to upload to a thread, at least some information is there.

aoberoi: Sounds good!

tcd156:
alright @aoberoi, changes are made 👍

PR #688, opened 2019-02-08
Adds Block types

shanedewael:
###  Summary

Adds Block types

### DIFF
--- a/src/methods.ts
+++ b/src/methods.ts
@@ -106,7 +106,157 @@ export interface Dialog {
   state?: string;
 }
 
+/*
+ * Block Elements
+ */
+
+export interface ImageElement {
+  type: 'image';
+  image_url: string;
+  alt_text: string;
+}
+
+export interface UserElement {
+  type: 'user';
+  user_id: string;
+}
+
+export interface PlainTextElement {
+  type: 'plaintext';
+  text: string;
+  emoji?: boolean;
+}
+
+export interface MrkdwnElement {
+  type: 'mrkdwn';
+  text: string;
+  parse?: 'none';
+}
+
+export interface Option {
+  text: PlainTextElement;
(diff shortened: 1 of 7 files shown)


src/methods.ts
```diff
@@ -106,7 +106,157 @@ export interface Dialog {
...
+  type: 'section';
+  block_id?: string;
+  text?: PlainTextElement | MrkdwnElement; // either this or fields must be defined
+  fields?: (PlainTextElement | MrkdwnElement)[]; // either this or text must be defined
+  accessory?: Button | Overflow | StaticSelect | UsersSelect | ConversationsSelect |
+    ChannelsSelect | ExternalSelect | Datepicker;
+}
+
 export interface MessageAttachment {
+  blocks?: Block[];
```
aoberoi: thinking about the future-proofing scenario, if we don't have an explicit type for the block in  the `Block` union, then this would be a type checker error. i think we might also need to use the trick we've been applying where you rename `Block` to `KnownBlock`, and then change this to `(KnownBlock | Block)[]` (or pick better names). is there anything we know all `Block` objects should have (probably just `type: string`, right)?

src/methods.ts
```diff
@@ -294,6 +444,7 @@ export type ChatPostEphemeralArguments = TokenOverridable & {
   user: string;
   as_user?: boolean;
   attachments?: MessageAttachment[];
+  blocks?: Block[];
```
aoberoi: if we want to use `(KnownBlock | Block)[]` in all these places, maybe its worth giving that a type alias. so maybe its `type Block = KnownBlock | OtherBlock` (where `type KnownBlock = ImageBlock | ContextBlock | ...`, and `interface UnknownBlock { type: string; }`.

Issue #701, opened 2019-02-13
Method for paginated API calls


aoberoi:
### Description

Automatic pagination is implemented in the SDK, but its not recommended to depend on this functionality. That is because the SDK will paginate until the end of the collection, and then merge the contents from the individual pages in a destructive manner. However, manually paginating takes a lot of effort to do right - recursion isn't easy.

It would be useful to for the `WebClient` to provide a method that "did pagination right". It would be better than automatic pagination because it should "yield" the full results for each page (not destructive), and provide a means to
stop pagination at any point (not needlessly exhaustive).

Here's the current proposal:

```javascript
const { WebClient } = require('@slack/client');

const web = new WebClient(process.env.SLACK_TOKEN);

// I'd like to find a public channel with a certain name, but this requires paging
// through potentially all the channels.

const channelName = 'team-awesome';

const done = web.paginate(
  // the first argument is the method name of a cursor-paginated method
  'conversations.list',

  // the second argument is the options to pass to that method
  { types: 'public_channel' }

  // the third argument is a predicate which returns `true` when pagination
  // should stop (so that not returning anything is treated as continue).
  // it is called with the individual page result
  (res) => res.channels.some(c => c.name === channelName),

  // the fourth (and optional) argument is a reducer function
  // it is called with an accumulator, the individual page result, and the page number
  // acc is initialized to undefined
  (acc, res, idx) => {
    // the accumulated value will be what the Promise returned (done) resolves to
    for (channel of res.channels) {
      if (channel.name === channelName) { return channel.id }
    }
  },
});

(async () => {
  // The method returns a Promise when pagination is complete. The resolved
  // value could be undefined (when there's no reducer defined) or the
  // accumulated value (when there is a reducer defined).
  const channelId = await done;
  console.log(`Channel name: ${channelName}, ID: ${channelId}`);
})()
```

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

aoberoi:
Problem with this design: there's no way to know what method to call for the next page, and with which options. Should we assume the options are static (the same for every page) except for the `cursor`?

Lemmmy:
[...]

PR #741, opened 2019-03-13
Paginate method

aoberoi:
###  Summary

Adds `WebClient#paginate()` method, as described in #701.

This is still a work in progress, but I'm looking for some help to figure out why one of the integration tests is failing.

As a result of the assertion in `test/types/webclient-paginate-types.ts` L44, the following error is emitted:

```
ERROR: 41:3  expect  Expected type to be:
  Dummy | undefined
got:
  any
```

From the implementation, I expect the conditional type `PageAccumulator<R>` to infer the type of the accumulator from the function. I believe that by returning a `Dummy` in the test case reduce param, the conditional type has all it needs to infer this correctly, and it seems to be applying types for the other arguments correctly.

Honestly, I don't like having to do build the signature of `WebClient#paginate()` with so many overloads. I'm hoping to find an approach to defining it with a conditional type as the return type. I've tried to do this locally, and kept getting type checking errors where I didn't think I should be. I'm going to keep exploring this approach by trying to produce a minimal reproduction case (that could possibly be shared on an issue at Microsoft/TypeScript if it turns out to be a bug or feature request), and I'll report back.

Oh yeah, and I also have to write some unit tests for this. But I figured I should work on getting the types right before I do that.


aoberoi:
Reference docs generation is borked on dev-v5. There's an incompatibility between one of the tools we depend on (`ts-jsdoc`) and `typescript@3`.

I don't think we should spend time overhauling and rebuilding our reference documentation generation process. The output is already very bare bones, and I don't think a whole lot of people find it valuable. I propose that we just remove the generated reference documentation from the site until we can find a decent alternative.

PR #756, opened 2019-03-28
Release v5.0.0

### LINKED ISSUE #689: Availability of Node Slack SDK v5 (timeline for end of support on v4)
We're excited to announce the next major version of this package - or maybe we should say _packages_. This [milestone](https://github.com/slackapi/node-slack-sdk/milestone/3) is set to track the progress for its availability in March 2019.

**TL;DR**: If you're using a modern version of Node, and are not depending on any deprecated features, you can simply update your `package.json` with the latest version of the package(s) you use. No code changes! Take a look below for the new package names.

**Please update to v5 before v4 is no longer supported, which is May 31, 2019.**

### New structure, same great experience 📦

This repository will become home to six, separately installable packages:

*  `@slack/web-api` - a package for using the Web API, whose main export is the `WebClient`.

*  `@slack/rtm-api` -  a package for using the RTM API, whose main export is the `RTMClient`.

*  `@slack/webhook` - a package for using Incoming Webhooks, whose main export is the `IncomingWebhook`.

*  `@slack/client` - an umbrella package that groups together the previous three packages. It combines those packages so that the exports are in the same structure as in versions `v4.x`. As long as you're not using deprecated functionality, you can update to v5 with a simple `npm install @slack/client@latest`. We'll stop publishing this package in the future, its only meant to alleviate migration friction in the short term.

*  `@slack/events-api` - same package as it exists today, just moving the code into this repo.

*  `@slack/interactive-messages` - same package as it exists today, just moving the code into this repo.

We recommend all users install only the package(s) they use in their app. So if you only call the Web API, you only need the `@slack/web-api` package. This will result in smaller dependency trees, which are better for performance (install time, cold boot time, etc).

This eliminates confusion for developers who need help when using our packages together. When there's only one repo and one set of issues, you're more likely to find what you need, and get attention for problems you're having.

We'll be updating our processes to reflect the structure change. The most relevant change for users is that we will close the issues in the [`slackapi/node-slack-events-api`](https://github.com/slackapi/node-slack-events-api) and [`slackapi/node-slack-interactive-messages`](https://github.com/slackapi/node-slack-interactive-messages) repositories, and create copies of them in this repo. We'll update documentation to redirect visitors to this repo, and then archive those repos.

### Add HTTP and HTTPS proxy support 🔒

The `WebClient` and `RTMClient` have been able to support using HTTP and HTTPS proxies through the `agent` configuration. That required using another dependency (such as `https-proxy-agent`) - but no more! Both of these classes now have a `proxy` option which can be set directly. This also means that the `agent` option has been removed, since there was no major use cases for it besides setting a proxy.

### Update minimum Node version 8️⃣ 

As the support period for Node v6 LTS comes to a close, we're updating these packages to require v8 LTS or later. This change follows the policy we announced last year in the [Support Schedule](https://github.com/slackapi/node-slack-sdk/wiki/Support-Schedule).

### Return Promises (only) from Web API methods 🔜

Promises have become ubiquitous and performance has been improving. With the availability of `async`/`await` in all supported versions of Node, code clarity has also improved. The `WebClient` stops handling callback functions, and always returns a Promise.

If you're not ready or otherwise don't prefer using Promises, you can use the Node
built-in [`util.callbackify`](https://nodejs.org/api/util.html#util_util_callbackify_original) in your code to bring back the callbacks 🔙.

### Logging updates 📝 

The `logger` option for `WebClient` and `RTMClient` accept objects that implement the `Logger` interface. This interface has separate methods for each `LogLevel`, which allows a developer to more easily describe the handling those logs at each level. It also means your logger will deal with rich objects instead of a flattened string. The `LoggerFunc` interface will no longer be supported.

### TypeScript 3.3 ✔️ 

This version takes advantage of the latest features in its exported type definitions. This includes the awaited [`/// <reference lib="..." /> reference directives`](https://github.com/Microsoft/TypeScript/pull/23893), which will resolve some [well-understood](https://github.com/slackapi/node-slack-sdk/issues/612) [usability](https://github.com/slackapi/node-slack-sdk/issues/615) [challenges](https://github.com/slackapi/node-slack-sdk/issues/649).

aoberoi:
###  Summary

Fixes #689 


aoberoi:
Travis has been stalled for many hours. We're going to merge this now, and work on a release. If we find any problems, we may revert.

packages/client/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist/**/*"
+  ],
+  "engines": {
+    "node": ">= 8.9.0",
+    "npm": ">= 5.5.1"
+  },
+  "repository": "slackapi/node-slack-sdk",
+  "homepage": "https://slack.dev/node-slack-sdk/tutorials/migrating_to_v5/",
```
CvX: ☝️ This URL leads to a 404

Roach: Correct URL: https://slack.dev/node-slack-sdk/tutorials/migrating-to-v5/

aoberoi: Fixed in https://github.com/slackapi/node-slack-sdk/pull/761

We’ll likely be releasing that today.

PR #759, opened 2019-03-29
Fix eventemitter3 import

Marsup:
###  Summary

Latest v5.0.0 fails to build on our typescript project with this:
```
node_modules/@slack/web-api/dist/WebClient.d.ts:5:8 - error TS1192: Module '"node_modules/eventemitter3/index"' has no default export.
```

This is very likely due to your modules using `esModuleInterop: true`, while we don't.

### DIFF
--- a/packages/rtm-api/src/RTMClient.ts
+++ b/packages/rtm-api/src/RTMClient.ts
@@ -1,6 +1,6 @@
 import { Agent } from 'http';
 
-import EventEmitter from 'eventemitter3'; // tslint:disable-line:import-name
+import EventEmitter = require('eventemitter3'); // tslint:disable-line:import-name no-require-imports
 import WebSocket from 'ws'; // tslint:disable-line:import-name
 import Finity, { StateMachine } from 'finity'; // tslint:disable-line:import-name
 import PQueue from 'p-queue'; // tslint:disable-line:import-name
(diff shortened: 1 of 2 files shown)


aoberoi:
@Marsup Thanks for catching this one. I was able to reproduce the same error just by turning `esModuleInterop` off.

Thanks for making the PR, too! I took a closer look and found another solution that I think we should prefer. In #760, instead of leaving the ES Module import syntax, I adjusted it to use the named export, rather than the default export. I also adjusted the integration test to verify this compatibility.

Any objection to that approach?

Marsup:
Sure why not. What's the meaning of all those `gitHead`s in your PR ? It's the 1st time I see that in node.

If you're going this way you should probably fix `packages/rtm-api/src/KeepAlive.ts` import as well, I went with this format since it was already used there. You can also remove the ts-lint comments.

aoberoi:
> Sure why not. What's the meaning of all those `gitHead`s in your PR ? It's the 1st time I see that in node.

I think this is how `lerna` tracks which packages in the repo have changed since their last release. We're also still getting used to this, so it will take us a little time to learn.

> If you're going this way you should probably fix `packages/rtm-api/src/KeepAlive.ts` import as well, I went with this format since it was already used there. You can also remove the ts-lint comments.

good call!

aoberoi:
Closing this PR in favor of #760

PR #760, opened 2019-03-29
Change eventemitter3 import for compatibility without esModuleInterop

aoberoi:
###  Summary

When working in a TypeScript project whose configuration doesn't set `esModuleInterop` to true, compilation would fail with the following error:

```
node_modules/@slack/web-api/dist/WebClient.d.ts:5:8 - error TS1192: Module '"/mypath/node_modules/eventemitter3/index"' has no default export.

5 import EventEmitter from 'eventemitter3';
         ~~~~~~~~~~~~
```

The `eventemitter3` package has a tricky export where it uses both a value in CommonJS and a named export. This PR adjust the import statement to rely on the named export.

It also adjusts the integration tests to check for this compatibility (the test failed before the change in import was applied).

#759 

### DIFF
--- a/packages/rtm-api/src/KeepAlive.ts
+++ b/packages/rtm-api/src/KeepAlive.ts
@@ -1,4 +1,4 @@
-import EventEmitter = require('eventemitter3'); // tslint:disable-line:import-name no-require-imports
+import { EventEmitter } from 'eventemitter3'; // tslint:disable-line:import-name
 import { RTMClient } from './RTMClient';
 import { LogLevel, Logger, getLogger } from './logger';
 import { ErrorCode, CodedError } from './errors';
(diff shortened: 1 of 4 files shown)


packages/rtm-api/src/KeepAlive.ts
```diff
@@ -1,4 +1,4 @@
-import EventEmitter = require('eventemitter3'); // tslint:disable-line:import-name no-require-imports
+import { EventEmitter } from 'eventemitter3'; // tslint:disable-line:import-name
```
Marsup: You can also remove the ts-lint comment, import-name is now valid.

Marsup:
Are you going to release a new version with this anytime soon?

PR #787, opened 2019-05-07
fix examples of proxy agent usage

ericzon:
###  Summary

Fix documentation for using https-proxy-agent. If I follow the current docs about this, I get this error:

TypeError: HttpsProxyAgent is not a constructor

Checking original lib https://github.com/TooTallNate/node-https-proxy-agent I see the right way to initialize the agent:
```
const HttpsProxyAgent = require('https-proxy-agent');
```


clavin:
Closed in favor of merging #811 (even though it was created later). The changes happen to be the same.

ericzon:
So weird, I was already signed the CLA:

image

Anyway, the important here is that now the example is fixed :)

PR #792, opened 2019-05-26
Add Logger#getLevel(): LogLevel (to test what current level is)

seratch:
###  Summary

This pull request adds a new method `getLevel(): LogLevel` to `Logger` interface. Having the method means to enable developers to do expensive operations only when debug logging is enabled. For example, dumping request headers only when debug level logging is enabled is a common use case. 

My motivation to add this method is to have more detailed logging in node-slack-sdk & bolt without any performance compromise in production env.

### DIFF
--- a/packages/logger/.mocharc.json
+++ b/packages/logger/.mocharc.json
@@ -1,0 +1,4 @@
+{
+  "require": ["ts-node/register", "source-map-support/register"],
+  "timeout": 3000
+}
--- a/packages/logger/package.json
+++ b/packages/logger/package.json
@@ -30,13 +30,19 @@
     "build": "npm run build:clean && tsc",
     "build:clean": "shx rm -rf ./dist",
     "lint": "tslint --project .",
-    "test": "npm run build && echo \"Tests are not implemented.\" && exit 0"
+    "test": "npm run build && nyc mocha --config .mocharc.json src/*.spec.js"
   },
   "dependencies": {
     "@types/node": ">=8.9.0"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


seratch:
I've fixed the lint errors.

packages/logger/tslint.json
```diff
@@ -42,7 +42,7 @@
     "no-empty": true,
     "no-floating-promises": true,
     "no-for-in-array": true,
-    "no-implicit-dependencies": true,
+    "no-implicit-dependencies": [true, "dev"],
```
seratch: This one could be helpful for other packages if you'd like to write tests in TypeScript. (Currently all the existing tests in other packages are written in JS)

seratch:
I'm still not sure why the integration tests are failing. Will check it out when I have time...

packages/logger/tslint.json
seratch: I realized other projects never use any transpilers and also found the following rule in `node-slack-interactive-messages`:

> Test code should be written in syntax that runs on the oldest supported Node.js version, without transpiling. This ensures that backwards compatibility is tested and the APIs look reasonable in versions of Node.js that do not support the most modern syntax.

https://github.com/slackapi/node-slack-interactive-messages/blob/master/.github/maintainers_guide.md#testing

I've updated this pull request not to use TypeScript this time.

seratch:
Just for reference: pull request #804 is also modifying `logger` package to add more tests.

aoberoi:
this looks really great! also, i think #804 is complimentary to this PR.

PR #793, opened 2019-05-28
fix incorrect homepage for legacy wrapper

felixrieseberg:
###  Summary

:wave: The link on https://www.npmjs.com/package/@slack/client is incorrect. We don't handle a final `/`, which in itself is probably worth looking at, but this ensures that npm doesn't route people to a disappointing 404 🙇 


aoberoi:
thanks @felixrieseberg!

PR #797, opened 2019-05-31
Add 'style' to Button type

deremer:
https://api.slack.com/docs/message-guidelines

###  Summary

Add style to Button type to allow setting as 'default', 'primary', or 'danger'

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -127,7 +127,7 @@ export interface Button extends Action {
   text: PlainTextElement;
   value?: string;
   url?: string;
-  style?: 'danger' | 'default' | 'primary';
+  style?: 'danger' | 'primary';
 }
 
 export interface Overflow extends Action {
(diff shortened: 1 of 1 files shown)


shanedewael:
@deremer Good catch adding the button styles. 🙌 

One small thing - the docs you referenced are for the old message buttons. The newer (Block Kit) button element's `style` doesn't include `default` (only `primary` and `danger`). Would you mind making that change?

deremer:
Ah, yes actually I had a hard time finding a good page with the actual reference, and that link was the best I could find. “Block kit” is the key to that search query. It should be this: https://api.slack.com/reference/messaging/block-elements#button

deremer:
I will update using the Block Kit Docs when I get a moment later. Thanks.

deremer:
@shanedewael I removed `default`

deremer:
Any idea when this will make it into a release?

clavin:
Sorry for the drop in communication 😬

I'm seeing if I can manage to release later today, so this will make its way into a release very soon! (Watch out for your @ in the changelog 😉)

deremer:
Thanks!

PR #799, opened 2019-06-12
Convert `@slack/events-api` and `@slack/interactive-messages` to TypeScript

### LINKED ISSUE #816: Add TypeScript definitions for events
Reopened from https://github.com/slackapi/node-slack-events-api/issues/35

### Description

Get some typing built into this package.

My 2c is that we should steer away from over-typing the actual events, but we could document at least a couple of "baseline" properties that all events have.

It shouldn't be too hard to describe the actual API of the module though. We should figure out how much JSDoc will do for us since the typescript language service can read JSDoc comments for types now. At this point I don't think there's any need to port to code to .ts, but open to hear feedback on this.

- [ ] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

clavin:
###  Summary

Migrates the two remaining JS packages (`@slack/events-api` and `@slack/interactive-messages`) to TypeScript. There should be no change in functionality.

> Fixes #816


packages/interactive-messages/src/adapter.ts
```diff
@@ -525,73 +509,99 @@ export class SlackMessageAdapter {
...
+ *   [message](https://api.slack.com/docs/interactive-message-field-guide#top-level_message_fields) for the message in
+ *   which the action occurred. It may also be a Promise for a message, and if so and the Promise takes longer than the
+ *   `syncResponseTimeout` to complete, the message is sent over the `response_url`. The message may also be a new
+ *   message in the same conversation by setting `replace_original: false`. When the action is a dialog submission,
+ *   this object is a list of [validation errors](https://api.slack.com/dialogs#input_validation). It may also be a
+ *   Promise for a list of validation errors, and if so and the Promise takes longer than the `syncReponseTimeout` to
+ *   complete, Slack will display an error to the user. If there is no return value, then button presses and menu
+ *   selections do not update the message and dialog submissions will validate and dismiss.
  */
+type ActionHandler = (payload: object, respond: Respond) => object | Promise<object> | undefined;
```
aoberoi: again, i suggest `object` -> `any`

packages/interactive-messages/src/adapter.ts
```diff
@@ -525,73 +509,99 @@ export class SlackMessageAdapter {
...
- * Dialog submissions do not allow `resplace_original: false` on this message.
- * @returns {Promise} there's no contract or interface for the resolution value, but this Promise
- * will resolve when the HTTP response from the `response_url` request is complete and reject when
- * there is an error.
+ * @param message - a [message](https://api.slack.com/docs/interactive-message-field-guide#top-level_message_fields).
+ *   Dialog submissions do not allow `resplace_original: false` on this message. @returnsthere's no contract or
+ *   interface for the resolution value, but this Promise will resolve when the HTTP response from the `response_url`
+ *   request is complete and reject when there is an error.
  */
+type Respond = (message: any) => Promise<void>;
```
aoberoi: this should return a `Promise<any>`, right?

packages/interactive-messages/src/http-handler.ts
```diff
@@ -141,3 +139,38 @@ export function createHTTPHandler(adapter) {
       });
   };
 }
+
+export type HTTPHandler = (req: IncomingMessage & { body?: string, rawBody?: Buffer }, res: ServerResponse) => void;
```
aoberoi: is there a meaningful difference between this type and `RequestListener` (provided by `@types/node`)?

packages/interactive-messages/src/http-handler.ts
```diff
@@ -40,13 +38,13 @@ export function createHTTPHandler(adapter) {
   /**
    * Parses raw bodies of requests
    *
-   * @param {string} body - Raw body of request
-   * @returns {Object} Parsed body of the request
+   * @param body - Raw body of request
+   * @returns Parsed body of the request
    */
-  function parseBody(body) {
+  function parseBody(body: string): object {
```
aoberoi: probably return `any`.

packages/interactive-messages/src/http-handler.ts
```diff
@@ -55,26 +53,28 @@ export function createHTTPHandler(adapter) {
...
-   * @returns {boolean} Indicates if request is verified
+   * @param signingSecret - Signing secret used to verify request signature
+   * @param requestHeaders - Request headers
+   * @param body - Raw body string
+   * @returns Indicates if request is verified
    */
-  function verifyRequestSignature(signingSecret, requestHeaders, body) {
+  function verifyRequestSignature(
+    signingSecret: string,
+    requestHeaders: Record<string, string>,
```
aoberoi: Should we instead use the `IncomingHttpHeaders` type from `@types/node`?

packages/interactive-messages/src/http-handler.ts
```diff
@@ -121,7 +119,8 @@ export function createHTTPHandler(adapter) {
 
           const dispatchResult = adapter.dispatch(body);
 
-          if (dispatchResult) {
+          if (dispatchResult !== undefined) {
+            // tslint:disable-next-line no-floating-promises
```
aoberoi: this seems dangerous. why would there be a floating promise? if we should be catching this and doing something with it, let's maybe mark that with a TODO?

packages/interactive-messages/src/http-handler.ts
```diff
@@ -141,3 +139,38 @@ export function createHTTPHandler(adapter) {
...
+ */
+type ResponseHandler = (dispatchResult: {
+  status: number,
+  content?: string | object,
+}) => void;
+
+/**
+ * Parameters for calling {@link verifyRequestSignature}.
+ */
+export interface VerifyRequestSignatureParams {
```
aoberoi: this seems unused.

packages/interactive-messages/src/index.ts
```diff
@@ -0,0 +1,16 @@
+import { SlackMessageAdapter, MessageAdapterOptions } from './adapter';
+import { ErrorCode } from './errors';
+
+/**
+ * Dictionary of error codes that may appear on errors emitted from this package's objects
```
aoberoi: we should probably drop a TODO here to get the rest of the error codes exported

packages/interactive-messages/tsconfig.json
```diff
@@ -0,0 +1,39 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
```
aoberoi: ditto with target from `@slack/events-api`.

packages/interactive-messages/types/lodash.isfunction.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'lodash.isfunction' {
+  export default function(x: any): x is Function;
```
aoberoi: it seems like you could have just installed `@types/lodash.isfunction`. i understand that its smaller on the bundle size to just include this here, but i feel like code duplication is a bigger burden.

aoberoi:
do we know why coverage isn't reporting for this PR?

packages/events-api/package.json
```diff
@@ -36,40 +36,39 @@
...
   "scripts": {
     "prepare": "npm run build",
-    "build": "babel src -d dist --source-maps both",
-    "lint": "eslint src test",
-    "test": "npm run build && nyc --reporter=html mocha test/**/*.js",
-    "coverage": "nyc report --reporter=text-lcov > coverage.lcov && codecov -F eventsapi"
+    "build": "npm run build:clean && tsc",
+    "build:clean": "shx rm -rf ./dist ./coverage ./.nyc_output",
+    "lint": "tslint --project .",
+    "test": "nyc mocha --config .mocharc.json src/*.spec.js test/integration/*.js",
```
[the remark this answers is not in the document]
clavin: Yes! The tests `require` the source files directly; there’s no need to build before testing as things currently are.

packages/events-api/package.json
```diff
@@ -36,40 +36,39 @@
...
     "tsscmp": "^1.0.6",
     "yargs": "^6.6.0"
   },
   "devDependencies": {
-    "babel-cli": "^6.18.0",
-    "babel-eslint": "^7.1.1",
-    "babel-preset-es2015": "^6.18.0",
-    "babel-preset-es2016": "^6.16.0",
-    "chai": "^4.1.2",
+    "@types/debug": "^4.1.4",
```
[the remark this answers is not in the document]
clavin: Only those types that show up in the exposed API (`.d.ts` files) need to be under `"dependencies"`. I checked the `dist/` dirs of each of the two packages and both only have a `<reference />` to `@types/node`. I should change the dependencies of (only) these two to only include `@types/node` in `"dependencies"` and other dependent types in `"devDependencies"`

packages/events-api/package.json
```diff
@@ -36,40 +36,39 @@
...
+    "@types/node": "^12.0.6",
```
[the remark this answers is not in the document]
clavin: (Oops.) I forgot to note this somewhere when I was installing deps. Good catch!

packages/events-api/src/adapter.ts
```diff
@@ -0,0 +1,156 @@
+import EventEmitter from 'events'; // tslint:disable-line
```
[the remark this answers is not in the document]
clavin: 🤔 I remember searching for & writing `import-name` after these. 🤷‍♀ I wholeheartedly agree & will gladly add the rule name!

packages/events-api/src/adapter.ts
```diff
@@ -0,0 +1,156 @@
...
+  private server?: http.Server;
+
+  /**
+   * @param signingSecret - The token used to authenticate signed requests from Slack's Events API.
+   * @param opts.includeBody - TODO:
+   * @param opts.includeHeaders - TODO:
+   * @param opts.waitForResponse - TODO:
+   */
+  constructor(
+    signingSecret: string | String,
```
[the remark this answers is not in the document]
clavin: The current JS version accepted strings *and* `String`s. `new String('x')` would make it through the current "type check" and be used because [`lodash.isstring`](https://github.com/lodash/lodash/blob/master/isString.js#L20) would check if the argument's type was a string or (essentially) if it was an instance of `String`. This typing reflects that.

packages/events-api/src/adapter.ts
```diff
@@ -0,0 +1,156 @@
...
+    {
+      includeBody = false,
+      includeHeaders = false,
+      waitForResponse = false,
+    }: EventAdapterOptions = {},
+  ) {
+    if (typeof signingSecret !== 'string' && !(signingSecret instanceof String)) {
```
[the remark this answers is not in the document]
clavin: From a design perspective, I honestly don't know. But from the perspective of adding typings, because it already was 🤷‍♀ [(see above)](https://github.com/slackapi/node-slack-sdk/pull/799#discussion_r296864473).

packages/events-api/src/adapter.ts
```diff
@@ -0,0 +1,156 @@
...
+const debug = debugFactory('@slack/events-api:adapter');
+
+/**
+ * An adapter for Slack's Events API.
+ */
+export class SlackEventAdapter extends EventEmitter {
+  /**
+   * The token used to authenticate signed requests from Slack's Events API.
+   */
+  public readonly signingSecret: string;
```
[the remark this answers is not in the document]
clavin: In `events-api` these fields are used in `http-handler`: `createHTTPHandler` accepts the adapter as an argument & accesses these options in a few different places (check for references therein to `adapter`).

I originally typed them as private, however given that they're accessed outside of their class implementation they can't really be "`private`".

There's two ways this *could*<sup>1</sup> be solved. Both involve passing the options that `createHTTPHandler` relies on through its arguments, but both vary on how that's done:

1. Modify the adapter's constructor to not destructure the opts it receives & store the `EventAdapterOptions` as a private field in `SlackEventAdapter`. These options can then be passed to `createHTTPHandler`.

```ts
class SlackEventAdapter {
  constructor(signingSecret: string, opts: EventAdapterOptions) {
    // ...
    this.opts = opts;
  }

  // ...

  public requestListener(): ... {
    return createHTTPHandler(this.signingSecret, this.opts);
  }
}
```

2. Recreate the opts arg (i.e. `{ includeBody: this.includeBody, includeHeaders: this.includeHeaders, ... }`) at the call site of `createHTTPHandler`.

```ts
public requestListener(): ... {
  return createHTTPHandler(this.signingSecret, {
    includeBody: this.includeBody,
    includeHeaders: this.includeHeaders,
    waitForResponse: this.waitForResponse,
  });
}
```

(1) is very tempting to do, however it creates a *very* odd situation. The previous fields (`includeBody`, `includeHeaders`, and `waitForResponse`) can't *really* be removed since somebody could easily be depending on reading them; however, if the pass-through options are stored in a different field, these legacy option fields are now suddenly completely unused 😬.

(2) is a bit tedious as any change in `EventAdapterOptions` needs to be propagated to `SlackEventAdapter#requestListener`, but it's overall the safer option.

Even more, both of these break a previous functionality: if a dev were to change the value of one of these options (i.e. `adapter.includeBody = false;`) *after* calling `requestListener`. Currently, the change would propagate to the HTTP handler, whereas these two solutions only record a single, static instant of these options. This *could* be remedied by making the options passed to `createHTTPHandler` dynamic:

```ts
return createHTTPHandler(() => this.signingSecret, {
  includeBody: () => this.includeBody,
  includeHeaders: () => this.includeHeaders,
  waitForResponse: () => this.waitForResponse,
});

// then in http-handler:
if (adapterOpts.includeBody()) {
  // ...
}

// or:
verifyRequestSignature({
  signingSecret: signingSecret(),
  // ...
});
```

but by this point we're *very* far down the rabbit hole.

Finally, my view on a JS -> TS conversion is that in JS-land (unless previously documented with `@private` or named with a leading `_`) everything is public--intentionally or not. That doesn't<sup>2</sup> hold back adding a `private` access modifier because a JS consumer won't see the difference after compiling--that is, they can still read the field from JS-land. The addition of `private` would only affect a JS user if they converted their codebase to TS (unlikely, but also filled with more roadbumps that this one would be relatively minor).

This doesn't translate to other API-level changes like a name or type change, however. Those kinds of changes propagate to the compiled JS output & therefore break JS code that could depend on the fields.

### tl;dr
The current impl. is very tied up and wasn't built on the same public/private access concerns.

---

<sup>1</sup> I haven't really analyzed these & they could easily break some minute piece of functionality.

<sup>2</sup> In the near future where [private class fields (a stage-3 proposal!)](https://github.com/tc39/proposal-class-fields) exist, TypeScript could theoretically compile `private x` to an `#x` field on classes, making adding `private` to a previously-untyped field a breaking change in the same way that changing its name would be a breaking change.

packages/events-api/src/http-handler.ts
```diff
@@ -0,0 +1,285 @@
+/* tslint:disable:import-name */
```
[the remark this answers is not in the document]
clavin: Another good catch! I definitely left out a `tslint:enable:import-name`.

packages/events-api/src/http-handler.ts
```diff
@@ -0,0 +1,285 @@
...
+   * @param res - The response object.
+   * @returns The responder funciton bound to the input response.
+   */
+  function sendResponse(res: ServerResponse): ResponseHandler {
+    // This function is the completion handler for sending a response to an event. It can either
+    // be invoked by automatically or by the user (when using the `waitForResponse` option).
+    return (err, responseOptions) => {
+      debug('sending response - error: %s, responseOptions: %o', err, responseOptions);
+      // Deal with errors up front
+      if (err !== undefined) {
```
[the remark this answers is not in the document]
clavin: I'll add an `isFalsy` handler to the `util` files of each package based off [MDN's list of Falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). As is, tslint wants booleans in the condition of if-statements.

packages/events-api/src/http-handler.ts
```diff
@@ -0,0 +1,285 @@
...
+  NotFound = 404,
+  Failure = 500,
+}
+
+type HTTPHandler = (req: IncomingMessage & { body?: string, rawBody?: Buffer }, res: ServerResponse) => void;
+
+/**
+ * A response handler returned by `sendResponse`.
+ */
+type ResponseHandler = (err?: (Error & Partial<Omit<CodedError, keyof Error>>) | { status: number }, responseOptions?: {
```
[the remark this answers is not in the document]
clavin: `Omit` was added [in TypeScript 3.5](https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#the-omit-helper-type) as a shorthand for excluding a few keys from an interface/object:

```ts
type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
// "pick the fields of T with keys that do not apply to K"
```

So.. `Partial<Omit<CodedError, keyof Error>>` is "shorthand" for `Partial<Pick<CodedError, Exclude<keyof CodedError, keyof Error>>>`! Hopefully that's simpler 🙃

---

Jokes aside, I don't really remember why it was like this. I believe at some point it had to accept either an `Error` or `CodedError`, but that doesn't explain why it's not just `Error | CodedError`. As an aside, the package never passes an `Error` to a `ResponseHandler`, however the response handler *is* exposed to devs so I believe it should accept an error.

I think it should be `err?: Error | CodedError | { status: number }`. Yeah?

packages/events-api/src/http-handler.ts
```diff
@@ -0,0 +1,285 @@
...
+   * What kind of error occurred.
+   */
+  code: ErrorCode;
+}
+
+/**
+ * Factory for producing a {@link CodedError} from a generic error.
+ */
+function errorWithCode(error: Error, code: ErrorCode): CodedError {
+  const codedError = error as CodedError;
```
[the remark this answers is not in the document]
clavin: We need to be able to set the `code` field on the `error`. If we access `code` (like on the next line) before this, `tsc` will tell us that the property `code` does not exist on the type `Error`.

packages/events-api/src/adapter.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /**
+   * Creates an HTTP server to listen for event payloads.
+   *
+   * @param path - (UNUSED) The path to listen on.
+   */
+  public createServer(path: string = '/slack/events'): Promise<http.Server> {
+    // TODO: options (like https)
+    // NOTE: this is a workaround for a shortcoming of the System.import() tranform
```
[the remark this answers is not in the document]
clavin: I think it should definitely be replaced with something else, as seeing `Promise.resolve().then(() => { ... })` is definitely a pattern that should be explained. I don't know what to write here, though (because I largely don't understand why it's like this 🤷‍♀)

packages/events-api/tsconfig.json
```diff
@@ -0,0 +1,39 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
```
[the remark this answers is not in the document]
clavin: For some reason I thought the test would catch this, but we don't test against 4 😳

Running on Node.js v4.9.1, I found out `"es2017"`, `"es2016"`, `"es2015"`, and `"es6"` all fail because v4 doesn't support destructuring (i.e. `constructor(signingSecret, { x = false, y = false, z = false } = {})`) and default parameter values (and possibly more, that's just the first error).

So the right target is `"es5"` which produces 0 syntax errors in v4.9.1.

packages/events-api/package.json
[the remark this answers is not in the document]
aoberoi: ooo good point. can you file an issue to propagate this change to the other packages which can also benefit from this optimization?

packages/events-api/package.json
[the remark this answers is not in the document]
aoberoi: While I agree that we can get away with not putting any packages whose API is not exposed into `"devDependencies"`, what problems could arise from putting them all in `"dependencies"`?

What I'd like to avoid is a complex heuristic for judging each `@types/*` package individually in order to correctly spread them across the two distinctions. I'm okay with slightly less optimized `npm install` time and `node_modules` size if the win is that we verifiably know that the type definitions are available when needed. In lieu of an integration test that can do that verification, I believe it's easier to just put them all in `"dependencies"` so that it's harder to make a mistake.

packages/events-api/src/adapter.ts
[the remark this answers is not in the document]
aoberoi: thanks for the thorough discussion! i didn't notice that the properties were being accessed by the HTTP handler, so i thought it was unintentional to make these properties public, but you were being very intentional! 🙇 

> Finally, my view on a JS -> TS conversion is that in JS-land (unless previously documented with @private or named with a leading _) everything is public--intentionally or not. [...]

I agree with this point of view. As such I think the change here is perfectly fine. By the time we're ready to revise the public API, we may or may not need some or all of those options, so I think it would be a drain on effort to implement the dynamic getter suggestion just yet.

packages/events-api/src/adapter.ts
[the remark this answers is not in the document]
aoberoi: I see your point. My personal opinion is the TypeScript interfaces to public API should be written to deliver the most sensible experience to another TypeScript author (and that will 99% of the time also give the sensible experience to a JavaScript author).

The actual implementation might allow for different types, but that always has been and will be the reality: values that behave enough like the type expected will pass our runtime checks. Therefore I think we should declare the type as `string` (as this communicates the intention) and still validate through `lodash.isstring` (as this enforces the exact same constraints as before, which holds the API from breaking).

packages/interactive-messages/src/http-handler.ts
clavin: We *could*, and that'd be very correct! But then we'd need to either handle that headers can technically be arrays or add `as string | undefined` in a lot of places.

As an aside, I didn't like this signature because the proper version of this would be `Record<string, string | undefined>` but there is no handling in place for `undefined` values.

packages/interactive-messages/src/http-handler.ts
clavin: I didn't know this type existed 😮 Neat!

packages/interactive-messages/src/http-handler.ts
clavin: I can't say I really know *what* there would be to do after responding & I don't think its feasible to try and route this back to the dev 😬 so it looks like a TODO is the current choice.

packages/events-api/package.json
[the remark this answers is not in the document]
clavin: Sure thing! #802

packages/events-api/package.json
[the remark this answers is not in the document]
clavin: 🤔 Hum. Let me share my process for **solidly** and **verifiably** determining whether a given `@types/` package should be in `"dependencies"` or `"devDependencies"`:

1. `npm run build`
2. check built `.d.ts` files for the `<reference types="xyz" />` comments at the top of the file
  a. packages that are `<reference />`'d should be in `"dependencies"`
  b. all others should be in `"devDependencies"`

Creating a script (e.g. after building) or integration test to do that feels just as simple. Pass any `package/*/dist/**/*.d.ts` file through something that extracts the `<reference />` comments (i.e. programmatically compile as a TypeScript program or just match against the regex `^// <reference types="([^"]+)" />$`), then make sure all and only all referenced types packages are in `"dependencies"` (for its respective `package.json`).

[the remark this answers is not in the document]
aoberoi: That seems like quite a bit of added complexity. What gain are we expecting in return? I can see some benefit from quicker downloads of dependencies, but this is very very small.

I believe its this same cost/benefit analysis that has led the TypeScript project maintainers to recommend just putting them all in `"dependencies"` (which cannot be wrong) instead of splitting the dependencies across the two (which can be wrong unless we write some specific verification tooling, and continue to maintain it).

packages/interactive-messages/src/http-handler.ts
aoberoi: Your concerns with using `IncomingHttpHeaders` are very keen! Can we offer a more useful type than `Record<string, string>`? We can create an interface that has properties `x-slack-signature: string;` and `x-slack-request-timestamp: number;`. In fact, that interface can extend `IncomingHttpHeaders` (where it will get an index signature), and the comment block can add helpful comment about being able to use the `req.headers` where `req` is an incoming request.

packages/interactive-messages/src/adapter.ts
```diff
@@ -0,0 +1,637 @@
...
+    // when a response_url is present,`respond()` function created to to send a message using it
+    const respond: Respond | undefined = payload.response_url ? (message: any): Promise<any> => {
+      if (typeof (message as any).then === 'function') {
+        throw new TypeError('Cannot use a Promise as the parameter for respond()');
+      }
+      debug('sending async response');
+      return this.axios.post(payload.response_url, message);
+    } : undefined;
+
+    let callbackResult: ReturnType<ConstrainedHandler[1]>;
```
aoberoi: this type expression simplifies to `any`

packages/interactive-messages/src/adapter.ts
```diff
@@ -0,0 +1,637 @@
...
+   */
+  public syncResponseTimeout: number;
+
+  /**
+   * Whether or not promises that resolve after the syncResponseTimeout can fallback to a request for the response_url.
+   * This only works in cases where the semantic meaning of the response and the response_url are the same.
+   */
+  public lateResponseFallbackEnabled: boolean;
+
+  private callbacks: ConstrainedHandler[];
```
aoberoi: (the following comment isn't just about this line, but its as good a place to anchor it as any I could find)

I see a few code smells in the way the types have been modeled.

1. The tuple type `ConstrainedHandler` is never used (except where its used to describe the type of the `callbacks` instance variable) without indexing for a certain position (this holds true for each of the constituent types in the union). It seems like its individual items are more important to the APIs and implementation than the tuple type.
2. The return type of `formatMatchingConstraints()` doesn't seem to matter because its always erased (casted away).

Here is a counter-proposal (in a playground so we can see how the type checker evaluates it):

https://www.typescriptlang.org/play/index.html#code/MYGwhgzhAEDK7ANYFkCmUwHNUEEAmYADgC6oBO0A3gFDTQD090AdK7dIWQJYBuYp0YGBAgARmCQQAXNADasYgHsyqPAGFFAOwjEyYLpuIQAPDk0BPDdt37DEAHwAaaGuFiJiALqzPAbnaMLGx0wFo6ZACuwEpkABQAlFTsdHTEABZcEMxCIuKS0AC8cn7sAL4BTKzM7OyEEaIgXMDQEsRcWrHJ0AC2-MAZmphW4bZGMuEGmNAAPtAASqiYAKIAHoQz0DjR7ZrDNgZGjl057kgyW21aABJgmngg5Ed08TLpmUkpKYEAMoqKiD0wM5NIpBJA2oMAIRdEJhYjQABmyl6xFI6jhegOMCKSLIKOQfQGQwxowgsRR-Ume0xdni-k+sOs0B0yjADVQ1NJhWgAHlRAArVDRZiQCBcTCaWK4lFozlY5yUaBpW73cgAFXMhFQMgA5K0djqWjBQkzSnSYQxKmwLYEAFIRHTQJTMtKKADuTuV8PSqCNYolbK4jWI5kEaSFiBgigi8NiBk9vpUwkEijwvreMG6qHSqbBIhgGeYKkwmVIZFcuQ8CXiFsLJzykeYdQgaVishZenZcrsznrHk85oZ0BUxAiZE0nsy9Og5Vq9UazUUJB2ZK6FKJ3bGzN0kw2C2Waw2POXYU3ECeKT7Z0222uKoeZAvL0nMBoQ5Njul-FlJKx3K-xAEsQlKDGe5KElSv60tOnwfvCHZsg8Z7cnygrCqK4qSgBP7WDShxUEq97qpq2rQDqS6XNohqQCmpqDgygRVLWGRZFejbNq27YxIhHJQfhbEDjBKQjmOE4ZtOs50IECzdIoPDoMyihZtAigIgmzJZiIiJkEp0AGGmKy7gcoI+k6ESEA8TokbU3B8AIxaluQFanIgsRwXh0hwDEqhnqYFhnk4eYuTIzkNs+GYfAxVrVEOInji+EnUBU0BqixmwAAoAJLQNgRjMo0mBpMQIChhEmCNI80CiDGemTpoAJcHlbrKOk6k2GmqkIjZvD8L666hVWhBgOYICKGAeAyLc5jPvI3norhpJ+ZYfEOM4A1IJ4GwRHcqAIgYqiRZ8NpMJlYlpVmOZ4L2Wh4I1K7MlEaRGtAAAG7mkswyp3A+GpaoUBRFORJ5US9LQqNADqqDFUXQAA6r6QgTioFkSOm4bdNAbqNU9wAOkoGMhn9mARGAZB4DAEAmV6r3vVioOI4R8ltSRvLAxAyHxlovqiHomj9CpammSat2UUlQ6McEDJxROsiKl9qpkL9pF6remiGqUzgJIU9hUKUJR0OUs6BBcOycnYLR3KzlHsytYO+kNZDwqp6l1A0TQZZlzBwwj0YgHgYZ-BA6agqgKw2NE6lit0Qak9QgScEu5BtAp6T8AmoYtqTvrxqE3SyROxkW3phjkAiqOYxk-M+qGDOiIorWh6QdzONV3rhioOowCCLR4HwhhYNnE54Iou6U9UgRK7AwDcCQKbdA7KckTA8ZgNukTRGOyaxHoPoUKnE4gtHmjCIkbrDYpCYd13oIYQGohBo1oa3QiCLkKgfPoF7mVqW6vq-+DzpG56AjrcYuZYy7AFQEcQIqd4QHFLuXZqEQ-bQE0KgeSFBRC+khv7LQek8rugnLiC2oYcBZWgAAWnwZjX2-t+R42HKgWSTNKbKWRm0SiVUappyFooR0zt9JcBUKrREygWgQyDhQN0rpCIQDjkwZ0o1-gQ3WK6D0qBsbkAFgmIOYCEGQJgFnaAaY9poLwNUagodCAtT0WQCBvoTZaGQm+S8bgGyZTwAAfnGDuQYe5FirEIEJBoigkAeO8WvXcsx9yBKEvqLQ4SfHcD8dEgJax-BdEJqgCJExBj+FnFYmx8C7Hl2PNbZxxw3EeESZElJ8w0lBK6CEsJXiklRPqQeRpnx4maBqbkqYqTOkZM+FjN4mgIk6maYgAA+j0iAhpZg6mKfqeS0yswYGwAssit1hCKEwDqags4smbH8rbIojjdi21mGUlcZ4MmBAUMoHyts0wQGnlwLBBZwy2OPiAbcrJsCiLxGnZ2U1aIjAOFZLU1BjmPJUPNCFdhjBqGgI3d+5MTnLQWliHWRQUUADJDqEW+sRLUuoelbKBtbHUQlAgeNQG4cwzhf4pmQf7SG4jhZ3S0MmY5zoICjKrt8454gg64LOoPOo8JabmyyReKSTBW5VSFGADloLoAIABM7UOrQoXoGcKZdBwgSacOHugNW8JmpkABKZM+oYWVn0MNQ50VrEDQwVcShWSsQqorDuimANzTy208WRCiK5DRSC6IEIcMaUVoruDAC5yEQ0q0ohGqNTAY1DjQRg-JSUETbREQBICIFiTYqRcheNGKzBYsRUYewnRPjrkguWrc-T-GdI2Geagz4KmfC4GpcCwENxnIBhDHapiDrTFmM20Co6iiaGQSARILihzpB0h6NBHolZLDIDpOIOpkKIxBPCLB46TH7VwRQRdIgdT0QNl0B48IRznFOa27EBFShCQHdAWIWTnazrLXW7EY7O6+P2TMGdEE53vuLjoW4kDnYxLWCui0dARzZCqS07kgG7ldFKKikAujV1S2zChAUQpiAiigJhWIlANaAmHS24D96ZxdGlow+ENE8NGyYKldAvojGhEIFwA6cktEYPEG0boMKWYLAgNYy2RRyToAgAPSaFhEgFB1ulHS0cg7GCmvYfwxyLk3BJRQZTQ0RpjQmiQ5wKgFM3RkPJxTeAtM6zBbMXTSlMioEMxYHWsxtoXrMSZlmgbtDmYVtyWI1nRrjQ09NbWJCNg+f0-5ozW0J2XvC39daAJzmq2iw+I8bMSvkAyUAA

aoberoi:
Note: Coverage looks like it's decreasing, but that shouldn't block merging this PR. This is essentially a rewrite so the previous scores shouldn't weigh too much. As long as the new coverage is at an acceptable level (IMHO >90%), we should continue to merge.

aoberoi:
@clavin I just want to point out that there are two unresolved discussions from my previous code review, plus the new discussions created just now. I think we're super close and I'm so excited to be 100% TypeScript!

giphy (1)

packages/interactive-messages/src/http-handler.ts
clavin: 😕 But then users would need to pass `req.headers as VerificationHeaders` to assert that those headers "will" be present.

(Also, it'd be `'x-slack-request-timestamp': string`)

packages/events-api/package.json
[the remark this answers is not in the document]
aoberoi: Can we agree that the idea to separate `@types/*` dependencies according to whether or not they are re-exported via any of the package's own interfaces can become a separate proposal (which could apply to each of the packages)?

If so, I'd like to move forward by keeping all `@types/*` dependencies in `"dependencies"`, and then carry out that discussion in a separate issue. I think that will make it easier for others to weigh in, too (this discussion is buried in the code review of one PR). If we reach a different conclusion there, we can make the updates across all the packages at once.

packages/interactive-messages/src/http-handler.ts
aoberoi: At face value, it does seem like it would be unergonomic. But in my opinion, the type assertion is a **good thing** for people's code. It forces a user to understand that using `verifyRequestSignature()` requires them to think about whether those specific headers will be present as they write the call site.

For the majority of the users of this package, they will not use `verifyRequestSignature()` directly. It should only "get in the way" for advanced users, who I assume would appreciate a little extra guidance.

(nice catch!)

packages/events-api/src/http-handler.ts
```diff
@@ -59,18 +60,20 @@ export function createHTTPHandler(adapter: SlackEventAdapter): HTTPHandler {
     return (err, responseOptions) => {
       debug('sending response - error: %s, responseOptions: %o', err, responseOptions);
       // Deal with errors up front
-      if (err !== undefined) {
+      if (!isFalsy(err)) {
         if ('status' in err) {
           res.statusCode = err.status;
-        } else if (err.code === ErrorCode.SignatureVerificationFailure ||
-            err.code === ErrorCode.RequestTimeFailure) {
+        } else if ('code' in err && (
```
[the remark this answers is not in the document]
clavin: `err` here is typed as `err?: Error | CodedError | { status: ResponseStatus }`. The handlers are checked in reverse order of that union:

1. `{ status: ResponseStatus }` by checking `'status' in err`
2. *some* codes of `CodedError` by checking `'coded' in err` and if a specific code is present
3. the remaining codes of `CodedError` or `Error`
4. (`undefined` is ruled out right before this if-else chain)

This was indeed added to satisfy the typechecker because once `{ status: ... }` is eliminated, `code` may or may not exist on `Error | CodedError`. As far as I know, adding `'code' in err` to this condition should not change its truthiness.

I don't know if there's a better way to describe the type of `err` nor what it would be 🤷‍♀.

packages/events-api/src/http-handler.ts
```diff
@@ -211,12 +214,12 @@ enum ResponseStatus {
...
 }
 
-type HTTPHandler = (req: IncomingMessage & { body?: string, rawBody?: Buffer }, res: ServerResponse) => void;
+type HTTPHandler = (req: IncomingMessage & { body?: any, rawBody?: Buffer }, res: ServerResponse) => void;
 
 /**
  * A response handler returned by `sendResponse`.
  */
-type ResponseHandler = (err?: (Error & Partial<Omit<CodedError, keyof Error>>) | { status: number }, responseOptions?: {
+export type ResponseHandler = (err?: Error | CodedError | { status: number }, responseOptions?: {
```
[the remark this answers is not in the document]
clavin: No, because when `respond` is called with an object that has `status`, that object is *not* a `CodedError`; that is the call signature usually looks like `respond({ status: ResponseStatus.XYZ }, ...);`.

packages/events-api/src/adapter.ts
```diff
@@ -119,23 +120,18 @@ export class SlackEventAdapter extends EventEmitter {
 
   /**
    * Returns a middleware-compatible adapter.
-   * @param middlewareOptions - (UNUSED)
    */
-  public expressMiddleware(
-    middlewareOptions: object = {},
-  ): (req: IncomingMessage, res: ServerResponse, next: () => void) => void {
-    const requestListener = this.requestListener(middlewareOptions);
+  public expressMiddleware(): (req: IncomingMessage, res: ServerResponse, next: () => void) => void {
```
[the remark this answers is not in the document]
clavin: Erm, this begs the question: what's the difference between `expressMiddleware` & `requestListener`?

Also, changing the return type of this method to a function that accepts only `req` and `res` means that the returned function only has a `length` of 2 (instead of the current 3). `express` *does* read this number (last I checked) when determining whether a middleware should be run serially or async, and I'm afraid of downstream apps somehow depending on this.

packages/interactive-messages/src/adapter.ts
clavin: Thank you for writing this all this out! 🤩 I agree that there was definitely spaghetti types going on before. I think half of it was me not wanting to change *any* of the functionality of the code.

All of your occurrences of `StoredConstraints` was used with `AnyConstraints` as the generic type, so I merged the two into a union type. This made the expanded signature of `matchCallback` to be more graceful. Because of it being a union instead I didn't really have to change the body of this function anymore to guard the typechecker into figuring out what type of constraints were being used.

As a side effect, this also drops the `lodash.isplainobject` dependency. 🎉

packages/events-api/package.json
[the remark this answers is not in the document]
clavin: Sounds valid to me! ➡️ #830

packages/events-api/src/http-handler.ts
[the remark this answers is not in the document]
43081j: FYI i was browsing around and saw this... a `ResponseHandler` below is `Error | CodedError | {status: number}` but that isn't quite right. You shouldn't be specifying the redundant type here, it should really be `Error | {status: number}` because a `CodedError` _is_ an `Error`. 

Because you already have a union there, you already have to do a check to figure out if its a coded error, making its involvement in the type redundant AFAIK.

packages/interactive-messages/src/util.ts
```diff
@@ -0,0 +1,62 @@
...
+  );
+  return Object.keys(identifierMap).reduce((acc, k) => `${acc} ${escape(k)}/${escape(identifierMap[k])}`, '');
+}
+
+/**
+ * Tests a "thing" for being falsy. See: https://developer.mozilla.org/en-US/docs/Glossary/Falsy
+ *
+ * @param x - The "thing" whose falsy-ness to test.
+ */
+export function isFalsy(x: any): x is 0 | '' | null | undefined {
```
43081j: out of interest, why? everywhere you use this seems like it'd be happy with `!theThing`. are you trying to work around the type system or something?

clavin: Mostly working around the linter, actually. It could be better to just disable that rule for this project.

packages/events-api/src/http-handler.ts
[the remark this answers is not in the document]
clavin: If I was refactoring this code, I'd agree! But one of my goals was to try and touch as little functionality as possible and simply add types. But this is definitely a good point, and I think it would ultimately be the right way to do this 👍

I also think there's some value in having `CodedError` in the type signature to suggest there is different functionality for (some) `CodedError`s that is different than for `Error`s. Yeah or no?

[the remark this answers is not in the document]
43081j: in an ideal world i think it'd actually be just `Error`. you'd usually do this so any type of `Error` can be thrown, even ones the handler may not support (but at least understands the base interface).

it should be up to the handler to decide if it wants to handle specific kinds of `Error`.

similar to how DOM event handlers work (most take an `Error` unless they are known events which _always_ fire a specific kind of error, e.g. Click).

packages/interactive-messages/src/util.ts
43081j: which rule was it and whats an example of the line that caused it? there's probably a cleaner solution i suspect

packages/events-api/src/http-handler.ts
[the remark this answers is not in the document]
aoberoi: I agree with @43081j on this one. The intention of the `ResponseHandler` type is to represent a function that the user can provide any `Error` as the first argument, and the specific behavior that occurs when subtypes are provided is up to the library.

I do think that the documentation should describe that behavior. I think we can add a little more description to L106 to make this clearer.

In fact, I think we might want to go a little further and take `| { status: ResponseStatus }` out of the type, but I'll reserve that discussion for [another part of the review](https://github.com/slackapi/node-slack-sdk/pull/799/files#r303687812).

packages/events-api/src/http-handler.ts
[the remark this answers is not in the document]
aoberoi: I've been able to jog my memory about the design and intention of the API of a response handler function. 🤔💭 

From the [documentation page](https://slack.dev/node-slack-sdk/events-api#custom-responses):

> When the event was handled normally, call respond() with no arguments. If there was an error, call respond(error) with an object that has a status property set to a valid HTTP status code.

For better or worse, `status` is seen as a property of the `error` argument. Yes, we could discuss that in the next major version iteration of this package/API, `status` should be considered one of the possible `responseOptions`, but this is where we are.

In fact, there is some precedent behind this design. In the [`finalhandler` package](https://github.com/pillarjs/finalhandler#finalhandlerreq-res-options), which is used by Express to handle all errors that aren't explicitly handled already, the `error.status` value is used to set the response status code. The main reason I could find for this design choice is so that when a server is running in `NODE_ENV=development`, Express can use the status code the application desired for the error and still display a stack trace in the browser (since the stack trace is provided by `Error` objects). Adhering to this design makes this package more compatible with express response handlers and should make for easier debugging. There are gaps in how this package has implemented this feature (such as only accepting this for 4xx and 5xx status codes), but that's okay.

In conclusion, I think we should define a new `ResponseError` type as an interface that extends `Error`, which has an optional `status?: number;` property. The type of `ResponseHandler` would then change to the following:

```typescript
export type ResponseHandler = (err?: ResponseError, responseOptions?: {
  failWithNoRetry?: boolean;
  redirectLocation?: boolean;
  content?: any;
}) => void;
```

Even though [I said earlier](https://github.com/slackapi/node-slack-sdk/pull/799/files#r305971898) that the first argument should just be an `Error` object, I believe this preserves the same benefit that all `Error`s are a `ResponseError` because `ResponseError` only defines optional properties. In a structural typed language like TypeScript this is a valid design decision, whereas in a nominally typed one we would run into some trouble.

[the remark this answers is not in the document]
aoberoi: @43081j would you like to weigh in on this point?

packages/events-api/src/adapter.ts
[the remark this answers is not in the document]
aoberoi: The only difference should be the arity (number of arguments expected) of the function. `expressMiddleware` should have an arity of 3, which is what the `RequestHandler` type has. `requestListener` should have an arity of 2, which is what the `RequestListener` type has.

packages/interactive-messages/src/adapter.ts
aoberoi: Sounds good! Would love to see the result of this update.

packages/events-api/src/http-handler.ts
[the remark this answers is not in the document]
43081j: it would be a breaking change (from a types POV) i guess? as you would no longer be able to pass a plain object (`{ status: 400 }`). If we type it as a `ResponseError` (which is `interface ResponseError extends Error { status?: number; }`), you'll have to always pass an error.

That does sound right to me but it seems it'd mean the types no longer permit what the source does.

You could also have `type ResponseError = (Error & {status?: number}) | {status: number}`.

But that is _almost_ like `err?: Error | {status: number}` except the error can have a status.

packages/events-api/src/adapter.spec.js
```diff
@@ -1,29 +1,30 @@
-var http = require('http');
-var assert = require('chai').assert;
-var sinon = require('sinon');
-var noop = require('nop');
-var getRandomPort = require('get-random-port');
-var EventEmitter = require('events');
-var systemUnderTest = require('../../dist/adapter');
-var createStreamRequest = require('../helpers').createStreamRequest;
-var SlackEventAdapter = systemUnderTest.default;
+require('mocha');
```
43081j: this shouldn't be needed, mocha sets the globals its self

packages/interactive-messages/src/http-handler.ts
```diff
@@ -95,24 +96,21 @@ export function createHTTPHandler(adapter) {
...
-        const rawBody = r.toString();
+      .then((bodyBuf) => {
+        const rawBody = bodyBuf.toString();
 
-        if (verifyRequestSignature(adapter.signingSecret, req.headers, rawBody)) {
+        if (verifyRequestSignature(adapter.signingSecret, req.headers as VerificationHeaders, rawBody)) {
           // Request signature is verified
           // Parse raw body
-          const body = parseBody(rawBody);
+          const body = parseBody(rawBody) as any;
```
43081j: `parseBody` already returns `any` (but see my comment there too)

packages/interactive-messages/src/http-handler.ts
```diff
@@ -40,13 +38,13 @@ export function createHTTPHandler(adapter) {
   /**
    * Parses raw bodies of requests
    *
-   * @param {string} body - Raw body of request
-   * @returns {Object} Parsed body of the request
+   * @param body - Raw body of request
+   * @returns Parsed body of the request
    */
-  function parseBody(body) {
+  function parseBody(body: string): any {
```
43081j: can we make some better assumptions around the type here? maybe at least `Record<string, any>`?

as long as know `payload` is always a plain object or undefined (not an array/number/string).

packages/interactive-messages/src/util.ts
clavin: For example, the [check inside of `sendResponse`'s return function](https://github.com/clavin/node-slack-sdk/blob/d54766ccc903360c84e6c5bad21b58e6ce959026/packages/events-api/src/http-handler.ts#L63), if it were:

```typescript
if (err) {
  // ...
}
```

Would cause the linter warning:

```
This type is not allowed in the 'if' condition because it could be undefined. Allowed types are boolean or boolean-or-undefined. (strict-boolean-expressions)
```

packages/interactive-messages/src/http-handler.ts
clavin: @aoberoi 👀💭☝️❓ (I'm impartial on this)

packages/interactive-messages/src/util.ts
43081j: yeah in all honesty its OTT. if my PR merges too, tslint will be removed (as it should be these days) and ESLint doesn't recommend such a rule by default.

i'd recommend you turn the rule off entirely in `tslint.json` for now and drop this helper function. it really is over the top for what is being done here (a check to see if an error is set). if you really wanted to be specific, you'd just do `err !== null && err !== undefined` everywhere but is there really a need?

packages/interactive-messages/src/http-handler.ts
43081j: FYI too the reason being we cast elsewhere in the file where we really shouldn't have to. also `any` is best avoided really, especially with the introduction of `unknown` these days. it is better and safer to be more specific about the type or at least use `unknown` and do stricter checks before using the value.

packages/interactive-messages/src/util.ts
aoberoi: @43081j what you're proposing is a change to the project's style guide. that's welcome! i just don't think this specific PR is the right place to start enforcing that change. i think it would make more sense to first discuss it, and if we all agree on the change then create a new PR which applies it across the codebase.

packages/interactive-messages/src/http-handler.ts
aoberoi: This function returns the result of `JSON.parse()`, [which is an `any`](https://github.com/microsoft/TypeScript/blob/d982014d733445f478b5cad8a938d2d2a56f1080/lib/lib.es5.d.ts#L1053-L1059). Using any other type would require a cast here, so I don't agree with the motivation of removing casts. I think the cast being made local to the call site of `parseBody()` has the benefit of keeping that knowledge of what the input was close to the location of the cast.

I think the TypeScript team wasn't just lazy and did this for good reason. `any` is actually a better model for an arbitrary JavaScript object than the other options. We had a similar discussion about using `object` to describe this type of object. Specifically compared to [`unknown`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type), the following properties make it a less correct model for this value:

```
// keyof unknown

type T40 = keyof any;  // string | number | symbol
type T41 = keyof unknown;  // never
```

The keys of an arbitrary JavaScript object are actually `string | number | symbol`. That may not be true for every kind value that `JSON.parse()` produces, but since `string` is a subset, that is still true. Whereas `never` is actually untrue.

packages/interactive-messages/src/util.ts
43081j: Am I? In the current sources does this helper function exist? I was under the impression it doesn't and had been introduced pretty much to work around the linter or compiler. 

In which case the particular lines triggering the error should really either do a more specific check or should turn the rule off for themselves. 

These functions are typed to accept an error or null usually. It shouldn't be possible to accept some non-null falsy value. Checking `!err` was fine to be honest.

packages/interactive-messages/src/adapter.ts
```diff
@@ -0,0 +1,652 @@
...
+/**
+ * A function used to send message updates after an action is handled. This function can be used
+ * up to 5 times in 30 minutes.
+ *
+ * @param message - a [message](https://api.slack.com/docs/interactive-message-field-guide#top-level_message_fields).
+ *   Dialog submissions do not allow `resplace_original: false` on this message. @returnsthere's no contract or
+ *   interface for the resolution value, but this Promise will resolve when the HTTP response from the `response_url`
+ *   request is complete and reject when there is an error.
+ */
+type Respond = (message: any) => Promise<any>;
```
aoberoi: This might be better described as returning a `Promise<unknown>`. The idea is that the user calls this function and knows it will complete asynchronously, but knows nothing about the value that the returned Promise resolves to. We actually don't want the user to depend on that value at all (it should be treated as opaque). An `any` would indicate its some meaningful value but the type is anything.

packages/interactive-messages/src/adapter.ts
```diff
@@ -0,0 +1,651 @@
...
+type StoredConstraints =
+ | ({ handlerType: StoredConstraintsType.Action } & ActionConstraints)
+ | ({ handlerType: StoredConstraintsType.Options } & OptionsConstraints);
+
+/**
+ * A function used to send message updates after an action is handled. This function can be used
+ * up to 5 times in 30 minutes.
+ *
+ * @param message - a [message](https://api.slack.com/docs/interactive-message-field-guide#top-level_message_fields).
+ *   Dialog submissions do not allow `resplace_original: false` on this message. @returnsthere's no contract or
```
aoberoi:
```suggestion
 *   Dialog submissions do not allow `replace_original: false` on this message.
 * @returns there's no contract or
```

packages/interactive-messages/src/adapter.ts
```diff
@@ -0,0 +1,651 @@
...
+/**
+ * The type of stored constraints.
+ */
+const enum StoredConstraintsType {
+  Action = 'action',
+  Options = 'options',
+}
+
+/**
+ * Internal storage type that describes the constraints of an ActionHandler.
```
aoberoi:
```suggestion
 * Internal storage type that describes the constraints of an ActionHandler or OptionsHandler.
```

packages/interactive-messages/src/adapter.ts
```diff
@@ -0,0 +1,651 @@
...
+    }
+
+    // The following result value represents:
+    // * "no replacement" for message actions
+    // * "submission is valid" for dialog submissions
+    // * "no suggestions" for menu options TODO: check that this is true
+    return Promise.resolve({ status: 200 });
+  }
+
+  private registerCallback(constraints: StoredConstraints, callback: Callback): this {
```
aoberoi: you're right, this is better!

packages/interactive-messages/src/adapter.ts
```diff
@@ -24,10 +25,10 @@ function formatMatchingConstraints<C extends AnyConstraints>(matchingConstraints
   }
 
   let ret: AnyConstraints = {};
-  if (typeof matchingConstraints === 'string' || matchingConstraints instanceof RegExp) {
-    ret.callbackId = matchingConstraints;
+  if (!isPlainObject(matchingConstraints)) {
```
aoberoi: out of curiosity, which test was failing? its a shame we couldn't axe this dependency 😿

clavin: There were a few tests that would pass a `number` (and some other invalid types) to `actions()` and `options()` to make sure it threw a `TypeError`. RIP 💀

clavin:
For reference: I plan on resolving conflicts at the end, after all my changes are made 🙆

packages/interactive-messages/src/util.ts
aoberoi: > In the current sources does this helper function exist? I was under the impression it doesn't and had been introduced pretty much to work around the linter or compiler.

You're right that this function does not exist previously. But its addition is to help conform the code to satisfy the *linter*, not the typechecker. The linter's job is to enforce the project's style guide.

> In which case the particular lines triggering the error should really either do a more specific check or should turn the rule off for themselves.

As far as I can tell, this function encapsulates what you've called "a more specific check" so that its not repeated several times. The alternative, "turn the rule off" is updating the project's style guide. So I think we agree on this point.

> These functions are typed to accept an error or null usually. It shouldn't be possible to accept some non-null falsy value. Checking `!err` was fine to be honest, or `!== null`

I'm not sure which functions you're referring to. If you mean the conditions where this `isFalsy()` function is being called, then these are not typically functions, they are typically the conditions of an `if` statement.

Whether or not it accepts a non-null falsy value isn't something we can change today without changing the API contract. Even if we agree that the intention is that `null` or `undefined` are the only falsy values we care about, someone who uses this library may not have agreed, and therefore may be triggering a code path in their application that uses some other falsy value. We're trying to apply this change without breaking any existing code, so this function is a useful hygiene utility to accomplish just that.

packages/interactive-messages/src/adapter.ts
aoberoi: darn!

packages/events-api/src/http-handler.ts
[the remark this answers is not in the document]
aoberoi: @43081j i think you're right that this would alter the API contract. however, a JavaScript developer using this package would never experience a breakage. We aren't adding any code that uses properties of the argument that are unique to `Error` when a status is provided.

And until this point, every user of this package was using it as a plain JavaScript package. We never offered typings. If third party typings were used and they disagreed with these, we wouldn't know and there's no way for us to align with them.

Therefore I think this small alteration of the contract is safe.

packages/interactive-messages/src/util.ts
43081j: What I mean is the functions which make use of this workaround function are just trying to check if an error is set or some other value is set. 

More than likely you should be checking specifically if the value is null or undefined. That is what the linter really wants, this function is a workaround catch all to pretty much bypass the lint rule IMO.

packages/interactive-messages/src/http-handler.ts
43081j: I understand exactly that JSON parse is any and it definitely should be. 

My question was if you can make any assumptions about the payload. If it really can be anything, the answer is no. But I wasn't sure if you knew that payloads here have a known structure.

packages/interactive-messages/src/util.ts
aoberoi: > More than likely you should be checking specifically if the value is null or undefined.

We're intentionally choosing not to depend on any sort of probability. We are trying to produce a change that cannot impact the API of this package in a breaking manner (no matter how badly the user abused the API). This function provides the exact semantics of the "falsy" concept without circumventing the linter or changing the style guide - so it works.

You're right that there are probably places where this gets used that don't need as extensive of a check. But if we're going to encapsulate it in a function for a couple places, we might as well apply that function throughout for consistency (and we get even more confidence that we didn't break the API).

aoberoi: If we were open to API breaking changes, I would agree that we should simplify so that this function would become unused and vanish.

packages/interactive-messages/src/http-handler.ts
aoberoi: 🤔 If we wanted to produce a type that could describe the structure any more specifically than "the type that `JSON.parse()` returns", I think it would need to model the kinds of payloads Slack might deliver over the Interactive Components part of the platform. That's a worthy idea! In fact, [we've done much of that work over in Bolt](https://github.com/slackapi/bolt/tree/master/src/types/actions). IMHO, the `Record<string, any>` model doesn't add much value over `any` - in fact I think it's not really what the `Record` type is meant for. The one thing you'd get is the ability to index the value with strings, [but the `any` type already allows you to do this](http://www.typescriptlang.org/play/#code/CYUwxgNghgTiAEYD2A7AzgF3gDwFzyhQE8BuAKDOXSQhADoIkBzACmzoDMkkBKcs0JFgIqmeEXwAlcEhjAAPJhgBLFEwA0BYgD5+omvUasidAEaw+FIA). Side note: if I wanted a way to do this without `any`, I'd likely reach for an interface with an index type instead of `Record`, which I think communicates a very different intention.

So, can we give this a type that actually adds value? Tactically, we've made the choice to extract common Slack types that are used across many of our packages in `@slack/types` so that third party packages (ones that Slack doesn't maintain) may also depend on them. We haven't done much to promote that yet, because it's still early days. We probably want to do some more testing before we start encouraging that type of use so that certain usages don't regress (there's not a lot of precedent for how to do semver for the types of a package). Long story short: I think there's an opportunity to move many of the types currently in Bolt into `@slack/types`, and then to use them in this package. That would be cool!

Now that this PR has landed, if we want to continue this discussion, we should create a new issue for it. I'd be happy to hear more.

PR #803, opened 2019-06-28
Add WebClient test cases for invalid Retry-After headers

copperwall:
###  Summary

This adds test cases for when there's no Retry-After header on a 429
response and for when there's an invalid Retry-After header on a 429
response.

This also includes a test case to assert that buildResult includes
retrySec in the response_metadata object of the data response. This
test case is kind of finicky though, because it looks like in the
case of a 429 response, that particular block isn't run because
the makeRequest function either throws an error if rejectRateLimitedCalls
is true or waits and trys again, in which case there won't be a Retry-After
header on the non-429 response after the retry timeout ends.

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -817,6 +817,18 @@ describe('WebClient', function () {
         });
     });
 
+    it('should set retrySec info on the response_metadata object', function () {
+        const scope = nock('https://slack.com')
+          .post(/api/)
+          .reply(200, { ok: true }, { 'retry-after': 100 });
+        const client = new WebClient(token);
+        return client.apiCall('method')
+          .then((data) => {
+            assert(data.response_metadata.retryAfter === 100);
+            scope.done();
+          });
+    });
+
     it('should pause the remaining requests in queue', function () {
       const startTime = Date.now();
       const retryAfter = 1;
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.spec.js
```diff
@@ -817,6 +817,18 @@ describe('WebClient', function () {
         });
     });
 
+    it('should set retrySec info on the response_metadata object', function () {
```
clavin: I think this test could use a different description that describes the *functionality* it's testing, rather than what the code in the test does 🙂. The current description might be confusing if a new contributor makes a change and sees that this test doesn't pass, only to find that `retrySec` isn't part of the test body 😕.

 It looks like this test checks that some(?) metadata is present on a response. A different description would make this a stellar test 🤩!

copperwall: Awesome, thanks for the feedback. I'm on it!

copperwall:
Hey @clavin, I updated the test description to describe in what case there should be retryAfter metadata on the `WebClient.apiCall` response. If you have any opinions on how to make that better I'm happy to make some changes too. Thanks for the feedback!

packages/web-api/src/WebClient.spec.js
```diff
@@ -856,9 +868,35 @@ describe('WebClient', function () {
...
   });
 
+  it('should throw an error if the response has no retry info', function (done) {
+      const scope = nock('https://slack.com')
+        .post(/api/)
+        .reply(429, {}, { 'retry-after': undefined });
+      const client = new WebClient(token);
+      client.apiCall('method')
+        .catch((err) => {
+          assert(err.message.match(/Retry header did not contain a valid timeout/i) !== null);
```
clavin: 🤔 In case this message ever changes, I think it's good to replace this with something that [just makes sure `err` is an `Error`](https://github.com/slackapi/node-slack-sdk/blob/7247d387de3645415eac83a41d8ec0116e3730f1/packages/web-api/src/WebClient.spec.js#L164). Ditto on the other new test case below this one 👇.

copperwall: That makes sense to me, thanks for including a link to that other example.

aoberoi:
@copperwall @clavin i'm just catching up on changes and i'm glad this got merged! 

i was wondering if either of you agree that the two last tests could be inside the `describe('has rate limit handling', { ... });` block. i think that any functionality that has to do with the `retry-after` response header is roughly grouped under the concept of rate limit handling.

PR #804, opened 2019-06-30
add tests to logger package

43081j:
felt like adding some tests to this package, let me know if you think they're worth merging in.

couple of oddities i noticed along the way you might want to think about in future:

* most of your tests are in JS but can be in typescript as you already pass them through 'ts-node'
* your types (e.g. `@types/mocha`) are often or always unused because of this
* weirdly you get the older (typescript <3.1) sinon types, im not too sure this is as you're on a recent enough typescript, maybe its some mistake by ts-node

### DIFF
--- a/packages/logger/.nycrc.json
+++ b/packages/logger/.nycrc.json
@@ -1,0 +1,14 @@
+{
+  "include": [
+    "src/**/*.ts"
+  ],
+  "exclude": [
+    "src/**.spec.js"
+  ],
+  "reporter": ["lcov"],
+  "extension": [
+    ".ts"
+  ],
+  "all": false,
+  "cache": true
+}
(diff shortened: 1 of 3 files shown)


packages/logger/src/index.spec.ts
```diff
@@ -0,0 +1,134 @@
+import { assert } from 'chai'; // tslint:disable-line:no-implicit-dependencies
```
aoberoi: this PR also pushes for tests to be in TypeScript, and [we want to move that decision](https://github.com/slackapi/node-slack-sdk/pull/808#issuecomment-530593159) outside unrelated PRs.

aoberoi:
this is really awesome @43081j, thanks for taking the time to implement the tests!

43081j:
@aoberoi since i made this PR, a couple of tests got added.

i've thrown away the typescript change, rebased onto master and tried to merge the existing tests with mine. there was a confusing one in there which i deleted (see the diff) as i _think_ whatever it was trying to assert is covered by the other tests

seratch:
The change in this PR has been merged by #1269 - Thanks for your contribution!

PR #805, opened 2019-07-01
introduce eslint

43081j:
TSLint has been deprecated for some time and non-de-facto for even longer since ESLint introduced typescript support.

It would definitely make sense for you guys to drop it from all packages and introduce eslint instead.

This happens by:

* Introducing `eslint`, `@typescript-eslint` (plugin and parser) and the regular eslint airbnb config
* Enabling any equivalent eslint rules to the previous tslint ones we had
* Dropping any rules now covered by defaults or typescript its self

Let me know if you're interested in this and ill do the other packages too.

edit: In fact i would recommend you simply have eslint as a top-level dependency and lint the entire monorepo (by extension) to save you having duplicated configs and dependencies.


43081j:
FYI i did make another branch where i did:

* introduce top-level eslint and lint `packages/*/src` instead of having eslint per-package
* introduce a top-level tsconfig which all others inherit
* move a bunch of tests to be typescript
* introduce missing `@types` for the typescript tests
* fix all the lint errors (there are a lot) except the events-api and interactive-messages as they are being rewritten at the min in another PR

it seems to work well but depends if you prefer having linting at top level of the monorepo or in each individual package.

if you don't want that or this PR though thats fine too, i.e. if you're happy with it as it is now

43081j:
i also merged my querystring branch into this which removes the no longer useful `querystring` dependency.

aoberoi:
yay! ditto to this comment: https://github.com/slackapi/node-slack-sdk/pull/808#issuecomment-510082878

43081j:
going to close this because we can make any necessary changes in #808 instead, even if we were to decide against a global lint task

PR #808, opened 2019-07-07
introduce ESLint at top-level and unify tsconfigs

43081j:
This is same as #805 but:

* ESLint is a top level dependency
* Linting is done on a glob, not per package (so the dep is installed once and eslint runs once)
* Unified all tsconfigs to inherit a base one
* Moved a test or two to be TS rather than JS

Ill leave both of these PRs around, let me know if you want either of them. I will close them if not.


aoberoi:
Thank you for sending us these contributions! Moving to ESLint is in fact something we have hoped to do, especially since the announcement of the TSLint and ESLint projects' intention to "merge".

I haven't had a chance to review these PRs just yet, but it has sparked the interest of a few members of the team. I just wanted to let you know it might take us a few more days before we're ready to engage.

43081j:
Yup thats no worries.

This one is preferred to the other PR because it fixes all packages rather than select ones, and it means you only need eslint once.

ill fix the build problems today too (dependency weirdness)

packages/rtm-api/src/RTMClient.ts
```diff
@@ -463,14 +473,15 @@ export class RTMClient extends EventEmitter {
   public addOutgoingEvent(awaitReply: true, type: string, body?: {}): Promise<RTMCallResult>;
   public addOutgoingEvent(awaitReply: false, type: string, body?: {}): Promise<void>;
   public addOutgoingEvent(awaitReply: boolean, type: string, body?: {}): Promise<RTMCallResult | void> {
-    const awaitReplyTask = (messageId: number) => {
+  /* eslint-enable jsdoc/require-jsdoc,lines-between-class-members */
```
43081j: `>>`

clavin:
😯 I've been thinking about a change like this (moving to ESLint & hoisting configuration files) a lot over the last month. I'm definitely behind a change like this! 👏🙇‍♀️

🤔 I think having a root-level `tsconfig.json` in a monorepo like ours is a bit misleading because it labels the monorepo as a whole as a TypeScript program. This *could* create an issue with tools that crawl through parent directories to find `tsconfig.json` files. This is somewhat of a trivial nit because just renaming it to `tsconfig.base.json` resolves this.

I'm still doing more thinking about hoisting `tsconfig.json` files from each package & what implications it has (e.g. on each package's individuality & modularity). Everything else is 👌 ✨. I'm curious, maybe as an extension after this PR, if it's possible to also hoist the `.nycrc.json` and `.mocharc.json` files from each package.

43081j:
You are right. For now, i'll rename it at least and we can figure that out at some point.

it doesn't have to be part of this PR if we can't settle on how it should be. i have seen it done this way in a few monorepos but usually named like you suggested

the mocharc is an odd one for me as its a different approach to testing than im used to. in most repos, i would expect to build (`tsc` in most cases) then point mocha at the built tests (`lib/**/*.spec.js` for example).

instead, you pass the sources into mocha and use ts-node to intercept it. this produces some oddities around types (as ts-node doesn't load `includes` from tsconfig, leading to you having `paths` set) and seems like a bit of an extra layer to me. if you instead passed built JS to mocha (with sourcemaps), you'd no longer need the mocharc files.

its personal pref. more than anything though i think. here i left it as is because its consistent with what was there already

43081j:
do you also have any idea why there's a problem loading the internal dependencies?

```
npm i
npx lerna bootstrap
```

this results in things like `@slack/types` not being resolved (as you can see in the CI builds too).

clavin:
It's because all the build artifacts (the `dist/` directory) is being built relative to the top-level `tsconfig`. Any path in the top-level `tsconfig`, like `outDir`, will be relative to **that** `tsconfig` file, not to the `tsconfig` that inherits it.

In case that isn't clear, the directory tree looks kind of like this:

```
/
- tsconfig.base.json
- packages/
  - package-name/
    - tsconfig.json
```

In `packages/package-name/tsconfig.json`, we have an `"extends": "../../tsconfig.base.json"`, so it inherits all properties of that extended config file. In `tsconfig.base.json`, we have `"outDir": "dist"` which is resolved relative to *that* `tsconfig` file it's in:

```
/
- tsconfig.base.json
- dist/            <--- the base config points here!
- packages/
  - package-name/
    -              <--- not here!
    - tsconfig.json
```

In each package, the `package.json` has a `"main"` that points to `"dist/index.js"` which would be inside of the package's directory, not at the top level.

This discrepancy of where `dist/` is pointing to is causing these errors.

---

Tl;dr: any config value in a `tsconfig` that relates to a path inside of a package *has to* be inside of that package's `tsconfig.json` in order to have the correct relative path.

43081j:
ah yeah its me being blind.

i didn't see the `outDir`, it belongs in each individual tsconfig, my mistake.

43081j:
i had a play around w/ mocha setup.

you can drop ts-node and pass the built files just fine (`mocha dist/*.spec.js`) as long as we keep the sourcemap support package for nice stack traces.

IMO it'll be a much nicer experience as you're only then using typescript directly rather than ts-node (which introduces some limitations).

worth looking at in another branch probably

43081j:
ill catch this up some time this week and do the linting for the since-converted packages (interactive messages and events api)

43081j:
Just to update...

i went down a rabbit-hole of sorts. rebased onto master and started sorting the vast amount of lint errors interactive-messages produces.

i also made the mistake of thinking, "y'know what'd be useful while im in there, move the tests to typescript". but it turns out huge amounts of it really don't typecheck well (like stubbing private members all over the place).

so yeah ill probably finish it off eventually but who knows maybe it'll be too absurd of a diff to ever merge by then and we'll discard the tests anyway 🤷‍♂

43081j:
tests fail at the min because of this in the adapter unit tests:

```ts
        assert.throws(() => {
          adapter.action({ blockId: null }, actionHandler);
        }, TypeError);
        assert.throws(() => {
          adapter.action({ blockId: undefined }, actionHandler);
        }, TypeError);
```

i dont really understand how this ever passed... the `action` method's validation passes when `blockId` is falsy and has done for some time.

here:
https://github.com/slackapi/node-slack-sdk/blob/35b40af1dc28ef40cb00eb9a47f902e73c27b03f/packages/interactive-messages/src/adapter.ts#L41

and here:
https://github.com/slackapi/node-slack-sdk/blob/35b40af1dc28ef40cb00eb9a47f902e73c27b03f/packages/interactive-messages/src/adapter.ts#L250

in the second link `error` is false because the constraints check will have seen that `blockId` is falsy... meaning it won't throw.

but again thats been that way for a long time so i haven't a clue how the test ever passed before.

hilariously i didn't notice master changed again so will rebase again when i have the motivation again 🙊

packages/client/src/index.ts
```diff
@@ -1,3 +1,7 @@
+import { ErrorCode as WebAPIErrorCode } from '@slack/web-api';
```
aoberoi: i prefer having these import statements closer in the file to where the values are used. while that's not generally a preference across all source files, in the case of `index.ts`, it serves as symbol-indirection-map soo its useful.

aoberoi:
@43081j thanks for following through on all the changes up until now, wow.

i'd really like to see this PR land! i think the first thing we've got to do is resolve all the conflicts with `master` (and since #799 in particular). once this is done, it will be easier to conduct a meaningful code review.

a couple points that we might want to discuss:
 * let's use #830 as the place to decide how we proceed with the idea of moving `@types/*` dependencies into `"dependencies"` or `"devDependencies"`.
 * it seems as though you're still interested in moving tests from `*.js` files to `*.ts` files. if that's the case, let's start a new issue for that discussion.

43081j:
no worries, apologies for this going quiet too. looks like ive had a bit of an inactive month on github, good excuse to get back into it 👍 

ill catch up from master this week and re-read to get up to date on where i got to.

43081j:
@aoberoi i have yet to read through the diff again (as its a monster) but i've rebased onto master so should have all the new changes.

a few tests were moved to `.ts` so i had to manually diff by eye and move changes from the JS to the TS. so we should take extra care when reading those.

aoberoi:
> a few tests were moved to .ts so i had to manually diff by eye and move changes from the JS to the TS. so we should take extra care when reading those.

can we take the conversation outside of this PR? i don't see the idea of using ESLint over TSLint as one that should be coupled to moving the tests from `*.js` to `*.ts`. frankly, i'm not convinced the latter is worth doing at this point, but I'd rather have a separate discussion for that (outside this PR) so that others feel welcome to chime in.

43081j:
rebased & rolled back the typescript conversion of tests

filmaj:
Closing in favour of #1325.

Issue #690, opened 2019-02-08
Drop support for node versions before v8 LTS


aoberoi:
### Description

Adjust the test matrix so that we only test on versions we care to.

We should also update the documentation to state the supported versions.

Lastly, there may be some work to update type definitions to the appropriate versions.

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

PR #810, opened 2019-07-11
async/await in tests

43081j:
This cleans up tests a little on top of #808.

I noticed quite a few, if they failed, threw uncaught promises rather than ever reaching mocha so the errors weren't so useful.

Changing the async tests to use async/await instead of the `done` callback means uncaught promise exceptions will be handled by mocha. Makes it a bit cleaner to read too.

if you want to have a look at the diff, view the last commit by its self

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -86,15 +86,13 @@ describe('WebClient', function () {
           });
       });
 
-      it('should return results in a Promise', function () {
-        const r = this.client.apiCall('method');
-        return r.then(result => {
-          assert(result.ok);
-          this.scope.done();
-        });
+      it('should return results in a Promise', async function () {
+        const result = await this.client.apiCall('method');
+        assert(result.ok);
+        this.scope.done();
       });
 
-      it('should send warnings to logs', function() {
+      it('should send warnings to logs', async function() {
         const logger = {
           debug: sinon.spy(),
           info: sinon.spy(),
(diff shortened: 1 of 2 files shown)


aoberoi:
this is a great idea. i think we really should have done this work back in #690 when we dropped support for versions that couldn't run `async` functions.

since then, we've moved to the monorepo structure and pulled in the `@slack/interactive-messages` and `@slack/events-api` codebases (as i know you know). this complicates things. these two packages were never aligned with the rest in terms of node version support - they currently are meant to support `>= v4.2.0` (as shown in the `package.json` `"engines"` property). we would like to align the minimum node version support across the repo, but that means releasing a semver major change for those two packages. according to our published [Support Schedule](https://github.com/slackapi/node-slack-sdk/wiki/Support-Schedule) commitments, if we do that now, we still have to support the v1 and v2 branches of those packages until at least April 2020. however, the team feels like taking on the additional work of supporting two major versions isn't worth it just to get `async` functions into the tests.

so there's a middle road that i propose: can we move the tests for all the remaining packages to use `async`/`await` and leave the tests for those two packages alone?

PS. i recognize that we're living in a world where those packages are not being tested on CI against the node versions going back to v4.2.0. we should probably address that with #851.

43081j:
true. do both of these packages support async/await? (i.e. is the node version high enough?)

webhook and web-api

aoberoi:
oh yeah! both `@slack/webhook` and `@slack/web-api` have a minimum node version that is high enough to be able to use `async` functions 🎉

43081j:
rebased and dropped the typescript.

build doesn't seem to have started but tests did pass locally

packages/web-api/src/WebClient.spec.js
```diff
@@ -846,54 +802,55 @@ describe('WebClient', function () {
...
-        .catch((err) => {
-          assert.instanceOf(err, Error);
-          scope.done();
-          done();
-        });
+      try {
+        await client.apiCall('method');
+        assert.fail();
+      } catch (err) {
+        assert.instanceOf(err, Error);
```
aoberoi: similar to above, this version can suffer from false positives

packages/web-api/src/WebClient.spec.js
```diff
@@ -846,54 +802,55 @@ describe('WebClient', function () {
...
```
aoberoi: similar to above, this version can suffer from false positives

packages/webhook/src/IncomingWebhook.spec.js
```diff
@@ -46,62 +44,63 @@ describe('IncomingWebhook', function () {
...
-          agent.addRequest.restore();
-          agent.destroy();
-          throw error;
-        });
+      try {
+        await webhook.send('Hello');
+        assert.fail();
+      } catch (error) {
+        assert(spy.called);
+        agent.addRequest.restore();
```
aoberoi: the reason these lines are repeated in the original is to work similar to a `finally` block. when `assert(spy.called)` throws, these lines still need to run.

packages/web-api/src/WebClient.spec.js
```diff
@@ -158,79 +147,81 @@ describe('WebClient', function () {
           .reply(500);
       });
 
-      it('should return a Promise which rejects on error', function (done) {
-        const r = this.client.apiCall('method')
-        r.catch((error) => {
+      it('should return a Promise which rejects on error', async function () {
+        try {
+          await this.client.apiCall('method');
+          assert.fail();
```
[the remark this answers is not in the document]
43081j: thats a nice catch! maybe i can do a second assertion that its not an `AssertionError`

packages/webhook/src/IncomingWebhook.spec.js
43081j: ah! i was wondering what was going on here. we can just restore and destroy before our assertion so assertion exceptions propagate naturally

43081j:
@aoberoi so i've updated to assert for non-assertionerror errors.

it still doesn't seem ideal but its simpler than adding complexity or new dependencies to have an async `throws` kinda pattern/assertion. some tests assert for a specific type of error so i left those as-is, too.

ajarouih (approved):

Issue #830, opened 2019-07-17
Move non-exposed `@types/*` dependencies to "devDependencies"


clavin:
### Description

In #799, a [discussion](https://github.com/slackapi/node-slack-sdk/pull/799#discussion_r295082316) unfolded about the trade-offs of moving `@types/*` dependencies that don't contain exposed types to `"devDependencies"`. A [previous related dicussion](https://github.com/slackapi/node-slack-sdk/issues/514#issuecomment-375118480) can be found in #514.

##### In favor

- Developers who consume these packages don't care about or need `@types/debug` to consume `@slack/events-api`. This is just an example, but it points out the disconnect between these "dependencies" and each package's types.
- Minor bundle size difference.

##### Against

- Determining which `@types/*` goes where is a complex process that involves understanding what *is* and *isn't* exported by each package's API.
- It's relatively simple to make a small change that bubbles up to the API layer, exposing a type from a package that wasn't previously exposed. That is, it's too easy to get this wrong.
- A solution implies more code to maintain.

#### Proposed solution

The generated `.d.ts` files include triple-slash directives (`/// <reference ... />`) that tell the TypeScript compiler what `@types/*` packages are needed for that `.d.ts` file. The set of `@types/*` packages referenced by the `.d.ts` files for a given package could be reconciled with those listed in `"dependencies"` of that package's `package.json`. If any are missing or any extra are present, there's an error.

This could be implemented as an integration test or a post-build tool. The algorithm would look like:

1. For every package in the root directory `packages/`:
   1. Read `package.json` as `packageJson`
   1. Create a Set `deps` containing:
      1. Every dependency in `packageJson.dependencies` starting with `@types/`
         1. The `@types/` prefix should be *removed* when adding to `deps`, e.g. the dependency `@types/node` would be added as `node`
   1. Create a Set `refs`
   1. For every file ending in `.d.ts` in that package's `dist/`:
      1. Read the contents of that file as a string `fileContents`
      1. Split `fileContents` into `lines` (split by `\n`)
      1. Filter `lines` of `fileContents` to include only those starting with `///`
      1. Map each `line` to the match of that line on the regular expression `types="(.+)"`
      1. Filter `lines` to exclude non-matches
      1. For each remaining match `ref` (which is a string like `node` or `debug` or `express`):
         1. Add `ref` to `refs`
   1. If the [set difference](https://2ality.com/2015/01/es6-set-operations.html#difference) of `deps` without `refs` is not empty:
[...]

PR #831, opened 2019-07-19
Re-add reference documentation

clavin:
## Reference Documentation

Since the restructuring into a monorepo in April (marking the v5 release), `node-slack-sdk` has been missing reference documentation. This leaves some developers using the Node SDK packages lost when writing code that depends on these packages, aside from the cookie-cutter tasks documented by the tutorials for each. This PR addresses this concern by reimplementing a reference documentation generation pipeline.

### Pipeline

The pipeline is composed of a few different parts:

- Type and documentation information is extracted into models by [API Extractor](https://api-extractor.com/), which is fed into...
- A custom script (`support/ref-docs/`) that, for each package's API model, generates content about the package, represented as...
- [`mdast`](https://github.com/syntax-tree/mdast) (markdown AST), which is transformed into a markdown document by...
- [Remark](https://github.com/remarkjs/remark), a mature project centered around markdown parsing and compiling

Ref Docs Pipeline Diagram

<details>
<summary>ASCII Diagram</summary>

```
       `packages/`
+-----------+-----------+
|           |           |                           `support/ref-docs/models/`
| `web-api` |    ...    |                            +----------------------+
|           |           |                            | `web-api.model.json` |
+-----------+-----------+   === API Extractor ===>   | `rtm-api.model.json` |
|           |           |                            |          ...         |
|    ...    | `rtm-api` |                            |          ...         |
|           |           |                            +----------------------+
+-----------+-----------+                                        |
                                                                 |
                                                                 |
           +------------------- Custom Script -------------------+
           |
           |
           v
      `mdast` AST                                      Documentation         
+ - - - - - - - - - - +                     +--------------------------------+
\  root               /                     | # Heading                      |
/  \-- heading        \                     |                                |
\   |- paragraph      /                     | Some absolutely awesome con... |
/   |- table          \                     |                                |
\   | \-- table row   /   === Remark ===>   | |  Name | Type of Unicorn |    |
/   |  |- table row   \                     | | :---: | --------------- |    |
\   |  |- ...         /                     | | Ralph | Rainbow         |    |
/   |- heading        \                     | |  Alex | Possibly a dog  |    |
\   |- ...            /                     | |  ...  | ...             |    |
+ - - - - - - - - - - +                     +--------------------------------+
                                                             |
                                                             |
                                                        [Written to]
                                                             |
                                                             v
                                                    `docs/_reference/`
                                                     +--------------+
                                                     | `web-api.md` |
                                                     | `rtm-api.md` |
                                                     |      ...     |
                                                     +--------------+
```

</details>

Most of the work is done by other tooling that does the heavy lifting. The only custom code is generating the content itself.

### To-Do

- [x] Anything that says "TODO"
- [ ] Handling inline interfaces
- [x] Content for type aliases
- [x] Manual handling to make sure `@slack/web-api` can export & render w/o a million methods
- [ ] Documenting *how* to generate reference docs


packages/interactive-messages/package.json
```diff
@@ -63,6 +64,10 @@
     "tsscmp": "^1.0.6"
   },
   "devDependencies": {
+    "@microsoft/api-extractor": "^7.3.4",
+    "@types/debug": "^4.1.4",
```
aoberoi: not sure why this package got new devDependencies besides `@microsoft/api-extractor`, when most of the other packages did not.

aoberoi: also, we haven't changed convention to moving `@types/*` dependencies into `devDependencies` (discussion in #830)

packages/events-api/package.json
```diff
@@ -55,6 +56,10 @@
     "yargs": "^6.6.0"
   },
   "devDependencies": {
+    "@microsoft/api-extractor": "^7.3.4",
+    "@types/debug": "^4.1.4",
```
aoberoi: [ditto](https://github.com/slackapi/node-slack-sdk/pull/831/files#r309500528)

.github/maintainers_guide.md
```diff
@@ -22,9 +22,9 @@ and to allow lerna to link them to each other.
 ### Testing
 
 This project has tests for individual packages inside of each's respective `test` directory. It also has
-integration tests in the `integration-tests` directory at the root. You can run the entire test suite using the npm
-script `npm test` at the top level. This script will use Lerna to invoke tests in each package and the integration
-tests.
+integration tests in the `integration-tests` directory under the root `support` directory. You can run the entire test
```
aoberoi: i'd like to understand the motivation behind moving the `integration-tests` into the `support` directory.

from my point of view, the `support` directory communicates that the contents are there to support the repository itself. so it contains tools and scripts and assets needed to support the repo. i don't think integration tests fit that description, they are meant to be updated with the codebase as it changes. in this view, they are a part of the product rather than support.

support/ref-docs/src/index.ts
```diff
@@ -0,0 +1,68 @@
+/// <reference types="./types/remark-toc" />
```
aoberoi: 🤔 in other packages, we've chosen to keep the `types` directory for locally defined types (as opposed to those installed from `@types/*` packages) at the root of the package instead of inside the `src` directory. is that something you think might be worth changing?

also, we've used the `tsconfig.json` to configure the compiler to look in that directory for packages by name, instead of using `/// <reference .../>` directives. do you feel like this is something we should change?

packages/interactive-messages/package.json
clavin: 🤔 Good question. I think this is the result of a bad merge commit (read: my fault while merging).

.github/maintainers_guide.md
clavin: > the `support` directory communicates that the contents are there to support the repository itself.

I think integration tests *do* fit that description. It might just be my perspective: tests are a tool that help ensure that the codebase--the meat of the repo--are working as expected. I'd even go as far as saying it feels *odd* to me if `integration-tests/` were top-level alongside `support/` (as opposed to inside `support/`).

I think the notion that tools that support the repo remain constant while the codebase changes is flawed. The tools that support the repo also will need to grow as the codebase changes, albeit at different rates.

aoberoi: thanks for clarifying and supporting your case! i'm comfortable with your definitions now. let's keep it as you proposed 👍

support/ref-docs/src/arrays.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Joins (or concatenates) all arrays passed to this function into one
+ * continuous array.
+ */
+export function joinArrs<T>(...arrs: readonly (readonly T[])[]): T[] {
```
aoberoi: the [`Array.prototype.flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) method might be a good substitute for this helper. for example, `joinArrs([a, b], [c, d, e])` is the same as `[[a, b], [c, d, e]].flat()`.

it's only available in Node 11.0 and above, but then i think including a polyfill with the same API is better than using this helper. see: https://www.npmjs.com/package/array.prototype.flat

support/ref-docs/src/arrays.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * An implementation of the standardized `flatMap`.
+ *
+ * See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap
+ *
+ * @remarks
+ * Adapted from https://2ality.com/2017/04/flatmap.html
+ */
+export function flatMap<T, U>(
```
aoberoi: do you think it makes sense to instead depend on [`array.prototype.flatmap`](https://www.npmjs.com/package/array.prototype.flatmap), which is a polyfill for this API? the advantage is that the dependency can simply be removed the next time we update the minimum node requirement based on the support schedule, rather than having to delete the code itself. also, we'd get bug fixes (if there were any) and other updates for free.

support/ref-docs/src/documenter.ts
```diff
@@ -0,0 +1,624 @@
...
+ */
+function tryLinkName(
+  ctx: DocumentationContext,
+  name: string,
+  child: Node
+): Node {
+  // Attempt to find the target item
+  const [target] = ctx.pkgEntry.findMembersByName(name);
+  // TODO: handle multiple targets?
+  // TODO: better extraction by using tokens instead of
```
aoberoi: using tokens instead of... name?

support/ref-docs/src/documenter.ts
```diff
@@ -0,0 +1,624 @@
...
+
+  // If the target isn't found, simply return the child w/o a link
+  if (target === undefined) {
+    return child;
```
aoberoi: does this mean that when the target isn't found, the node will link to itself? i'm wondering if returning `undefined` would keep things simpler.

support/ref-docs/src/documenter.ts
```diff
@@ -0,0 +1,624 @@
...
+
+/**
+ * Formats certain types to be more human-friendly.
+ */
+function formatExcerptReferences(
+  ctx: DocumentationContext,
+  excerpt: Excerpt
+): Node[] {
+  // Bail early for object types/interface shorthand
+  if (excerpt.text.match(/^{(?:\s|\S)*}(?:\[\])*$/) !== null) {
```
aoberoi: can we get a comment to describe what the RegExp is looking for? maybe with an example of what should (or should not) match?

support/ref-docs/src/documenter.ts
```diff
@@ -0,0 +1,624 @@
...
+    // TODO: better substitution that explains properties
```
aoberoi: can you help me understand what this comment means?

support/ref-docs/src/documenter.ts
```diff
@@ -0,0 +1,624 @@
...
+        .filter(node => !(node instanceof DocSoftBreak))
+        // Convert DocNodes to mdast nodes
+        .map(node => {
+          if (node instanceof DocPlainText) {
+            // Plain text maps to `text` nodes
+            return text(node.text);
+          } else if (node instanceof DocLinkTag) {
+            // Links map to links. The URI depends on the destination.
+            if (node.codeDestination !== undefined) {
+              // TODO: better code link resolution
```
aoberoi: what would you do to improve it?

support/ref-docs/src/documenter.ts
```diff
@@ -0,0 +1,624 @@
...
+            } else if (node.urlDestination !== undefined) {
+              return link(node.urlDestination, undefined, [
+                text(
+                  node.linkText === undefined
+                    ? node.urlDestination
+                    : node.linkText
+                )
+              ]);
+            } else {
+              throw new Error('misformatted DocLinkTag');
```
aoberoi: just so i understand, if you throw here, it will bubble up and terminate the documentation generation process, right?

support/ref-docs/src/documenter.ts
```diff
@@ -0,0 +1,624 @@
...
+    return [];
+  }
+
+  // Attempt to find the `@example` block
+  const exampleBlock = item.tsdocComment.customBlocks.find(
+    block => block.blockTag.tagNameWithUpperCase === '@EXAMPLE'
+  );
+
+  return joinArrs(
+    // TODO: Deprecation notice
```
aoberoi: hmm? does this mean outputting something for methods with `@deprecated` tags?

support/ref-docs/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+import { promisify } from 'util';
+import { ApiModel } from '@microsoft/api-extractor-model';
+import remarkStringify = require('remark-stringify');
+import remarkToc = require('remark-toc');
+import unified = require('unified');
+import { VFile } from 'vfile';
+import documenter from './unist-attacher';
+
+/** Directory that contains API models. */
+const API_MODELS_PATH = joinPaths(__dirname, '..', 'models');
```
aoberoi: one day... maybe we take paths like this and `OUTPUT_PATH` below as CLI parameters. for today, this is great!

support/ref-docs/src/arrays.ts
```diff
@@ -0,0 +1,41 @@
...
+    result.push(...mapper(elem, index, arr));
+  }
+
+  return result;
+}
+
+/**
+ * Filters out any elements from an array that are `undefined`.
+ */
+export function filterUndef<T>(arr: T[]): Exclude<T, undefined>[] {
```
aoberoi: if the rest of this file goes away, then this might actually belong in `macros.ts`.

support/ref-docs/src/arrays.ts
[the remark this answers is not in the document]
clavin: `.flat` was added to Node.js in v11 (as MDN's compatibility table claims). Our supported target versions go all the way down to v8 right now. I wanted to be able to support building docs on our entire supported range.

support/ref-docs/src/documenter.ts
clavin: I don't *quite* know of what situation would cause this, but it means that it's not known how to format this link. So yes, it would terminate the doc generation process because it bubbles up.

support/ref-docs/src/documenter.ts
clavin: If it can't find a target API item, it returns the node (text, inline code) that is displayed inside of the `<a>` tag/`[]` markdown brackets.

That means if you get `tryLinkName(..., 'thisNameDoesntExist', text('thisNameDoesntExit'))` (the last element being an AST `text` node) then it'd just return the `text` node instead of a `link` node with the `text` node as its child.

So it doesn't link to anything because it won't be a link.

stevengill:
Hey @clavin!

I'm new on the Slack DevRel tools team. @aoberoi asked me to take a look at this PR and see if I can get it over the finish line. 

After checking out the code I think I got a pretty good understanding of how it is all working.

Wondering if you want to add me as a collaborator on your fork of this repo so I can continue pushing to your `ref-docs` branch and updating this PR. Alternatively, I could open a new PR based on this work but we would lose the context from the conversation already had on this PR.

Your remaining TODOs:

- [ ] Handling inline interfaces (e.g. see RTMCallResult in the example ☝️)
> How do you envision this being handled? What would the expected output look like? I'm not sure I understand how it isn't being handled currently?

- [ ] Manual handling to make sure @slack/web-api can export & render w/o a million methods
> Currently I'm running into https://github.com/microsoft/rushstack/issues/1029 when trying to run `api-extractor` on `web-api`. I'll have to dig deeper on this.

- [ ] Documenting how to generate reference docs
> Should be pretty straight forward. 

Let me know if it would be easier to have a quick call/video chat about this PR instead of responding on here.

clavin:
Hi! 👋👋👋 Glad to see a new face 😄

## Handling inline interfaces

This refers to when the type of something is an inline interface, e.g. the `rectangle` field in:

```typescript
interface Shapes {
  circle: number;
  rectangle: {
    width: number;
    height: number;
  };
  square: number;
}
```

I was pondering how the fields of `rectangle` should end up rendering, especially if `recangle` was an array and/or multiple layers deep.

My thought was to expand the field in place in the table like this:

| Name | Type |
|------|------|
| circle | `number` |
| rectangle | `object` |
| rectangle.width | `number` |
| rectangle.height | `number` |
| square | `number` |

I don't know how this solution generalizes to arrays though 🤔... maybe that edge case can be ignored for now? Maybe just tack an `[i]` to the end to symbolize its an array?

---

As for the other two TODOs, you seem to be in the same spot as me 😄. Ask away if you run into anything! 🚀

(I'm definitely happy to do either! If you peek at [my other comments on this repo](https://github.com/slackapi/node-slack-sdk/issues/830), you'll see I'm definitely no stranger to giving long responses 😉)

stevengill:
Thanks for the detailed response @clavin!

For inline interfaces, I can definitely see us hitting interesting edge cases. For instance, what if an inline interface contains an inline interface. Or an inline interface contains an array of inline interfaces. It can go pretty deep. I like the idea of going one level with inline interfaces to start. So add support for `rectangle.width` and `rectangle.height` like your example. I'll think about how to show array (the `[i]` syntax sounds fine to me)

For `Manual handling to make sure @slack/web-api can export & render w/o a million methods`, I've got a prototype working where I make `methods.ts` its own npm package. This allows `import Method, * as methods from '@slack/methods';` to work with `api-extractor`. Seems like a clean way to handle it but I'll have to see how @aoberoi feels about it. Better than adding a massive import to the top of `WebClient.ts`

clavin:
@stevengill If you don't mind, I have an idea for how the nested situation can be solved 💡

Imagine that, currently, the table that represents the fields of an interface looks like a simple array:

```json
[
  "circle",
  "rectangle",
  "square",
]
```

(Lets ignore types for a moment.)

Now, we have some procedure to convert those field names into the markdown AST:

```typescript
function fieldRows(fields: string[]): mdast.TableRow[] {
  return fields.map(field => tableRowWithText(field));
}
```

The result currently is a one-to-one mapping, where one field in the input array corresponds exactly to one field in the result, no more and no less. The current result would be something like:

```js
[
  { type: "tr", text: "circle" },
  { type: "tr", text: "rectangle" },
  { type: "tr", text: "square" },
]
```

Now let's add types:

```json
[
  ["circle", "number"],
  ["rectangle", [
    ["width", "number"],
    ["height", "number"],
  ]],
  ["square", "number"],
]
```

And a corresponding TypeScript type for this model (assuming support for recursive types):

```typescript
type Field =
  | [string, string]   // A regular field
  | [string, Field[]]; // An inline interface
```

With this information, we could just make this `fieldRows` function recursive based on type of each field! The easiest way is to use `flatMap`:

```typescript
function fieldRows(fields: Field[], prefix: string): mdast.TableRow[] {
  fields.flatMap(([name, type]) =>
    Array.isArray(type) // (this checks to see if the type is an inline interface)
      ? fieldRows(type, name + '.')
      : [tableRowWithText(`${prefix}${name}: ${type}`)]
  );
}
```

This gives us a really neat result!

```js
[
  { type: "tr", text: "circle: number" },
  { type: "tr", text: "rectangle.width: number" },
  { type: "tr", text: "rectangle.height: number" },
  { type: "tr", text: "square: number" },
]
```

An extra detail here is that we should probably include the name of the inline interface as its own row (that is there should be a row that says "rectangle: object"). That can be done by amending the branch in `fieldRows` that handles inline interfaces to prepend a row to the result of the recursive `fieldRows` call (e.g. `? [rowForThisField(), ...fieldRows(...)]`).

stevengill:
Hey @clavin, when you get a chance, can you review my PR of this at https://github.com/slackapi/node-slack-sdk/pull/933. I'd like to get it merged 😄 

For the nested interfaces, I'm going to file an issue for it so we can tackle it after this is merged and released.

clavin:
I've made a lot of changes! Much sweeping, much dusting, much polishing. Code is great, but here's some screenshots of what this reference *actually* looks like.

RTM API Reference Documentation
Web API Reference Documentation
Reference Documentation Home

The reference documentation is accessed on the sidebar of the documentation home:

Screen Shot 2020-06-24 at 1 46 34 PM

stevengill:
@clavin the third screenshot is missing OAuth, but I do see it in the sidebar in the last one

stevengill:
@clavin I just did a run through of the changes. Looks solid! I pulled it down locally and ran it with no issues. 

This doesn't currently include the @slack/oauth package. It would be nice if we could add reference docs for that too, but i think it is fine if that is a future PR.

clavin:
I hooked up the bits that let ref docs generate info about `@slack/oauth` and it Just Worked™ 😎

---

@aoberoi I added a lot of exports to packages that were somehow exposed in the already exported API--particularly the `CodedError` interface in probably every single package that has one. Is there a reason this interface and others aren't already being exported?

PR #835, opened 2019-07-22
Add method name to apiCall debug

clavin:
###  Summary

Adds the name of the method being called by `apiCall` to the debug log message at the start of the method.

### Requirements


aoberoi:
Not a huge deal (i've already approved) but if this is functionality we don't want to regress, then it would be nice to get some tests that verify it.

clavin:
I don't really know if the contents of a debug message is part of what we should be testing 🤔 Like, I don't know if this is really something developers should *depend* on so much as it is just useful information to figure out where something went wrong. Yay or nay?

PR #839, opened 2019-07-26
Adds support for the Remote Files API

aoberoi:
###  Summary

Adds methods and types related to Slack's new Remote Files API. Learn more: https://api.slack.com/messaging/files/remote

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -145,7 +145,7 @@ export interface Datepicker extends Action {
  * Block Types
  */
 
-export type KnownBlock = ImageBlock | ContextBlock | ActionsBlock | DividerBlock | SectionBlock;
+export type KnownBlock = ImageBlock | ContextBlock | ActionsBlock | DividerBlock | SectionBlock | FileBlock;
 
 export interface Block {
   type: string;
@@ -178,6 +178,12 @@ export interface SectionBlock extends Block {
   text?: PlainTextElement | MrkdwnElement; // either this or fields must be defined
   fields?: (PlainTextElement | MrkdwnElement)[]; // either this or text must be defined
   accessory?: KnownAction | Action | ImageElement;
+}
+
+export interface FileBlock extends Block {
+  type: 'file';
+  source: string; // 'remote'
+  external_id: string;
 }
 
 export interface MessageAttachment {
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -365,10 +365,13 @@ export interface FilesCommentsDeleteArguments extends WebAPICallOptions, TokenOv
 }
 
 // either file or external_id is required
-interface FilesRemoteInfoCommonArguments extends WebAPICallOptions, TokenOverridable { channel?: string; }
-type FilesRemoteInfoArgumentsFileRequired = FilesRemoteInfoCommonArguments & { file: string };
-type FilesRemoteInfoArgumentsExternalIdRequired = FilesRemoteInfoCommonArguments & { external_id: string };
-export type FilesRemoteInfoArguments = FilesRemoteInfoArgumentsFileRequired | FilesRemoteInfoArgumentsExternalIdRequired;
+export interface FilesRemoteInfoArguments extends WebAPICallOptions, TokenOverridable {
+  channel?: string;
+
+  // either one of the file or external_id arguments are required
+  file: string;
+  external_id: string;
+}
 
 export interface FilesRemoteListArguments extends WebAPICallOptions, TokenOverridable {
   ts_from?: string;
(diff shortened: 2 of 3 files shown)


packages/types/src/index.ts
```diff
@@ -180,6 +180,12 @@ export interface SectionBlock extends Block {
   accessory?: KnownAction | Action | ImageElement;
 }
 
+export interface FileBlock extends Block {
+  type: 'file';
+  source: string; // 'remote'
```
aoberoi: what string values can we expect other than `'remote'`?

packages/web-api/src/methods.ts
```diff
@@ -364,6 +364,55 @@ export interface FilesCommentsDeleteArguments extends WebAPICallOptions, TokenOv
   id: string; // comment id
 }
 
+// either file or external_id is required
+interface FilesRemoteInfoCommonArguments extends WebAPICallOptions, TokenOverridable { channel?: string; }
+type FilesRemoteInfoArgumentsFileRequired = FilesRemoteInfoCommonArguments & { file: string };
+type FilesRemoteInfoArgumentsExternalIdRequired = FilesRemoteInfoCommonArguments & { external_id: string };
+export type FilesRemoteInfoArguments = FilesRemoteInfoArgumentsFileRequired | FilesRemoteInfoArgumentsExternalIdRequired;
```
aoberoi: i understand the goal of this technique is to enforce that values fit either of the two shapes that are required. however this pattern isn't currently used anywhere else in this project, so it's worth looking at whether we want to do this or not. in other places, we just use comments to document which fields are dependent on which other fields (you can take a look at the relationship of `FilesUploadArguments` `file` and `content` properties).

one reason we haven't done this in other places where it could have been done is for simplicity for a user to read and understand the types. adding this complexity comes at a cost to the reader.

another reason is that we have a long term goal of being able to generate this file from some common metadata. techniques like this make it much more complex, maybe even impossible, for these types to come from a script to generate the types.

once you add types, you may not be able to take them away (even those that aren't exported, because they are accessible via some of the types that are exported). its based on these reasons that i recommend flattening out this type definition to just mark the properties as optional even when one of a set of them might be required, and use comments to help the user understand. we can come back and do this later, and apply whatever we decide makes sense to all the types where it makes sense at once (e.g. `FilesUploadArguments`).

packages/web-api/src/methods.ts
```diff
@@ -364,6 +364,55 @@ export interface FilesCommentsDeleteArguments extends WebAPICallOptions, TokenOv
...
+  title: string;
+  external_url: string;
+  filetype: string; // possible values (except for 'auto'): https://api.slack.com/types/file#file_types
+  preview_image?: Buffer | Stream;
+  preview_file?: Buffer | Stream;
+  indexable_file_contents?: Buffer| Stream;
+}
+type FilesRemoteUpdateArgumentsFileRequired = FilesRemoteUpdateCommonArguments & { file: string };
+type FilesRemoteUpdateArgumentsExternalIdRequired = FilesRemoteUpdateCommonArguments & { external_id: string };
+export type FilesRemoteUpdateArguments = FilesRemoteUpdateArgumentsFileRequired | FilesRemoteUpdateArgumentsExternalIdRequired;
```
aoberoi: same comment from above.

packages/web-api/src/methods.ts
```diff
@@ -364,6 +364,55 @@ export interface FilesCommentsDeleteArguments extends WebAPICallOptions, TokenOv
...
+
+interface FilesRemoteRemoveCommonArguments extends WebAPICallOptions, TokenOverridable {}
+type FilesRemoteRemoveArgumentsFileRequired = FilesRemoteRemoveCommonArguments & { file: string };
+type FilesRemoteRemoveArgumentsExternalIdRequired = FilesRemoteRemoveCommonArguments & { external_id: string };
+export type FilesRemoteRemoveArguments = FilesRemoteRemoveArgumentsFileRequired | FilesRemoteRemoveArgumentsExternalIdRequired;
```
aoberoi: same comment from above.

packages/web-api/src/methods.ts
```diff
@@ -364,6 +364,55 @@ export interface FilesCommentsDeleteArguments extends WebAPICallOptions, TokenOv
...
+
+interface FilesRemoteShareCommonArguments extends WebAPICallOptions, TokenOverridable {
+  channels: string; // comma-separated list of channel ids
+}
+type FilesRemoteShareArgumentsFileRequired = FilesRemoteShareCommonArguments & { file: string };
+type FilesRemoteShareArgumentsExternalIdRequired = FilesRemoteShareCommonArguments & { external_id: string };
+export type FilesRemoteShareArguments = FilesRemoteShareArgumentsFileRequired | FilesRemoteShareArgumentsExternalIdRequired;
```
aoberoi: same comment from above.

aoberoi:
@seratch I think the test we currently have (`when API arguments contain binary to upload`) covers this usage well! I just wanted to know if you've ever tried this code against the actual platform to verify that it works.

packages/web-api/src/methods.ts
seratch: Thanks for your comment. In particular, 
>another reason is that we have a long term goal of being able to generate this file from some common metadata. techniques like this make it much more complex, maybe even impossible, for these types to come from a script to generate the types.

This part really makes sense to me. I will make these changes simpler considering your points.

seratch:
>I just wanted to know if you've ever tried this code against the actual platform to verify that it works.

I already made sure the code (honestly not all of them) works with actual APIs when I wrote them. But I'm going to change the code regarding the above point (avoiding tricky type definitions). So I'll verify the updated code works for sure.

aoberoi:
@seratch if you haven't already started to change this, don't worry about it. i can make the change real quick. that will allow us to make a release to support remote files before the weekend starts here 😄 🙇

packages/web-api/src/methods.ts
```diff
@@ -364,6 +364,61 @@ export interface FilesCommentsDeleteArguments extends WebAPICallOptions, TokenOv
...
+// either file or external_id is required
+export interface FilesRemoteInfoArguments extends WebAPICallOptions, TokenOverridable {
+  channel?: string;
+
+  // either one of the file or external_id arguments are required
+  file: string;
+  external_id: string;
+}
+
+export interface FilesRemoteListArguments extends WebAPICallOptions, TokenOverridable {
```
clavin: nit: this can be `TraditionalPagingEnabled`, removing `count` and `channel` from this interface.

PR #850, opened 2019-08-20
Fix webhook httperror statuscode

aoberoi:
###  Summary

When using the `IncomingWebhook#send()` method, if an HTTP error occurs (a response with a non-200 status code), the error's message should contain the status code. This was broken and was returning `undefined`. It was because the `error.code` property was not the same thing as `error.response.status` as previously assumed.

This PR addresses this and adds a test to verify this behavior. It also solves a problem where the test was not returning the Promise, so it likely wasn't verifying the error behavior at all before.


aoberoi:
Codecov status check ignored because the coverage report is incorrect until we can get staged builds.

PR #853, opened 2019-08-21
Make the text field in ChatUpdateArguments optional

DominikPalo:
###  Summary
The `text` argument in the `chat.update` method [is not required](https://api.slack.com/methods/chat.update#arguments) when presenting `attachments`, so I've changed it to optional.

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -216,7 +216,7 @@ export interface ChatUnfurlArguments extends WebAPICallOptions, TokenOverridable
 }
 export interface ChatUpdateArguments extends WebAPICallOptions, TokenOverridable {
   channel: string;
-  text: string;
+  text?: string;
   ts: string;
   as_user?: boolean;
   attachments?: MessageAttachment[];


aoberoi:
Hey @DominikPalo, you raised a very interesting point. I hadn't noticed the contradiction before.

I have a few reasons that I'd prefer to keep `text` required:

1. Messages that don't have a `text` property render poorly in notifications. For example, if the user receives a push notification on their mobile device for a message, Slack uses the `text` string to describe the message in the banner/alert. Slack cannot render attachments (or Block Kit) within an iOS or Android push notification. There are other places where Slack uses `text` as the alternate rendering of a message. Therefore, it is a best practice to include `text` on all messages.

2. At some point in the future, we'd like to be able to programmatically generate all the types in `methods.ts`. The closer we align to the [API Spec](https://github.com/slackapi/slack-api-specs/blob/master/web-api/slack_web_openapi_v2.json), which lists `text` as required, the less compatibility issues we will have as we transition from hand-rolled types to generated ones. I believe it's marked required in the specification intentionally, in order to further the best practice. This is also the reason why we haven't decomposed any of the `Arguments` types into unions of more specific acceptable shapes - instead we type many properties as optional (even though one of them might be required without another).

I understand this presents a problem for developers who are specifically choosing not to send `text` and understand the consequences. After all, the API allows it, so why should the SDK make that decision for me? I believe most (like over 99%) developers would benefit from leaving it required, and for the ~1% who see this as a problem, there is [a reasonable workaround](http://www.typescriptlang.org/play/#code/PTAEEsDsBcFMCcBmBDAxrUBbA9gZ2gDYCeoq2ADuLACaiLzaagACuBaA1sAO6wBGAWmSUAUFDhI0GAMIALZNACq5agtgBBeAHMArplgxcoAN4jQpeZEiwCALlD54ULQG4zoOAA9o9x87fm0Li+0E6Qru7IuAD6OrgIAPz2fNjYBLDIkAGgCtBosvqGSTmQRADaALrZfATYqBy4xZnlVe4EUBzRkMj6jcmp6ZnZ5Mjw8cUA5Ig6BAQToAA+oBOQ2NYTbgC+IiIgOXQ6kKjQ4Gse8tBY4JjgqEYA6vzS7QbQAHSoF286KmoAFABKETUWCodjwDDTI4nM6fBTKVRwP6jLTBUByeG-OCaXSFIIA+wAN2w4Gobl2YFQyFmznOCgO0NOkFAcVpq1AAAMvNAOaA+CQqfhadBsOcMMJyAxJeA1B4iORYCI4UosbA-qZzOY4VYbPYJtIAIwAJgAzAAWACsEwANO5AmiJsbzRaAGwAdgAHABON4ABn9AZtdpy0DynzxaMqts2OSMzVj6IuCLUOL0r1wQJEQA).

Do you feel otherwise? If so, how strongly do you feel about it? I admit I don't feel strong on this stance, but I'd like to know if you have reasons I didn't consider or don't yet understand.

seratch:
@aoberoi @stevengill as we decided to make `text` optional in Bolt or JS, there is no reason to hold off merging this PR. https://github.com/slackapi/bolt-js/pull/512

aoberoi:
@seratch @stevengill The update in Bolt for JS makes `text` optional for using a `response_url`, not for `chat.update` (or any Web API method). In my previous comment I explained why I don't think we should make it optional here, and I asked for any further reasons why we should. Have we heard any new reasons?

seratch:
@aoberoi 
Sorry, I've overlooked the context of the change. https://github.com/slackapi/bolt-js/issues/416

In summary, you wanted to make the `text` field optional only for the cases sending `delete_orignal: true` to `response_url` endpoint. But due to some technical limitations, we were not able to come up with such a solution. Then, we've decided to make the field optional for all cases. So, you still believe we should not loosen the restrictions to others such as `chat.postMessage`. 

When it comes to this context, we don't have any new information. Also, I totally agree that omitting the `text` field may result in a non-optimal user experience with search results, mobile notifications, experience with older client apps, and so on. So, even when a developer utilizes `blocks` or `attachments` for composing a message, having `text` as well is highly recommended.

But, on the other hand, I think we may be able to improve the developer experience by thinking about the following points.

* :one: In the first place, the server-side doesn't have any validation on the absence of `text` field as of today. Also, other 1st party SDKs don't have a similar validation. Thus, the limitation is unique to node-slack-sdk only when using it in TypeScript. The situation sounds a minor inconsistency issue to me.
* :two: The restriction is effective only for the apps written in TypeScript. JavaScript users can bypass the restriction or they're not even aware of it.

This is just my two cents but we may consider:
* for :one:, we may be able to provide the server-side to return a warning when an API call omits `text` field with the reason for it.
* for :two:, we may be able to have a runtime validation for JavaScript users. Also, we may consider having similar in other SDKs but I'm not positive to do so.
* Needless to say, yet another alternative for :one: & :two: is changing `text` to be optional as I suggested here.

I wanted to mention that it looks inconsistent and I personally don't see big benefits to have such difference from the server-side and other SDKs. That said, I'm not having any strong opinions on this and it may be a matter of preference. I'm happy to respect @aoberoi 's design decision on this matter.

aoberoi:
@seratch great points! Sorry it's taken me so long to return to this conversation.

Regarding the options:

1. Emitting warnings from the platform (which would trigger a `LogLevel.WARN` log from `WebClient`): I like this idea a lot. It might take a while to ship (or possibly won't ship at all) because we need to work with other team(s) to make it happen. Let's try!

2. Validation of arguments before making a call to `chat.postMessage` (eager warning): I also like this idea, but I would want to remove this if (1) lands. If we do this, we should do it in all SDKs.

3. Changing `text` property in type definition to be optional: All this accomplishes is allow TypeScript developers to more easily break best practices. There's two cases.
    A) If you're determined to do what we think isn't good for users, you have plenty of choices on how to accomplish this. For example, you could use `// @ts-ignore`, you could use the `client.apiCall('chat.update', ...)` method, you could use cast such as `client.chat.update({ ... } as any as ChatUpdateArguments)`. I believe the friction this audience might experience in producing one of these options is good because it makes them reconsider how important it is to them.
    B) Otherwise, when you're not intentionally trying to omit `text` for a well considered reason, it's much more beneficial to see the error early (before the runtime warning).

    Since both groups of people net benefit, I'd advocate for leaving this as is, even if we accomplish (1) and/or (2).

seratch:
@aoberoi Thanks for sharing the thorough thoughts!

1. yes, let's try it :)
2. I'm positive to have such warnings on the SDK side. Regarding the execution, I would like other maintainers to own it as I have other priorities such as Bolt for Python.
3. if we move 2. forward quickly, we don't need to change the type definitions

aoberoi:
I've created an issue to capture (2) for each SDK (Node, Python, and Java). I'll close this PR now since all remaining steps are captured there.

PR #870, opened 2019-09-25
Add Block Kit in modals support

shanedewael:
###  Summary

Add support for Block Kit within modals
https://api.slack.com/block-kit/surfaces/modals

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -89,9 +89,9 @@ export interface Confirm {
  */
 
  // Selects and Multiselects are available in different surface areas so I've seperated them here
-export type KnownSelect = UsersSelect | StaticSelect | ConversationsSelect | ChannelsSelect | ExternalSelect;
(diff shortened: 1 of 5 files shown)


packages/types/src/index.ts
```diff
@@ -77,19 +88,30 @@ export interface Confirm {
  * Action Types
  */
 
-export type KnownAction = UsersSelect | StaticSelect | ConversationsSelect | ChannelsSelect | ExternalSelect |
-  Button | Overflow | Datepicker;
+ // Selects and Multiselects are available in different surface areas so I've seperated them here
+export type KnownSelect = UsersSelect | StaticSelect | ConversationsSelect | ChannelsSelect | ExternalSelect;
```
aoberoi: It doesn't seem like there's any reason to prefix these two types with `Known`. The type no longer gets used in unions with `Action`, so all the action types are "known" unless they are `Action`.

PR #888, opened 2019-10-09
Publish

shanedewael:
###  Summary
#887

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -45,7 +45,7 @@
     "test": "npm run build && echo \"Tests are not implemented.\" && exit 0"
   },
   "dependencies": {
-    "@slack/logger": "^2.0.0",
+    "@slack/logger": ">=1.0.0 <3.0.0",
     "@slack/rtm-api": "^5.0.3",
     "@slack/types": "^1.2.1",
     "@slack/web-api": "^5.2.1",
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -45,11 +45,11 @@
     "test": "npm run build && echo \"Tests are not implemented.\" && exit 0"
   },
   "dependencies": {
-    "@slack/logger": "^1.0.0",
-    "@slack/rtm-api": "^5.0.2",
-    "@slack/types": "^1.1.0",
-    "@slack/web-api": "^5.1.0",
-    "@slack/webhook": "^5.0.1"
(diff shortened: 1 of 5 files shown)


packages/client/package.json
```diff
@@ -45,11 +45,11 @@
     "test": "npm run build && echo \"Tests are not implemented.\" && exit 0"
   },
   "dependencies": {
-    "@slack/logger": "^1.0.0",
-    "@slack/rtm-api": "^5.0.2",
-    "@slack/types": "^1.1.0",
-    "@slack/web-api": "^5.1.0",
-    "@slack/webhook": "^5.0.1"
+    "@slack/logger": "^2.0.0",
```
aoberoi:
```suggestion
    "@slack/logger": ">=1.0.0 <3.0.0",
```

Since the `@slack/client` package doesn't use `getLevel()`, it doesn't actually require a `^2.0.0` version of `@slack/logger`. We should aim to describe the maximum range of compatibility, so that we don't unintentionally make users download and install this package more than necessary. How would that happen? If a package or app uses that uses `@slack/client` also decided to use `@slack/logger` and has `^1.0.0` as its version range, then this change would force them to download a second copy of the package. The suggestion allows npm to deduplicate into a single `1.x` version.

packages/rtm-api/package.json
```diff
@@ -41,8 +41,8 @@
     "test": "npm run build && echo \"Tests are not implemented.\" && exit 0"
   },
   "dependencies": {
-    "@slack/logger": "^1.0.0",
-    "@slack/web-api": "^5.1.0",
+    "@slack/logger": "^2.0.0",
```
aoberoi:
```suggestion
    "@slack/logger": ">=1.0.0 <3.0.0",
```

same reasoning as above ☝️

packages/web-api/package.json
```diff
@@ -41,8 +41,8 @@
     "coverage": "codecov -F webapi --root=$PWD"
   },
   "dependencies": {
-    "@slack/logger": "^1.0.0",
-    "@slack/types": "^1.1.0",
+    "@slack/logger": "^2.0.0",
```
aoberoi:
```suggestion
    "@slack/logger": ">=1.0.0 <3.0.0",
```

same reasoning as above ☝️

PR #895, opened 2019-10-24
Add types for radio buttons

shanedewael:
###  Summary

Supports type information for radio buttons in App Home

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -213,6 +213,7 @@ export interface RadioButtons extends Action {
   type: 'radio_buttons';
   initial_option?: Option;
   options: Option[];
+  confirm?: Confirm;
 }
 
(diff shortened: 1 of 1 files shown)


packages/types/src/index.ts
```diff
@@ -209,6 +209,12 @@ export interface Datepicker extends Action {
   confirm?: Confirm;
 }
 
+export interface RadioButtons extends Action {
+  type: 'radio_buttons';
+  initial_option?: Option;
+  options: Option[];
+}
```
seratch: Do we need to have `confirm` as well? or should we modify the api doc? https://api.slack.com/reference/block-kit/block-elements#radio

shanedewael: 🤦‍♂ You're right

PR #900, opened 2019-10-31
Add oauth.v2.access API support

seratch:
###  Summary

This pull request adds the support for `oauth.v2.access` endpoint introduced for Granular Permissions feature.

* https://api.slack.com/methods/oauth.v2.access
* https://api.slack.com/authentication/basics

Also, I noticed `oauth.access` has been missing an optional param named `single_channel`.

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -414,6 +414,26 @@ describe('WebClient', function () {
         scope.done();
       });
     });
+
+    it('should remove client_id/client_secret arguments and set Authorization header', function () {
+      const client = new WebClient(undefined);
+
+      const scope = nock('https://slack.com', {
+        reqheaders: {
+          'Authorization': 'Basic MTIzLjQ1NjpyYW5kb20tc3RyaW5n',
+        }
+      })
+        .post(/api/, 'code=123&other=abc')
+        .reply(200, { ok: true });
+
+        const r = client.apiCall(
+        'method',
+        { client_id: '123.456', client_secret: 'random-string', code: 123, other: 'abc', }
+      );
+      return r.then(() => {
+        scope.done();
+      });
+    });
   });
 
   describe('apiCall() - when using static headers', function () {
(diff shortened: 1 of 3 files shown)


packages/web-api/src/methods.ts
```diff
@@ -606,6 +606,13 @@ export interface OAuthAccessArguments extends WebAPICallOptions {
   client_secret: string;
   code: string;
   redirect_uri?: string;
+  single_channel?: string;
+}
+export interface OAuthV2AccessArguments extends WebAPICallOptions {
+  client_id: string;
```
shanedewael: https://api.slack.com/methods/oauth.v2.access notes that `client_id` and `client_secret` are optional?

aoberoi: i can try to verify, but i think they should be required.

seratch: @shanedewael hmm, I believe they're required. Rather than that, I noticed that this package should send them using basic auth instead as [the api doc strongly recommends](https://api.slack.com/methods/oauth.v2.access) (it's the same for v1). I will create a GitHub issue for the basic-auth issue and then make the change in this PR.

seratch: I've modified the implementation to send a base64 value in Authorization header instead of sending `client_id` and `client_secret` in `request.body` when detecting those in `options`. I believe this PR is ready to go.

packages/web-api/src/WebClient.ts
```diff
@@ -144,10 +144,20 @@ export class WebClient extends EventEmitter<WebClientEvent> {
...
-    const response = await this.makeRequest(method, Object.assign(
-      { token: this.token },
-      options,
-    ));
+    // Sending an encrypted value in Authorization header instead of sending them in request body
+    let headers = {};
+    if (typeof options !== 'undefined'
+      && typeof options.client_id !== 'undefined'
+      && typeof options.client_secret !== 'undefined') {
+      const credentials = Buffer.from(`${options.client_id}:${options.client_secret}`).toString('base64');
```
aoberoi: I believe this functionality belongs in the `serializeApiCallOptions()` method. all of the serialization-related logic is already there. that method is used with `axios`'s `transformRequest` pipeline and allows making modifications to `headers`.

packages/web-api/src/WebClient.spec.js
```diff
@@ -896,6 +896,47 @@ describe('WebClient', function () {
         });
   });
 
+  /*
```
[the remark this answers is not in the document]
aoberoi: I'm not sure what convenience these snippets add. In order to run them someone must first get an entire app configured and running. It feels like these snippets are present to make up for a lack of a unit test, which i think is a more appropriate way to verify the intended behavior.

seratch:
@aoberoi Thanks for your review. I just modified the code to follow your feedback and push it. But I noticed the code still doesn't work in unit tests using nock.

Thanks for cherry-picking the minimum part in #903 ! I think changing the way to pass client_id/client_secret is not so urgent that let me come up with another PR for it.

PR #904, opened 2019-11-08
Implement view submission and view closed handling on @slack/interactive-messages

aoberoi:
###  Summary

Implements view submission handling as described in #288 

### DIFF
--- a/packages/interactive-messages/src/adapter.ts
+++ b/packages/interactive-messages/src/adapter.ts
@@ -789,7 +789,7 @@ type ViewSubmissionHandler = (payload: any) => any | Promise<any> | undefined;
  *
  * TODO: describe the payload and return values more specifically?
  */
-type ViewClosedHandler = (payload: any) => any | Promise<any> | undefined;
+type ViewClosedHandler = (payload: any) => void;
 
 type Callback = ActionHandler | OptionsHandler | ViewSubmissionHandler | ViewClosedHandler;
 
--- a/packages/interactive-messages/src/adapter.ts
+++ b/packages/interactive-messages/src/adapter.ts
@@ -452,10 +452,15 @@ export class SlackMessageAdapter {
     return this.callbacks.find(([constraints]) => {
       // if the callback ID constraint is specified, only continue if it matches
       if (!isFalsy(constraints.callbackId)) {
-        if (isString(constraints.callbackId) && payload.callback_id !== constraints.callbackId) {
+        // The callback ID is located at a different path in the payload for view submission and view closed
+        const callbackId = (
+          constraints.handlerType === StoredConstraintsType.ViewSubmission ||
+          constraints.handlerType === StoredConstraintsType.ViewClosed
+        ) ? payload.view.callback_id : payload.callback_id;
+        if (isString(constraints.callbackId) && callbackId !== constraints.callbackId) {
           return false;
         }
-        if (isRegExp(constraints.callbackId) && !(constraints.callbackId as RegExp).test(payload.callback_id)) {
+        if (isRegExp(constraints.callbackId) && !(constraints.callbackId as RegExp).test(callbackId)) {
           return false;
         }
       }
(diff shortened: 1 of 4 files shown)


packages/interactive-messages/src/adapter.ts
```diff
@@ -640,4 +756,25 @@ type ActionHandler = (payload: any, respond: Respond) => any | Promise<any> | un
  */
 type OptionsHandler = (payload: any) => any | Promise<any> | undefined;
 
-type Callback = ActionHandler | OptionsHandler;
+/**
+ * A handler function for view submission requests.
+ *
+ * TODO: describe the payload and return values more specifically?
+ */
+type ViewSubmissionHandler = (payload: any) => any | Promise<any> | undefined;
```
shanedewael: The types for these are in Bolt if you want them - https://github.com/slackapi/bolt/blob/master/src/types/view/index.ts

packages/interactive-messages/src/adapter.ts
```diff
@@ -486,6 +543,52 @@ export class SlackMessageAdapter {
         }
       }
 
+      if (constraints.handlerType === StoredConstraintsType.ViewSubmission) {
```
shanedewael: These checks seem correct, but this and the ViewClosed constraints are the same. Is there a reason we can't combine them to reduce code duplication?

packages/interactive-messages/src/adapter.ts
```diff
@@ -573,22 +676,35 @@ export interface OptionsConstraints {
   within: 'block_actions' | 'interactive_message' | 'dialog';
 }
 
-type AnyConstraints = ActionConstraints | OptionsConstraints;
+/**
+ * Constraints on when to call a view submission or view closed handler.
+ */
+export interface ViewConstraints {
+  callbackId?: string | RegExp;
+  externalId?: string;
```
shanedewael: Should we allow `RegExp` for `externalId` too? Since they are unique on a per-user basis, it would be really hard to setup a handler for them with a type of string.

packages/interactive-messages/src/adapter.ts
aoberoi: Great idea!

packages/interactive-messages/src/adapter.ts
aoberoi: There's a slight difference which is the first conditional within each section. The first checks for `view_submission` and the second checks for `view_closed`. But now that you point it out there's likely a way to reduce the code duplication - thanks!

packages/interactive-messages/src/adapter.ts
aoberoi: Earlier I was trying not to expand scope so I didn't add RegExp matching to anything new - but now that you point it out that totally makes sense as a use case. It's not too much work to add that in now, so I might as well.

pitininja:
Hello,
I have been testing this feature on this branch.
When I do a view_submission with debug, the condition in `adapter.ts` on line 454 is always matched because `callback_id` is not in `payload` but in `payload.view`.
```
// from matchCallback function
if (isString(constraints.callbackId) && payload.callback_id !== constraints.callbackId) {
    return false;
}
```
Maybe I am doing something wrong?
Sorry if I should not be posted here 🙏 
Thank you for your work

aoberoi:
@pitininja you're totally right! thanks for catching that.

packages/interactive-messages/src/adapter.ts
aoberoi: On second thought - the options are to depend on `@slack/bolt` to get these types, to duplicate those types in `@slack/interactive-messages`, or to move those types into `@slack/types` (and then refactor Bolt and this package to grab those types from that package).

I think the ideal solution is the last one, but I don't think its wise to increase scope of this feature to get that done. In the meantime, I think we can proceed with this approach and revisit the refactoring of all payload types into `@slack/types` at a later time. Many other packages might also benefit from this.

aoberoi:
@pitininja fixed the issue you pointed out (and the ones @shanedewael pointed out too).

we're likely releasing this change tomorrow.

Issue #748, opened 2019-03-28
Update maintainers guide for mono repo structure


aoberoi:
### Description

Produce an updated maintainers guide for all the common workflows. This should include new information around how changelogs are published, new tag naming conventions, etc. There may need to be new documentation specific to managing the monorepo and some of the decisions that were put into this structure (thinking a high level philosophy and an FAQ). These materials are very important for continued collaboration with the community as well as onboarding new maintainers.

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

stevengill:
I'm reading through https://github.com/slackapi/node-slack-sdk/blob/master/.github/maintainers_guide.md#testing currently. I see that their is a todo for the following:
> update the following information for debugging with VSCode or with the protocol directly.

I've read through a guide + some stack overflow threads and got debugging working in vs code. I'd like to contribute my findings back to the maintainers guide. A few questions and thoughts though:

* Should I commit my `.vscode` folder? It contains `launch.json` which holds the configurations for debugging. Some [stack overflow posters](https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control) seem to say it is fine to do so. I just don't want to overwrite other maintainers `launch.json` files as I don't know what they might have in it. I could check it in as `launch.json.default` instead so maintainers have a choice if they want to use it or pull what they want out of it. Final resort would be I don't check anything in and just document it in the maintainers guide. 
* I currently have a config figured out that would allow debugging on all of the tests in a module at a time. Example:
```
        {
            "type": "node",
            "request": "launch",
            "name": "Mocha Tests",
            "program": "${workspaceFolder}/packages/events-api/node_modules/mocha/bin/_mocha",
            "args": [
                "--timeout",
                "999999",
                "--colors",
                "${workspaceFolder}/packages/events-api/test/integration/*.js",
                "${workspaceFolder}/packages/events-api/src/*.spec.js"
            ],
            "internalConsoleOptions": "openOnSessionStart",
            "skipFiles": [
                "<node_internals>/**"
            ],
            "cwd": "${workspaceFolder}/packages/events-api"
        }
    ]
[...]

PR #913, opened 2019-11-20
Added VS Code launch.json debugging configurations

stevengill:
###  Summary

I've added a `launch.json` to each package that contains configuration for running node debugging in `vscode`. I've also updated the maintainers guide with instructions on how to do the debugging. 

I also added a root `.vscode` folder. This makes it possible to run debugging from the root directory instead of in each package. It just allows flexibility for maintainers to open the entire `node-slack-sdk` directory in `vscode` and run debugging vs opening each package individually in `vscode` and run debugging if they please. 

Related to issue: #748 


.vscode/launch.json
```diff
@@ -0,0 +1,111 @@
+{
```
aoberoi: I don't think this file is actually necessary. By default, when you open a VSCode workspace, [the launch configurations from each of the subprojects is listed](https://code.visualstudio.com/docs/editor/multi-root-workspaces#_debugging) in the Debug pane.

If its not showing up, it's likely because you're simply opening the folder in VSCode, rather than opening the workspace. This is a pretty small detail, but if you're using the CLI, and your in the root of this repo, you'd run `code node-slack-sdk.code-workspace` instead of `code .`. Even if you do the latter, or you open VSCode and use the GUI to open the project, VSCode will warn you using a toast in the bottom right corner, and make it one click to open the workspace.

Screen Shot 2019-11-22 at 1 47 22 PM

The problem with adding this file redundantly, just in case you opened the folder instead of the workspace, is that the Debug pane lists a duplicated list of tasks:

Screen Shot 2019-11-22 at 1 43 32 PM

integration-tests/.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
...
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Integration Tests",
+            "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+            "args": [
+                "--timeout",
+                "999999",
+                "--colors",
+                "${workspaceFolder}/test/integration/*.js",
```
aoberoi: this doesn't seem to match any actual files.

integration-tests/.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
...
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Integration Tests",
+            "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
```
aoberoi: This launch config doesn't support the other part of running tests in here, which is `dtslint` (see scripts in `package.json`). Could we support the larger `npm run test` script, or is that kind of wild and not helpful for debugging?

aoberoi:
This is such a big improvement for debugging issues!!

I wanted to get these comments to you ASAP to unblock any updates you want to make as a result. I haven't yet run the tests but I'll also do that as soon as I have a few minutes, so there might be some more comments coming your way.

.vscode/launch.json
stevengill: Ahhh. I almost always do `code .` vs opening a workspace. Nice tip. I was already thinking that top level `launch.json` was a little redundant. Seeing the dupes in the workspace method reinforces that. Thanks!

integration-tests/.vscode/launch.json
stevengill: Nice catch. Accidental artifact from a copy/paste

integration-tests/.vscode/launch.json
stevengill: launch.json doesn't support `npm run test` unfortunately. I did just add a new config named `Integration Type Checker`. This runs `dtslint`. 

Sidenote: When I run `dtslint` in my CLI, I'm not seeing any output except blank console.logs. I'm seeing the same when I run the configuration in vscode.

stevengill:
Appreciate the review @aoberoi! Let me know if you have any other comments after you have a chance to run the tests.

PR #936, opened 2019-12-25
Fix a code error in events-api documentation

seratch:
###  Summary

This pull request fixes a code sample error in `@slack/events-api` documentation.

```
(async () => {
^

TypeError: slackEvents.on(...) is not a function
```

Just in case, I assigned @aoberoi as he is the original author. But I believe this one is safe enough. I may merge this within a few days.

### DIFF
--- a/docs/_packages/events_api.md
+++ b/docs/_packages/events_api.md
@@ -366,7 +366,7 @@ slackEvents.on('reaction_added', (event, respond) => {
   console.log('Reaction event received');
   // Normal success
   respond();
-})
+});
 
 (async () => {
   const server = await slackEvents.start(port);


docs/_packages/events_api.md
```diff
@@ -366,7 +366,7 @@ slackEvents.on('reaction_added', (event, respond) => {
   console.log('Reaction event received');
   // Normal success
   respond();
-})
+});
```
stevengill: Is this correct? I'm only seeing an addition of this semicolon as the only change.

seratch: Yes, it is. Without the semicolon, the code doesn't work 😢

stevengill: weird. I have approved it

Issue #945, opened 2020-01-13
app.error doesn't let me inspect the complete error object


jammiemountz:
### Description

APIs supporting Block Kit will respond with `invalid_blocks` if the payload is invalid, which the app will log with `app.error` callback. But, the error payload response from the API will actually include really helpful information about WHAT about the payload is invalid, that I don't seem to have access to in that callback.

You can see the larger helpful error messaging in the network tab for the block kit builder:
<img width="1377" alt="Screen Shot 2020-01-09 at 11 39 40 AM" src="https://user-images.githubusercontent.com/8782928/72280425-273c3580-35ed-11ea-8d55-483b33936c99.png">

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/bolt/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

#### Reproducible in:

package version:
"@slack/bolt": "v1.4.1"

#### Steps to reproduce:

1. Use bolt to send an invalid payload to an api that supports block kit. (aka sending block type "sections" instead of "section")
2. Log the error that comes through in the `app.error` callback
3. See that the information about which block is causing the error is not included

#### Expected result:

I would expect to at least have access to the information within the callback, even if it's not logged by default

#### Actual result:

Inspecting the "error" object does not have any other information than the error string.

stevengill:
Sorry, finally back to this. @aoberoi wanted me to surface up those messages a little differently. 

In your terminal, the error would look like the following now:

```
[ERROR]    unsupported type: sections [json-pointer:/blocks/0/type]
[ERROR]   Error: An API error occurred: invalid_blocks
```

Any Errors or Warnings in `response_metadata.messages` will now be outputted on their own line. Sending PR now.

For future context, this is related to https://api.slack.com/changelog/2016-09-28-response-metadata-is-on-the-way

stevengill:
Updated the existing PR: https://github.com/slackapi/node-slack-sdk/pull/946

stevengill:
[...]

PR #946, opened 2020-01-21
surface up error message from platform api calls

stevengill:
###  Summary

Issue number: #945 

On failed platform api calls which contain blocks, we don't properly surface up the error message from the api. This PR solves that. 

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -86,7 +86,11 @@ describe('WebClient', function () {
           .post(/api/)
           .reply(200, { ok: true,
             response_metadata: {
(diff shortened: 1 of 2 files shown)


aoberoi:
@stevengill and i talked about how picking the first message in the array doesn't feel like a complete solution. especially because there's no particular reason why `error.data.message` would be significantly more visible than where the message already was.

instead, we'd like to put the message(s) in a place where it's available to humans to read. the best place for this is the `message` property of the error itself.

so the suggestion is to flatten the array of messages in `result.response_metadata.messages` and use them as the first argument to the `Error` constructor (L94).

stevengill:
@aoberoi This is ready for review

PR #960, opened 2020-02-11
Fix type error with `for await (... of paginate)`

ypresto:
###  Summary

Writing `for await (const page of client.paginate("conversations.history", args)) {` produces below error:

```
Type 'AsyncIterator<WebAPICallResult, any, undefined>' must have a '[Symbol.asyncIterator]()' method that returns an async iterator.ts(2504)
```
In the function body (not the return type) AsyncIterableIterator is used and it has `[Symbol.asyncIterator]()`, this PR changes return type to it.


clavin:
Oh, it seems like this sort of change landed in #1040! Since that PR already landed I'm going to close this one, but feel free to @ me if you think this should be reopened!

Issue #979, opened 2020-03-30
Support radio_buttons block element in input block for TypeScript users


seratch:
### Description

`radio_buttons` block elements now work as part of input blocks. We can change the type definition of `InputBlock` interface.

Reference: https://github.com/slackapi/bolt/issues/450

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

PR #994, opened 2020-04-15
Fix #993 Support checkboxes block element in input block for TypeScript users.

### LINKED ISSUE #993: Support checkboxes block element in input block for TypeScript users
### Description

Like #979, the checkboxes block element should also be supported in input block for TypeScript users. 

Screenshot_20200415_145109

- [ ] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

yoichiro:
###  Summary

This pull request fixes #993 by updating type definitions for Block Kit like #979. For instance, the checkboxes block element should be supported in input block. In addition, it should also be supported in section block. This provides additional definitions to support them.

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -275,7 +275,7 @@ export interface SectionBlock extends Block {
   type: 'section';
   text?: PlainTextElement | MrkdwnElement; // either this or fields must be defined
   fields?: (PlainTextElement | MrkdwnElement)[]; // either this or text must be defined
-  accessory?: Button | Overflow | Datepicker | Select | MultiSelect | Action | ImageElement | RadioButtons;
+  accessory?: Button | Overflow | Datepicker | Select | MultiSelect | Action | ImageElement | RadioButtons | Checkboxes;
 }
 
 export interface FileBlock extends Block {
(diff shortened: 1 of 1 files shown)


packages/types/src/index.ts
```diff
@@ -275,7 +275,7 @@ export interface SectionBlock extends Block {
   type: 'section';
   text?: PlainTextElement | MrkdwnElement; // either this or fields must be defined
   fields?: (PlainTextElement | MrkdwnElement)[]; // either this or text must be defined
-  accessory?: Button | Overflow | Datepicker | Select | MultiSelect | Action | ImageElement | RadioButtons;
+  accessory?: Button | Overflow | Datepicker | Select | MultiSelect | Action | ImageElement | RadioButtons | Checkboxes;
```
seratch, line 278: Just for future reference: A `checkboxes` element in a section block is not allowed in messages. That said, considering the lack of ways to distinguish if the container of blocks is either a message or a view, I think there is no better way than just adding `Checkboxes` type for modals and Home tabs.

seratch:
@yoichiro Thanks again! This change looks good to me. Just in case, I will wait for other maintainers' response until tomorrow.

packages/types/src/index.ts
yoichiro: > A checkboxes element in a section block is not allowed in messages.

@seratch Ah, right. But, yeah, I also think that it would be better to choose a benefit to add `Checkboxes` type for both modal and home tab at this time.

yoichiro:
@seratch @stevengill Thank you for the reviewing and merging! :)

Issue #1002, opened 2020-04-26
views.update method doesn't accept only external_id


suyashgulati:
### Description

The views.update method does not accept object with only external_id, as the [documentation](https://api.slack.com/methods/views.update#arg_external_id) says it should accept but the typings don't allow it without view_id [code reference](https://github.com/slackapi/node-slack-sdk/blob/a4f3d5044ef66ee35b446f083ff0495f2560d477/packages/web-api/src/methods.ts#L942).

![Screen Shot 2020-04-27 at 3 06 56 AM](https://user-images.githubusercontent.com/20201768/80320762-f697a900-8835-11ea-9d46-4d82861f2306.png)

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [X] bug
- [X] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [X] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [X] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [X] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Packages:

Select all that apply:

- [X] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/interactive-messages`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] I don't know

#### Reproducible in:

package version: 2.0.0

node version: 12.13.1

OS version(s): MacOS 10.13.6

#### Steps to reproduce:

1.
2.
3.

#### Expected result:

Should allow ViewsUpdateArguments with only external_id or view_id

#### Actual result:

Currently view_id is required argument which should not be the case

#### Attachments:

Logs, screenshots, screencast, sample project, funny gif, etc.

seratch:
As #1003 has been reverted, this issue is not yet fixed. https://github.com/slackapi/node-slack-sdk/blob/%40slack/web-api%405.10.0/packages/web-api/src/methods.ts#L991-L996

We should fix this issue in the next version - https://github.com/slackapi/node-slack-sdk/pull/1003#discussion_r437157719

PR #1003, opened 2020-04-26
Divided ViewUpdateArguments into 2 types

suyashgulati:
###  Summary

Divided ViewUpdateArguments into 2 types ViewsUpdateArgumentsWithViewId | ViewsUpdateArgumentsWithExternalId. This is a fix for #1002 

* [X] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.

# Challenges
I was not able to test this code yet. Lerna tests was giving me some error, If someone can help me with it?

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -471,7 +471,8 @@ export class WebClient extends EventEmitter<WebClientEvent> {
     open: (this.apiCall.bind(this, 'views.open')) as Method<methods.ViewsOpenArguments>,
     publish: (this.apiCall.bind(this, 'views.publish')) as Method<methods.ViewsPublishArguments>,
     push: (this.apiCall.bind(this, 'views.push')) as Method<methods.ViewsPushArguments>,
-    update: (this.apiCall.bind(this, 'views.update')) as Method<methods.ViewsUpdateArgumentsWithViewId | methods.ViewsUpdateArgumentsWithExternalId>,
+    update: (this.apiCall.bind(this, 'views.update')) as
+      Method<methods.ViewsUpdateArgumentsWithViewId | methods.ViewsUpdateArgumentsWithExternalId>,
   };
 
   /**
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -471,7 +471,7 @@ export class WebClient extends EventEmitter<WebClientEvent> {
     open: (this.apiCall.bind(this, 'views.open')) as Method<methods.ViewsOpenArguments>,
     publish: (this.apiCall.bind(this, 'views.publish')) as Method<methods.ViewsPublishArguments>,
     push: (this.apiCall.bind(this, 'views.push')) as Method<methods.ViewsPushArguments>,
-    update: (this.apiCall.bind(this, 'views.update')) as Method<methods.ViewsUpdateArguments>,
+    update: (this.apiCall.bind(this, 'views.update')) as Method<methods.ViewsUpdateArgumentsWithViewId | methods.ViewsUpdateArgumentsWithExternalId>,
   };
 
   /**
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -940,10 +940,16 @@ export interface ViewsPublishArguments extends WebAPICallOptions, TokenOverridab
 }
 
 export interface ViewsUpdateArguments extends WebAPICallOptions, TokenOverridable {
-  view_id: string;
   view: View;
-  external_id?: string;
   hash?: string;
+}
+
+export interface ViewsUpdateArgumentsWithViewId extends ViewsUpdateArguments {
+  view_id: string;
+}
+
+export interface ViewsUpdateArgumentsWithExternalId extends ViewsUpdateArguments {
+  external_id: string;
 }
 
 export * from '@slack/types';


seratch:
@aoberoi @stevengill I like this improvements. Thoughts?
@suyashgulati If you don't mind, could you sign the CLA? https://github.com/slackapi/node-slack-sdk/pull/1003#issuecomment-619632989

suyashgulati:
> @aoberoi @stevengill I like this improvements. Thoughts?
> @suyashgulati If you don't mind, could you sign the CLA? [#1003 (comment)](https://github.com/slackapi/node-slack-sdk/pull/1003#issuecomment-619632989)

Hi @seratch I did that already, but it seems my commit was not associated with any github user (didn't configure user.email/user.name and no globals as well). I may need to re create a PR with new commit. 
Or any other solution?
Screen Shot 2020-04-27 at 2 52 20 PM

seratch:
@suyashgulati How about doing as below?

```
git reset HEAD~1
git add . -v
git config user.email {your github email address}
git commit -m'{your commit message}'
git push origin fix-issue-1002 -f
```

packages/web-api/src/WebClient.ts
```diff
@@ -471,7 +471,8 @@ export class WebClient extends EventEmitter<WebClientEvent> {
     open: (this.apiCall.bind(this, 'views.open')) as Method<methods.ViewsOpenArguments>,
     publish: (this.apiCall.bind(this, 'views.publish')) as Method<methods.ViewsPublishArguments>,
     push: (this.apiCall.bind(this, 'views.push')) as Method<methods.ViewsPushArguments>,
-    update: (this.apiCall.bind(this, 'views.update')) as Method<methods.ViewsUpdateArguments>,
+    update: (this.apiCall.bind(this, 'views.update')) as
```
aoberoi, line 474: i'm a little concerned that this is a break from the existing pattern. there are no other methods which use a union type here.

if we want to do this, i would rather seem a `type ViewUpdateArguements = ViewsUpdateArgumentsWithViewId | ViewsUpdateArgumentsWithExternalId` in `methods.ts` so that the complexity of underlying shapes stays out of `WebClient`.

packages/web-api/src/methods.ts
```diff
@@ -940,10 +940,16 @@ export interface ViewsPublishArguments extends WebAPICallOptions, TokenOverridab
...
 
 export interface ViewsUpdateArguments extends WebAPICallOptions, TokenOverridable {
-  view_id: string;
   view: View;
-  external_id?: string;
   hash?: string;
 }
 
+export interface ViewsUpdateArgumentsWithViewId extends ViewsUpdateArguments {
+  view_id: string;
```
aoberoi, line 948: Another way of fixing #1002 would be to make the `view_id` property optional. 

This change is technically more correct than what we have today. But it does breaks the pattern already established in this file. In fact, there are many arguments interfaces that could benefit from a union of several types like this, but we held ourselves back from making that change because we have an eventual goal of generating these types from an API specification. It's not clear how we'd be able to generate these more complex types from a specification because the specification we have today doesn't express these kinds of interdependencies.

aoberoi: if we keep this divided interface, i think it would be more accurate to still include `external_id` as optional when `view_id` exists, and vice versa. One of the two is required, but its okay to specify both.

aoberoi:
Sorry for the late review, this was on my plate for longer than I would have liked. @stevengill, given the comments i left above, do you think it would be worth reverting this merge? maybe we eventually want this change, but i think there's a discussion to be had.

packages/web-api/src/WebClient.ts
stevengill: That's a good point. Doing a quick scan, I don't see any union type methods like your suggested `ViewUpdateArugments` in the `mehods.ts` file. I know though that it is a pattern we have used in `bolt`.

packages/web-api/src/methods.ts
stevengill: Agreed that it would be more accurate to add `external_id` and `view_id` as optional.

suyashgulati: I don't think it is a good solution to make both of them optional, as a developer might think both of them can be left blank. Which is completely incorrect!

A better solution than this would be to make `ViewUpdateArugments = ViewsUpdateArgumentsWithViewId | ViewsUpdateArgumentsWithExternalId;` in method.ts so that this wouldn't break the symmetry like suggested here https://github.com/slackapi/node-slack-sdk/pull/1003/files#r417527527

stevengill: @suyashgulati sorry let me clarify the point @aoberoi was trying to make with the optional. 

I think for `ViewsUpdateArgumentsWithViewId` would have `external_id` set to optional, and `ViewsUpdateArgumentsWithExternalId` would set `view_id` to optional. This is because a developer could still include both `external_id` as well as `view_id`

suyashgulati: Ok I understood now.
Just a query, will adding a view_id makes a difference when I am using external_id for view reference?

seratch: @aoberoi @stevengill We've reverted this PR but haven't made the field optional yet. Let's re-fix #1002 when you have a chance.

Issue #1010, opened 2020-05-06
Improve successful Oauth flow for users


stevengill:
### Description

Currently, when oauth is successful, users are redirected back to the native slack app. The allow permissions tab is left open when a spinning button. It is currently expected users should close this tab manually since they were redirected back to the app. We can make this flow better.

Let's add a redirect page that takes users away from the the allow permissions tab indicating success and handles the redirect back to slack native app. 

Related issue: https://github.com/slackapi/bolt-js/issues/492

### What type of issue is this? (place an `x` in one of the `[ ]`)

- [x] enhancement (feature request)

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/interactive-messages`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [x] `@slack/oauth`
- [ ] I don't know

PR #1011, opened 2020-05-06
improved oauth success user experience flow

stevengill:
###  Summary

Fix for issue #1010.

It will now show a generic success page which will automatically launch the redirect to the slack app. It also includes a button to launch the slack app incase users have javascript turned off or this click no during the original popup window. 

This is an improved flow vs the allow permissions tab staying open with a spinning button

Screen Shot 2020-05-06 at 4 25 06 PM

### DIFF
--- a/packages/oauth/README.md
+++ b/packages/oauth/README.md
@@ -140,11 +140,15 @@ If you decide you need custom success or failure behaviors (ex: wanting to show 
 const callbackOptions = {
   success: (installation, metadata, req, res) => {
     // Do custom success logic here
-    res.send('successful!');
+    // tip: you can add javascript and css in the htmlResponse using the <script> and <style> tags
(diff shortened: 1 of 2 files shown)


stevengill:
Let me know if the redirect message should be tweaked.

aoberoi:
This is a good improvement! I can imagine folks wanting to update the styles on the success page to look prettier. Maybe we can update or extend the sub-task in the [example for customizing the success handler](https://github.com/slackapi/node-slack-sdk/tree/master/packages/oauth#handling-the-oauth-redirect) to show how someone could render their own HTML page instead of only showing `res.send('success')`? I also just noticed that `res.send()` assumes use of Express. Should it be `res.end()`?

stevengill:
Nice catch! I'll update that example part of this PR.

PR #1013, opened 2020-05-07
feat: @slack/oauth: add support and examples for CSRF mitigation

losandes:
###  Summary

> NOTE that #1014 presents an alternative solution to this PR

Adds support, and examples for mitigating Cross Site Request Forgery (CSRF) by synchronizing all 3 parties (user, Slack, app) in the OAuth flow. Uses the synchronizer token pattern (STP) and a JWT embedded in a cookie to bind the token to the device.

* Adds new `makeInstallUrl` function that returns the URL that is produced, and the token that is used in the state parameter, and the synchronizer value that is in the token
* Adds a 256 bit cryptographically random byte array (synchronizer) to the JWT body, so it can be timing safe compared with the same array stored on a device when verifying the redirect from Slack
* Updates the examples to add an http-only, secure cookie to the device when starting the OAuth flow, and then parses the cookie so the synchronizer can be compared with the JWT
* Updates examples to recommend using secrets of at least 256 bits
* Reduces the lifetime of JWT tokens, so they can't be used forever
* Adds support for configuring the lifetime of JWT tokens

#### Concerns / Limitations

This leaves responsibility for synchronizing the device to the consumer of this library. Since the library already deals with `req`, and `res`, it might make sense to support device synchronization as part of the library's features. That's a heavier lift, so this PR is intended to support conversation about the best way to promote CSRF mitigation when using this library.

Also, the README doesn't discuss CSRF at all. Even though it's addressed in the examples, it might be worth adding content to the README. For instance, we might want to warn against using the same token in the cookie, and the state param, since that makes it easier to spoof a synchronized device.

### DIFF
--- a/examples/oauth-v1/app.js
+++ b/examples/oauth-v1/app.js
@@ -2,6 +2,7 @@ const { InstallProvider } = require('@slack/oauth');
 const { createEventAdapter } = require('@slack/events-api');
 const { WebClient } = require('@slack/web-api');
 const express = require('express');
+const cookie = require('cookie');
 
 const app = express();
 const port = 3000;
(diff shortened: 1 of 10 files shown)


dinvlad:
This is really nice, along with #1014, thank you for these detailed examples!

I'd like to understand the pros/cons of various options for the cookie and the `state`:
1. JWT+synchronizer cookie, synchronizer state (#1014)
2. JWT+synchronizer cookie, JWT+synchronizer state (#1013)
3. Synchronizer cookie, JWT+synchronizer state (which from your note above, is less secure than 2, since it doesn't have a timing component on the cookie, besides browser expiration)
4. Encrypted versions of these (with or without JWT signing)

Is it purely a matter of taste, or are there security benefits to any of these approaches? 

Also, just to be a little pedantic here - strictly speaking, the [OWASP guide on  STP](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern) says that
> CSRF tokens should not be transmitted using cookies.

and notes that STP is more of a stateful protocol, where the CSRF token should be stored in user session on the server (instead of a cookie). For stateless apps, it's more appropriate to use Encryption, Hash, or Double Submit Cookie TP. The last one seems closest to what we're using here (though we're using HMAC+cookie+state instead of cookie+cookie). In any case, whichever terminology is used, they seem to recommend also encrypting the CSRF cookie to enhance the security of this solution. Any thoughts about the necessity of it?

Thanks again!

PR #1015, opened 2020-05-12
Add admin.usergroups.* APIs for Grid org admins

seratch:
###  Summary

This pull request adds admin.usergroups.* APIs for Grid org admins.

* admin.usergroups.addChannels
* admin.usergroups.listChannels
* admin.usergroups.removeChannels

Refer to https://api.slack.com/enterprise/managing#defaultchannels for details.

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -151,6 +151,20 @@ export interface AdminTeamsSettingsSetNameArguments extends WebAPICallOptions, T
   team_id: string;
   name: string;
 }
+export interface AdminUsergroupsAddChannelsArguments extends WebAPICallOptions, TokenOverridable {
+  usergroup_id: string;
+  team_id: string;
+  channel_ids: string | string[];
+}
+export interface AdminUsergroupsListChannelsArguments extends WebAPICallOptions, TokenOverridable {
+  usergroup_id: string;
+  include_num_members?: boolean;
+  team_id?: string;
+}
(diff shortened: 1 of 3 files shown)


packages/web-api/src/methods.ts
```diff
@@ -151,6 +151,20 @@ export interface AdminTeamsSettingsSetNameArguments extends WebAPICallOptions, T
   team_id: string;
   name: string;
 }
+export interface AdminUsergroupsAddChannelsArguments extends WebAPICallOptions, TokenOverridable {
+  usergroup_id: string;
+  team_id: string;
+  channel_ids: string | string[];
```
seratch, line 157: This can be out-of-scope of this PR but `methods.ts` is not consistent regarding the type of "ids" argument (e.g., channel_ids, users). Some are `string` while others are `string[]`.

Regarding this endpoint, Slack API works with either `"C123,C234"` (the surely compatible way with the API docs) or `["C123", "C234"]`. Doing the same for other endpoints may work but I haven't checked yet. 

@stevengill @aoberoi 
To make this consistent, we can modify the logic in `WebClient#serializeApiCallOptions` to consider an array object (=this means do `channel_ids.join(",")` instead of `JSON.stringify` anyway) and change some argument types to `string | string[]`. I think it's totally fine. Any thoughts?

stevengill: sounds like a good change to me

PR #1036, opened 2020-06-10
Move method bindings from `WebClient.ts` to `methods.ts`

clavin:
###  Summary

Everything still works as it has. There are no outside changes. Nothing but moving source code.

**Before**:
```typescript
/* WebClient.ts */

class WebClient extends EventEmitter<WebClientEvent> {
  // ...

  public readonly chat = {
    meMessage: (this.apiCall.bind(this, 'chat.meMessage')) as Method<methods.ChatMeMessageArguments>,
    postEphemeral: (this.apiCall.bind(this, 'chat.postEphemeral')) as Method<methods.ChatPostEphemeralArguments>,
    postMessage: (this.apiCall.bind(this, 'chat.postMessage')) as Method<methods.ChatPostMessageArguments>,
    // (more methods...)
  };
  // (and hundreds more methods...)
}
```

**After**:

```typescript
/* WebClient.ts */

class WebClient extends Methods {
  // ...

  // (no noise from hundreds of lines of method bindings)
}
```

```typescript
/* methods.ts */

export abstract class Methods extends EventEmitter<WebClientEvent> {
  // (abstract `apiCall` method)

  public readonly chat = {
    postMessage: bindApiCall<ChatPostMessageArguments, WebAPICallResult>(this, 'chat.postMessage'),
    // (more methods...)
  };
  // (and hundreds more methods...)
}
```

---

There's two rationales for this change:

* `WebClient.ts` is complicated and large without the method bindings; cutting out some of the noise would be nice.
* The `import * as methods from './methods';` at the top of `WebClient.ts` (which prevents it from having hundreds of imports for method arguments) is holding back #831.

I also took this opportunity to add a response type parameter to the `Method` interface in `methods.ts`. 🙂

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -996,7 +996,7 @@ describe('WebClient', function () {
     }
     const client = new WebClient(token);
 
-    it('can call admin.usergroups.addChannels with a string "channe_ids"', function (done) {
+    it('can call admin.usergroups.addChannels with a string "channel_ids"', function (done) {
       verify(
         client.admin.usergroups.addChannels({ team_id: 'T123', usergroup_id: 'S123', channel_ids: 'C123,C234' }),
         'admin.usergroups.addChannels',
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -11,13 +11,12 @@ import { Readable } from 'stream';
 import { SecureContextOptions } from 'tls';
 
 import isStream from 'is-stream';
-import { EventEmitter } from 'eventemitter3';
 import PQueue from 'p-queue'; // tslint:disable-line:import-name
 import pRetry, { AbortError } from 'p-retry';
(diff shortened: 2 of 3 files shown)


seratch:
The integration tests are failing but the change itself looks good to me.
https://travis-ci.org/github/slackapi/node-slack-sdk/jobs/696988175

clavin:
Oops, was importing `EventEmitter` in a way that the integration test didn't like (but everything else was fine with........?)

I changed it to match what it was before though, better consistency!

packages/web-api/src/WebClient.spec.js
```diff
@@ -996,7 +996,7 @@ describe('WebClient', function () {
     }
     const client = new WebClient(token);
 
-    it('can call admin.usergroups.addChannels with a string "channe_ids"', function (done) {
+    it('can call admin.usergroups.addChannels with a string "channel_ids"', function (done) {
```
seratch, line 999: Good catch 🙇

aoberoi:
This looks great!

PR #1042, opened 2020-06-19
Add error to WebClient's `Methods` superclass

clavin:
###  Summary

As an tag-on to #1036, this PR aims to prevent package consumers from using the new `Methods` abstract class as it is currently meant to only serve `WebClient` specifically.

If someone attempts to create a class that `extends Methods` then it should error when being created with `new`, unless that class `extends WebClient`.

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -2,7 +2,7 @@ require('mocha');
 const fs = require('fs');
 const path = require('path');
 const { Agent } = require('https');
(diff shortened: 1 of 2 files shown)


clavin:
Added those tests @aoberoi 👍! I found that it doesn't make sense to directly try `new Methods()` as that means you are not supplying the `apiCall` method it needs in the constructor and will just error ("cannot call bind on undefined").

aoberoi:
Looks great!

PR #1046, opened 2020-07-04
Adding support for org installations to oauth library

stevengill:
###  Summary

Adding support for org installations to oauth library. 

### DIFF
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -165,15 +165,34 @@ const installer = new InstallProvider({
     // returns nothing
     storeInstallation: (installation) => {
       // replace myDB.set with your own database or OEM setter
-      myDB.set(installation.team.id, installation);
+      if (installation.isEnterpriseInstall && installation.enterprise !== undefined) {
+        // enterprise app, org wide installation
+        myDB.set(installation.enterprise.id, installation);
+      } else if (installation.team.id !== undefined) {
+        // non enterprise org app installation
+        myDB.set(installation.team.id, installation);
+      } else {
+        throw new Error('Failed saving installation data to installationStore');
+      }
       return;
     },
     // takes in an installQuery as an argument
     // installQuery = {teamId: 'string', enterpriseId: 'string', userId: string, conversationId: 'string'};
     // returns installation object from database
-    fetchInstallation: (installQuery) => {
+    fetchInstallation: async (installQuery) => {
       // replace myDB.get with your own database or OEM getter
-      return myDB.get(installQuery.teamId);
+      let result = undefined;
+      // enterprise org app installation lookup
+      if (InstallQuery.enterpriseId !== undefined) {
+        result = await myDB.get(installQuery.enterpriseId);
+      }
+
+      // non enterprise org app lookup
+      if (result === undefined && InstallQuery.teamId !== undefined) {
+        result = await myDB.get(installQuery.teamId);
+      }
+
+      return result;
     },
   },
 });
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -25,7 +25,7 @@ It may be helpful to read the tutorials on [getting started](https://slack.dev/n
 
 ### Initialize the installer
 
-This package exposes an `InstallProvider` class, which sets up the required configuration and exposes methods such as `generateInstallUrl`, `handleCallback`, `authorize` for use within your apps. At a minimum, `InstallProvider` takes a `clientId` and `clientSecret` (both which can be obtained under the **Basic Information** of your app configuration). `InstallProvider` also requires a `stateSecret`, which is used to encode the generated state, and later used to decode that same state to verify it wasn't tampered with during the OAuth flow. **Note**: This example is not ready for production because it only stores installations (tokens) in memory. Please go to the [storing installations in a database](#storing-installations-in-a-database) section to learn how to plug in your own database.
+This package exposes an `InstallProvider` class, which sets up the required configuration and exposes methods such as `generateInstallUrl`, `handleCallback`, `authorize`, `orgAuthorize` for use within your apps. At a minimum, `InstallProvider` takes a `clientId` and `clientSecret` (both which can be obtained under the **Basic Information** of your app configuration). `InstallProvider` also requires a `stateSecret`, which is used to encode the generated state, and later used to decode that same state to verify it wasn't tampered with during the OAuth flow. **Note**: This example is not ready for production because it only stores installations (tokens) in memory. Please go to the [storing installations in a database](#storing-installations-in-a-database) section to learn how to plug in your own database.
 
 ```javascript
 const { InstallProvider } = require('@slack/oauth');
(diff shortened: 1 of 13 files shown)


packages/oauth/src/index.ts
```diff
@@ -100,6 +101,36 @@ export class InstallProvider {
...
+    try {
+      if (this.installationStore.fetchOrgInstallation === undefined) {
+        throw new Error('Installation Store is missing the fetchOrgInstallation method');
+      }
+
+      const queryResult = await this.installationStore.fetchOrgInstallation(source, this.logger);
+
+      if (queryResult === undefined) {
+        throw new Error('Failed fetching data from the Installation Store');
+      }
```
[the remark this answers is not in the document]
stevengill: @aoberoi so `queryResult.team.id` will not work here since this is an org installation (team is set to null). So for `orgAuthorize`, I assume `queryResult.team.id` is undefined/null and only check for `source.teamId`. If that exists, I add it to `authResult`

packages/oauth/src/index.ts
```diff
@@ -250,30 +292,51 @@ export class InstallProvider {
         }) as unknown as OAuthV2Response;
 
         // get botId
-        const botId = await getBotId(resp.access_token, this.clientOptions);
+        const authResult = await runAuthTest(resp.access_token, this.clientOptions);
+        const botId = authResult.bot_id;
+        const orgDashboardGrantAccess = authResult.url;
```
seratch, line 297: @stevengill Is there any reason you chose this name for the URL? As we fetch the value from `auth.test` this way:

```json
{
  "ok": true,
  "url": "https://your-domain.enterprise.slack.com/",
  "team": "Org Name",
  "user": "org-level-app",
  "team_id": "E111",
  "user_id": "W111",
  "bot_id": "B111",
  "enterprise_id": "E111",
  "is_enterprise_install": true
}
```

plus, the value is just `https://{org domain}.enterprise.slack.com/`, I prefer using a name indicating this is a URL. 

Let me share a few candidates:

* 1: `orgDashboardBaseUrl`, `org_dashboard_base_url` in Python
* 2: `enterpriseUrl` in JS/Java, `enterprise_url` in Python
* 3: `enterprise.url` in JS, `enterprise_url` in Python, `enterpriseUrl` in Java

I like 3: the most as it's easy to understand what it is.

stevengill: Kaz and I discussed this elsewhere and decided to use `enterpriseUrl` as the new name going forward.

docs/_packages/oauth.md
```diff
@@ -131,13 +131,17 @@ If you decide you need custom success or failure behaviors (ex: wanting to show
 
 ```javascript
 const callbackOptions = {
-  success: (installation, installOptions, req, res) => {
+  success: (installation, metadata, req, res) => {
     // Do custom success logic here
-    res.send('successful!');
+    // tip: you can add javascript and css in the htmlResponse using the <script> and <style> tags
+    const htmlResponse = `<html><body>Success!</body></html>`
+    res.writeHead(200, { 'Content-Type': 'text/html' });
```
seratch, line 138: In general, I recommend having charset here for i18n safety. I know this works without it but we can have it for developers that build their own pages containing multi byte characters.
```suggestion
    res.writeHead(200, { 'Content-Type': 'text/html; charset=utf-8' });
```

docs/_packages/oauth.md
```diff
@@ -165,15 +171,39 @@ const installer = new InstallProvider({
     // returns nothing
     storeInstallation: (installation) => {
```
seratch, line 172: nit: `async` has been missing. It's the same for `storeOrgInstallation`
```suggestion
    storeInstallation: async (installation) => {
```
https://github.com/slackapi/node-slack-sdk/blob/2d7a94c4f342fac94c7f6eb6be01b1257ce1df4e/packages/oauth/src/index.ts#L443

docs/_packages/oauth.md
```diff
@@ -182,12 +212,13 @@ const installer = new InstallProvider({
...
 
-You can use the the `installationProvider.authorize()` function to fetch data that has been saved in your installation store.
+You can use the the `installationProvider.authorize()` function to fetch data that has been saved in your installation store. For Org wide app installations, you can use `installationProvider.orgAuthorize()`
 
 ```javascript
 // installer.authorize takes in an installQuery as an argument
 // installQuery = {teamId: 'string', enterpriseId: 'string', userId: string, conversationId: 'string'};
-const result = installer.authorize({teamId:'my-Team-ID'});
+const result = installer.authorize({teamId:'my-team-ID'});
+const orgResult = installer.orgAuthorize({enterpriseId:'my-enterprise-ID'});
```
seratch, line 221:
```suggestion
const result = installer.authorize({teamId: 'my-team-ID'});
const orgResult = installer.orgAuthorize({enterpriseId: 'my-enterprise-ID'});
```

docs/_packages/oauth.md
```diff
@@ -203,13 +234,14 @@ result = {
...
-The `installer.authorize()` method only returns a subset of the installation data returned by the installation store. To fetch the entire saved installation, use the `installer.installationStore.fetchInstallation()` method. 
+The `installer.authorize()`/`installer.orgAuthorize()` methods only returns a subset of the installation data returned by the installation store. To fetch the entire saved installation, use the `installer.installationStore.fetchInstallation()`/`installer.installationStore.fetchOrgInstallation()` methods. 
 
 ```javascript
 // installer.installationStore.fetchInstallation takes in an installQuery as an argument
 // installQuery = {teamId: 'string', enterpriseId: 'string', userId: string, conversationId: 'string'};
 // returns an installation object
-const result = installer.installationStore.fetchInstallation({teamId:'my-Team-ID', enterpriseId:'my-enterprise-ID'});
+const result = installer.installationStore.fetchInstallation({teamId:'my-team-ID', enterpriseId:'my-enterprise-ID'});
+const orgResult = installer.installationStore.fetchOrgInstallation({enterpriseId:'my-enterprise-ID'});
```
seratch, line 244: As it returns `Promise`, is having `await` keyword expected here? https://github.com/slackapi/node-slack-sdk/blob/2d7a94c4f342fac94c7f6eb6be01b1257ce1df4e/packages/oauth/src/index.ts#L444
```suggestion
const result = await installer.installationStore.fetchInstallation({teamId: 'my-team-ID', enterpriseId: 'my-enterprise-ID'});
const orgResult = await installer.installationStore.fetchOrgInstallation({enterpriseId: 'my-enterprise-ID'});
```

packages/oauth/README.md
```diff
@@ -176,15 +178,39 @@ const installer = new InstallProvider({
     // returns nothing
     storeInstallation: (installation) => {
```
seratch, line 179: I don't repeat the same comments but there are the same missing `await` issues here.

packages/web-api/src/WebClient.ts
```diff
@@ -154,7 +161,10 @@ export class WebClient extends Methods {
     }
 
     const response = await this.makeRequest(method, Object.assign(
-      { token: this.token },
+      {
+        token: this.token,
+        team_id: this.teamId,
```
seratch, line 166: This means all web API calls have this `team_id` parameter in the case of org-level installations, right? My understanding is that the server-side just ignores the param(it doesn't return any errors/warnings about it) when it doesn't need it. On this assumption, this looks good to me.

seratch (approved):
I left some minor comments but overall it looks great to me.

packages/web-api/src/WebClient.ts
stevengill: yeah, we pretty much always pass `team_id` when making api calls if `team_id` is available in the `clientOptions`. server-side will ignore the param and won't give any errors or warnings.

PR #1056, opened 2020-07-16
Added three workflows.* methods for upcoming Workflow Steps feature

misscoded:
###  Summary

Added three `workflows` methods to web-api:

`workflows.stepCompleted`
`workflows.stepFailed`
`workflows.updateStep`

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -1420,11 +1420,11 @@ export interface WorkflowsStepCompletedArguments extends WebAPICallOptions, Toke
   workflow_step_execute_id: string;
   outputs?: object;
 }
- 
+
 export interface WorkflowsStepFailedArguments extends WebAPICallOptions, TokenOverridable {
   workflow_step_execute_id: string;
   error: {
-      message: string;
+    message: string;
   };
 }
 
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -380,6 +380,12 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
       set: bindApiCall<UsersProfileSetArguments, WebAPICallResult>(this, 'users.profile.set'),
     },
   };
+
+  public readonly workflows = {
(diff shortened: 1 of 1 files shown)


packages/web-api/src/methods.ts
```diff
@@ -1407,4 +1413,27 @@ export interface ViewsUpdateArguments extends WebAPICallOptions, TokenOverridabl
...
+   * `workflows.*`
+   */
+export interface WorkflowsStepCompletedArguments extends WebAPICallOptions, TokenOverridable {
+  workflow_step_execute_id: string;
+  outputs?: object;
+}
+ 
+export interface WorkflowsStepFailedArguments extends WebAPICallOptions, TokenOverridable {
+  workflow_step_execute_id: string;
+  error: object;
```
stevengill, line 1426: @misscoded We might be able to expand on the error object here now that we know it is required.

```suggestion
  error: {
      message: string;
  };
```

packages/web-api/src/methods.ts
```diff
@@ -1407,4 +1413,27 @@ export interface ViewsUpdateArguments extends WebAPICallOptions, TokenOverridabl
   hash?: string;
 }
 
+  /*
+   * `workflows.*`
+   */
+export interface WorkflowsStepCompletedArguments extends WebAPICallOptions, TokenOverridable {
+  workflow_step_execute_id: string;
+  outputs?: object;
```
stevengill, line 1421: This is user defined, so I think object type is fine for now for the beta but we might want to get more specific in the future if we can

packages/web-api/src/methods.ts
```diff
@@ -1407,4 +1413,27 @@ export interface ViewsUpdateArguments extends WebAPICallOptions, TokenOverridabl
...
+}
+ 
+export interface WorkflowsStepFailedArguments extends WebAPICallOptions, TokenOverridable {
+  workflow_step_execute_id: string;
+  error: object;
+}
+
+export interface WorkflowsUpdateStepArguments extends WebAPICallOptions, TokenOverridable {
+  workflow_step_edit_id: string;
+  inputs?: object;
```
stevengill, line 1431: This is also user defined, so I think object type is fine for now for the beta but we might want to get more specific in the future if we can

PR #1065, opened 2020-07-29
Convert types tests to use `tsd`

clavin:
###  Summary

Migrates the types tests to use [tsd](https://github.com/SamVerschueren/tsd#readme) instead of dtslint. The types tests were only testing `@slack/web-api` so I moved them closer to that package as well.


packages/web-api/test/types/webclient-paginate-types.test-d.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/* Testing the return type of WebClient#paginate() */
+
+expectType<AsyncIterable<WebAPICallResult>>(web.paginate('conversations.list'));
+
+expectType<AsyncIterable<WebAPICallResult>>(web.paginate('conversations.list', {}));
+
+expectType<Promise<void>>(web.paginate('conversations.list', {}, () => false));
+
+expectError(web.paginate('conversations.list', {}, () => 7));
```
seratch, line 16: nit: It may be also fine if we could make the validation here more specific by having `T` in `expectError<T>(value)`.

seratch:
see also: https://github.com/slackapi/bolt-js/pull/561

packages/web-api/test/types/webclient-paginate-types.test-d.ts
clavin: I think the two are subtly different as `expectError<T>` checks a value rather than function arguments. In all honesty, I don't know what the difference between `expectError<T>` and `expectNotAssignable<T>` is other than the error message.

That said, I don't see how `expectError<T>` would for this case?

seratch: Ah never mind! Somehow, I was just a bit confused at somewhere between type tests and runtime tests...

stevengill:
@clavin can you look into the failing tests here?

clavin:
So it looks like `tsd` is only supported for Node versions `>=10` 😭

stevengill:
@clavin want to rebase from master and see if the CI passes now?

PR #1075, opened 2020-08-12
oauth docs: correct nodejs `.url` usage

jakswa:
###  Summary

this is an [IncomingMessage](https://nodejs.org/docs/latest-v12.x/api/http.html#http_class_http_incomingmessage) object which gives the full URL for `.url`. We either want to switch to getting path, or check that we match the start of the URL. I just put what worked for me in dev/testing.


stevengill:
I just tested the original example and it worked fine for me. What OS are you using?

jakswa:
Thanks for testing. I was on Linux. can agree i confused the issue, if the example works for others. Feel free to reopen it.

PR #1078, opened 2020-08-17
Add response types to Web API methods

clavin:
###  Summary

This PR adds some rough types for Web API method responses to `@slack/web-api`. Rather than aim high for perfect coverage of the entire Web API's method responses, I've instead aimed a little lower to just recreating what is present on the [API Methods docs](https://api.slack.com/methods).

In its current stage, I genuinely expect there to be some really rough spots. There's a few `unknown`s left in where the API docs are too vague to write reasonable type immediately at a glance. I'm also expecting many `?` markers on fields to either be missing or unnecessarily present.

I'm not particularly against these being merged while being imperfect because updates to these types are, to me, a bugfix. That said, I also really would love eyes here to quickly catch what I've missed! An error doesn't have to make its way in if it can be caught beforehand. ☺️

### Requirements

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -44,15 +44,18 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
     apps: {
       approve: bindApiCall<AdminAppsApproveArguments, WebAPICallResult>(this, 'admin.apps.approve'),
       approved: {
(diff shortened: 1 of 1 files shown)


jonocairns:
I wonder if you just need to rebase off origin to get the build passing? Looks like the current version of ts in the build is 3.3 and the latest master is 4.1 (and should have Omit) @stevengill

seratch:
This is done by #1188 with a wider coverage. Thanks for taking the time to make this PR 🙇

Issue #1055, opened 2020-07-15
Add custom API URL support @slack/oauth


karenfann:
### Description

Custom API URL via `slackApiUrl` option is supported in @slack/web-api but not in @slack/oauth

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

stevengill:
Thanks @karenfann!

Relevant custom api url docs: https://slack.dev/node-slack-sdk/web-api#custom-api-url

@seratch mentioned one way to handle this would be to allow developers to pass in their own instance `WebClient` via method args. Also, `getBotId` can reuse the instance. 

`getBotId` needs to the token that was issued during the OAuth flow, so this [initialize without a token section](https://slack.dev/node-slack-sdk/web-api#initialize-the-client) seems relevant. 

We also need to update the hardcoded url in `generateInstallUrl`. So maybe accepting `slackApiUrl` would be the easiest

stevengill:
This has been merged and will be released later today!

PR #1081, opened 2020-08-19
added clientOptions and authorizationUrl

stevengill:
###  Summary

Added `clientOptions` and `authorizationUrl` as options that `@slack/oauth` accepts. `clientOptions` is passed along to `@slack/web-api`. `authorizationUrl` is used to override the authorization endpoint this library uses (default is `https://slack.com/oauth/v2/authorize`

Also added support for a `npm run watch` command that uses `nodemon` to watch for changes and rebuild the source. Great for developing on the package itself. 

Issue: #1055 

### DIFF
--- a/packages/oauth/src/index.spec.js
+++ b/packages/oauth/src/index.spec.js
@@ -218,6 +218,43 @@ describe('OAuth', async () => {
               assert.fail(error.message);
           }
       });
+      it('should return a generated url when passed a custom authorizationUrl', async () => {
+        const fakeStateStore = {
+          generateStateParam: sinon.fake.resolves('fakeState'),
+          verifyStateParam: sinon.fake.resolves({})
+        }
+        const authorizationUrl = 'https://dev.slack.com/oauth/v2/authorize';
+        const installer = new InstallProvider({ clientId, clientSecret, stateStore: fakeStateStore, authorizationUrl });
+        const scopes = ['channels:read'];
+        const teamId = '1234Team';
+        const redirectUri = 'https://mysite.com/slack/redirect';
+        const userScopes = ['chat:write:user']
+        const installUrlOptions = {
+          scopes,
+          metadata: 'some_metadata',
+          teamId,
+          redirectUri,
+          userScopes,
+        }
+        try {
+            const generatedUrl = await installer.generateInstallUrl(installUrlOptions)
+            assert.exists(generatedUrl);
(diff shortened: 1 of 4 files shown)


packages/oauth/src/index.spec.js
```diff
@@ -216,6 +218,43 @@ describe('OAuth', async () => {
               assert.fail(error.message);
           }
       });
+      it('should return a generated url when passed a custom authorizationUrl', async () => {
```
aoberoi, line 221: does this test also cover the new `clientOptions` option for `InstallProvider`?

stevengill: No :(. I wanted to but I think I need to change my instrumentation around how I rewire `Webclient`

packages/oauth/src/index.ts
```diff
@@ -71,6 +75,18 @@ export class InstallProvider {
     this.handleCallback = this.handleCallback.bind(this);
     this.authorize = this.authorize.bind(this);
     this.authVersion = authVersion;
+
+    this.authorizationUrl = authorizationUrl;
+    if (authorizationUrl !== 'https://slack.com/oauth/v2/authorize' && authVersion === 'v1') {
```
misscoded, line 80: This code is still new to me, but is there never a case where a user might (accidentally?) pass in `https://slack.com/oauth/authorize` and `v2`? If so, does that work without issue?

stevengill: Ooo good point! We don't do any conditional checks for `v2`, so in that scenario you describe, the `authorizationUrl` of `https://slack.com/oauth/authorize` will take priority and it should work as expected!

aoberoi: i think this is a reasonable approach. the risk is that the developer sets the `authorizationUrl` to the `v1` value, so the response we get from the platform is in the v1 shape, and then when we parse it later it could throw or produce some unexpected result since `authVersion: 'v2'` and we expect it in that shape.

but i think if you're setting `authorizationUrl` on your own, you're asking to get down into the guts of how it all works. with great power come great responsibility, right?

packages/oauth/src/index.spec.js
aoberoi: seems pretty low risk to me, but can we create an issue (labeled `tests`) to come back to this? it might even be a `good first issue` for someone.

Issue #561, opened 2018-05-15
Preserve stack track of original errors


bkeepers:
`WebClient.ts` has several methods for wrapping errors from the client:

https://github.com/slackapi/node-slack-sdk/blob/master/src/WebClient.ts#L678-L718

All of them create a `new Error`, which means the stack trace that is returned when these errors occur is not very helpful to users of the SDK. For example, here's the error I'm currently trying to debug:

```
13:36:03.418Z ERROR event: A request error occurred: undefined
  Error: A request error occurred: undefined
      at requestErrorWithOriginal (/Users/bkeepers/projects/slack-app/node_modules/@slack/client/src/WebClient.ts:681:5)
      at got.post.catch (/Users/bkeepers/projects/slack-app/node_modules/@slack/client/src/WebClient.ts:145:21)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

[This stackoverflow discusison](https://stackoverflow.com/questions/42754270/re-throwing-exception-in-nodejs-and-not-losing-stack-trace) has some options for preserving `error.stack` when wrapping errors.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [x] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

#### Reproducible in:

`@slack/client` version: 4.0.1

node version: 8.9.4

OS version(s): macOS 10.13.4

aoberoi:
my first thought is "why is the error message `undefined`, that should never happen!" i think that's probably worth looking into separately, so if you have any more information regarding how i might be able to reproduce this specific error, i'm very interested. i'll try to trace the possible code paths myself, but i'm not sure i have enough info.

[...]

PR #1086, opened 2020-08-24
Append original stack when WebClient#apiCall throws

vaskevich:
###  Summary

When an API call within `@slack/web-api` fails, the resulting error's stack loses the call context:

```
Error: An API error occurred: account_inactive
     at Object.platformErrorFromResult (/project/node_modules/@slack/web-api/src/errors.ts:94:5)
     at WebClient.apiCall (/project/node_modules/@slack/web-api/src/WebClient.ts:188:13)
     at runMicrotasks (<anonymous>)
     at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

This PR appends the original stack leading up to the call so that developers can more easily trace failing calls.

For more details, see https://github.com/slackapi/node-slack-sdk/issues/561#issuecomment-679265974

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -145,50 +145,57 @@ export class WebClient extends Methods {
    * @param options - options
    */
   public async apiCall(method: string, options?: WebAPICallOptions): Promise<WebAPICallResult> {
-    this.logger.debug(`apiCall('${method}') start`);
+    const {stack: originalStack} = new Error('thrown by');
 
-    warnDeprecations(method, this.logger);
+    try {
+      this.logger.debug(`apiCall('${method}') start`);
 
-    if (typeof options === 'string' || typeof options === 'number' || typeof options === 'boolean') {
-      throw new TypeError(`Expected an options argument but instead received a ${typeof options}`);
-    }
+      warnDeprecations(method, this.logger);
 
-    const response = await this.makeRequest(method, Object.assign(
-      { token: this.token },
-      options,
-    ));
-    const result = this.buildResult(response);
+      if (typeof options === 'string' || typeof options === 'number' || typeof options === 'boolean') {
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -145,50 +145,57 @@ export class WebClient extends Methods {
...
-      throw platformErrorFromResult(result as (WebAPICallResult & { error: string; }));
-    }
+      if (!result.ok) {
+        throw platformErrorFromResult(result as (WebAPICallResult & { error: string; }));
+      }
 
-    return result;
+      return result;
+    } catch (err) {
+      err.stack += `\n${originalStack}`;
```
vaskevich, line 196: This currently looks as follows:

```
Error: An API error occurred: account_inactive
     at Object.platformErrorFromResult (/project/node_modules/@slack/web-api/src/errors.ts:94:5)
     at WebClient.apiCall (/project/node_modules/@slack/web-api/src/WebClient.ts:188:13)
     at runMicrotasks (<anonymous>)
     at processTicksAndRejections (internal/process/task_queues.js:97:5)
Error: thrown by
    at WebClient.apiCall (/project/node_modules/@slack/web-api/dist/WebClient.js:120:40)
    at /project/my_file.ts:288:19
```

Happy to tweak as needed.

stevengill:
Hey @vaskevich!

Thanks for sending this in. 

In general, I agree that including the proper stack trace in valuable. 

I believe the solution you proposed though would have issues with debuggers such as vscode. It is also a little expensive to create a new error even when you don't plan on throwing the error every time the method is called.

Do you think there might be a different solution to this problem? Is Axios preventing us from taking advantage of node v8's async stack traces?

jaywhiletwo:
This would really improve the developer experience. Currently, errors thrown by the web client are only referencing the stack as they know it when the promise is being resolved. This means we often cannot determine the source of the error without painstaking debugging and isolation steps.

PR #1097, opened 2020-09-23
feat: upgrade p-queue (no changes)

FauxFaux:
###  Summary

`p-queue` now provides its own types, and `@types/p-queue` throws a deprecation warning,
so remove it. (I only care about the deprecation warning.)

### DIFF
--- a/packages/web-api/package.json
+++ b/packages/web-api/package.json
@@ -49,12 +49,11 @@
     "@slack/types": "^1.7.0",
     "@types/is-stream": "^1.1.0",
     "@types/node": ">=8.9.0",
-    "@types/p-queue": "^2.3.2",
     "axios": "^0.19.0",
     "eventemitter3": "^3.1.0",
     "form-data": "^2.5.0",
     "is-stream": "^1.1.0",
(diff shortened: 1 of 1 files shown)


stevengill (approved):
LGTM. I reviewed the breaking changes on pqueue since the previous major we used. Nothing breaks on our end to upgrade. 

Thanks for sending this in!

aoberoi:
@stevengill how do we verify changes like [`v6.5.0`](https://github.com/sindresorhus/p-queue/releases/tag/v6.5.0) don't break anything? specifically, that release updates TypeScript to v3.9, which is above our projects' minimum TypeScript version of v3.3.

i recall that in the past situations where dependencies (especially those that bundle their own type definitions instead of retrieving them from DT) start to use new TypeScript syntax, that effectively bumps our project's minimum TypeScript version, too. but i'm not sure if that was all cases, or only in cases where some part of our project's API was dependent on a type from the dependencies API.

i think, before the changes in #831 for the `sdk-integration-tests`, we actually had CI verify the compatibility with our oldest supported TypeScript version. i'm not sure the refactored version provides that verification anymore.

Issue #1108, opened 2020-10-20
Add dispatch_action_config to PlainTextInput


seratch:
### Description

`dispatch_action_config` in `PlainTextInput` is still missing.
https://github.com/stevengill/node-slack-sdk/blob/%40slack/types%401.8.0/packages/types/src/index.ts#L235-L242

References:
* https://api.slack.com/reference/block-kit/block-elements#input
* https://api.slack.com/reference/block-kit/composition-objects#dispatch_action_config
* https://github.com/slackapi/java-slack-sdk/pull/578
* https://github.com/slackapi/node-slack-sdk/pull/1102

#### Packages:

- [x] `@slack/types`

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

seratch:
Fixed by #1109

PR #1109, opened 2020-10-20
added dispatch_action_config to PlainTextInput

stevengill:
###  Summary

Fixes issue #1108 

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -241,6 +241,11 @@ export interface PlainTextInput extends Action {
   multiline?: boolean;
   min_length?: number;
   max_length?: number;
+  dispatch_action_config?: DispatchActionConfig;
+}
+
+export interface DispatchActionConfig {
+  trigger_actions_on?: ('on_enter_pressed' | 'on_character_entered')[];
 }
 
 /*


packages/types/src/index.ts
```diff
@@ -241,6 +241,11 @@ export interface PlainTextInput extends Action {
   multiline?: boolean;
   min_length?: number;
   max_length?: number;
+  dispatch_action_config?: DispatchActionConfig;
+}
+
+export interface DispatchActionConfig {
+  trigger_actions_on?: ('on_enter_pressed' | 'on_character_entered')[];
```
stevengill, line 248: I did this instead of `string[]` as it is more accurate.

packages/types/src/index.ts
```diff
@@ -241,6 +241,11 @@ export interface PlainTextInput extends Action {
   multiline?: boolean;
   min_length?: number;
   max_length?: number;
+  dispatch_action_config?: DispatchActionConfig;
```
stevengill, line 244: https://api.slack.com/reference/block-kit/block-elements#input is missing if `dispatch_action_config` is required or not. I'm assuming it isn't required.

mwbrooks: Good catch. We should probably report this to the api.slack.com team as well

mwbrooks (approved):

packages/types/src/index.ts
stevengill: Yup! I reported it to them!

seratch (approved):

packages/types/src/index.ts
seratch: >I'm assuming it isn't required.

This field is surely optional.

PR #1116, opened 2020-11-16
Update ViewsUpdateArguments to accommodate external_id and view_id

misscoded:
###  Summary

Fixes Issue #1002  

Current `ViewsUpdateArguments` interface used by `views.update()` does not accept `external_id` to be used as an alternative to `view_id`. This PR adds two additional interfaces and a new type that allows the use of either or both ids.

Previous fix was merged and reverted (#1003), but conversation on that thread guided the fix presented here.

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -76,11 +76,11 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
       rename: bindApiCall<AdminConversationsRenameArguments, WebAPICallResult>(this, 'admin.conversations.rename'),
       restrictAccess: {
         addGroup: bindApiCall<AdminConversationsRestrictAccessAddGroupArguments, WebAPICallResult>(
-            this, 'admin.conversations.restrictAccess.addGroup'),
+          this, 'admin.conversations.restrictAccess.addGroup'),
         listGroups: bindApiCall<AdminConversationsRestrictAccessListGroupsArguments, WebAPICallResult>(
-            this, 'admin.conversations.restrictAccess.listGroups'),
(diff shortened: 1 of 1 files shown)


seratch:
How about adding some tests here? https://github.com/slackapi/node-slack-sdk/tree/main/packages/web-api/test/types

misscoded:
> How about adding some tests here? https://github.com/slackapi/node-slack-sdk/tree/main/packages/web-api/test/types

@seratch I was waiting to have a conversation with Ankur about the proposed implementation before deciding the final state, but the decision from that conversation -- now pushed/present -- shouldn't require any additional tests (I don't believe).

seratch:
I see, y'all decided not to have the constraints to check if required field is present. It's also an option and is better than the current state. By the way, you can have another commit for formatting the source code (this one is OK, next time!). It makes reviews easier.

seratch (approved):
@aoberoi as you suggested this approach, can you approve this PR?

aoberoi (approved):

aoberoi:
Approved! And yes I agree that if we wanted to test the variance of using `view_id` without `external_id` and using `external_id` without `view_id`, then some type tests would be desired. But in this case, we're maintaining the status quo of making those fields optional and having a relatively shallow type model for the required-ness of the arguments. In this shallow model, I don't think we need any tests to prove this works.

seratch:
@misscoded 
Probably, we can safely merge this PR. But, just in case, can you confirm if it's okay to ignore this? https://travis-ci.org/github/slackapi/node-slack-sdk/builds/744279186

misscoded:
Confirmed. Per @aoberoi's rundown of the issue we're having with Codecov, we'll override this status check and look to fixing the issue in the meantime.

PR #1121, opened 2020-11-20
Typescript: Adds the optional OAuth2 state type to InstallURLOptions interface

darwin403:
###  Summary

Adds the missing type of OAuth2 `state` property to `InstallURLOptions` while generating the URL using the `generateInstallUrl` method.

### DIFF
--- a/packages/oauth/src/index.ts
+++ b/packages/oauth/src/index.ts
@@ -328,6 +328,7 @@ export interface InstallURLOptions {
   redirectUri?: string;
   userScopes?: string | string[]; // cannot be used with authVersion=v1
   metadata?: string; // Arbitrary data can be stored here, potentially to save app state or use for custom redirect
+  state?: string;
 }
 
 export interface CallbackOptions {


seratch:
Hi @darwin403 thanks for taking the time to make this pull request! But the `state` parameter is not missing. You can use your own `stateStore` to append the `state` parameter in the authorize URL: https://github.com/slackapi/node-slack-sdk/blob/%40slack/oauth%401.3.0/packages/oauth/src/index.ts#L138-L140

aoberoi:
@darwin403 the `metadata` property of the options passed to `generateInstallUrl()` can be used to insert your own data into `state`. the library will add that metadata into the `state` param, and once the flow has completed give you access to that `metadata` inside both the `success` and `failure` callbacks.

darwin403:
Hi @seratch! Perhaps, I'm misinformed. My understanding was that a typical OAuth2 protocol library allows for the `state` parameter in the SDK. In fact, we are currently using something like the following and it works, that is, it encrypts the `state` parameter as shown below.

```
 const url = await installer.generateInstallUrl({
      scopes: ["incoming-webhook"],
      state: campaignId,
      redirectUri: process.env.SLACK_OAUTH2_REDIRECT_URI,
    }
```

Therefore, I was under the impression that since the library is not ignoring the `state: campaignId` field. It only made sense to include the type as well. Not sure if I'm missing out on something here.

seratch:
@darwin403 Thanks for your response! As @aoberoi already mentioned, you can use `metadata` for the use case. https://github.com/slackapi/node-slack-sdk/blob/2d7a94c4f342fac94c7f6eb6be01b1257ce1df4e/packages/oauth/src/index.ts#L330

It seems everything is now clear. Would you mind if I close this?

stevengill:
@darwin403 just to add to this, the library under the hood will take all options passed into `generateInstallUrl()` and pass them into the `state`. As @aoberoi and @seratch mentioned, `metadata` is the property you want to store data in. By passing in `state`, it will be packaged up into our generated `state` value and then returned. So it currently works, but could change in a future update as we don't officially support it.

PR #1168, opened 2021-02-03
Standardize fetchInstallation to use installQuery as argument

misscoded:
###  Summary

Currently, there is a lack of standard arg usage with `fetchInstallation`. 

This PR aligns all instances to use `installQuery` instead of `InstallQuery` (which suggests a class is being passed) and `query` (which is used once and is the odd instance out). To note, `installQuery` is already referenced across the package in various forms of documentation.

### DIFF
--- a/docs/_tutorials/migrating_to_v6.md
+++ b/docs/_tutorials/migrating_to_v6.md
@@ -41,19 +41,19 @@ installationStore: {
       // change the line below so it saves to your database
       return await database.set(installation.team.id, installation);
     },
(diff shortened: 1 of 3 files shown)


seratch (approved):
Looks great to me!

seratch:
Thanks for fixing this! Could you fix bolt-js documents too? https://github.com/slackapi/bolt-js/search?q=InstallQuery

stevengill (approved):

PR #1171, opened 2021-02-05
Fix type for tlsConfig in WebClient.ts

KhushrajRathod:
###  Summary

Fixes type for tlsConfig in WebClient.ts as per line 109

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -61,7 +61,7 @@ export class WebClient extends Methods {
   /**
    * Configuration for custom TLS handling
    */
-  private tlsConfig: TLSOptions;
+  private tlsConfig: TLSOptions | {};
 
   /**
    * Preference for immediately rejecting API calls which result in a rate-limited response


seratch:
What is the use case where `TLSOptions` does not work for you?

KhushrajRathod:
It does -- this doesn't directly cause a problem, its just that I was porting this over to Deno and I realized the type was incorrect so I thought it'd be good to fix it

seratch:
Thanks for sharing the context. As of today, this project is not planning to support Deno in the short term. Thus, we are unable to merge this change for the reason.

KhushrajRathod:
@seratch This isn't a deno-specific thing, its just that I found it _because_ I was porting it to deno. The Type for tlsConfig should be `TLSOptions | {}` since it can be either `TLSOptions` or `{}` as seen on line 109

seratch:
As the code currently compiles, `{}` is compatible with the type `TLSOptions` (specifically `SecureContextOptions`) in Node.

KhushrajRathod:
Ah, okay, cool, thanks for clarifying

Issue #1172, opened 2021-02-10
type definition of ChatPostMessageArguments has incorrect property


kentac55:
### Description

Describe your issue here.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Type definition of [ChatPostMessageArguments](https://github.com/slackapi/node-slack-sdk/blob/a90787ea80185d198e7b508e20ce69ccefc34033/packages/web-api/src/methods.ts#L964) has `message` as required property but in [slack api doc](https://api.slack.com/methods/chat.postMessage#arg_text),  as optional property.

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/interactive-messages`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

#### Reproducible in:

package version: 6.0.0

node version: 14.15.4

OS version(s): Linux 5.10.14-arch1-1 SMP PREEMPT Sun, 07 Feb 2021 22:42:17 +0000 x86_64 GNU/Linux

#### Steps to reproduce:

1. write code as below

    ```typescript
    webClient.chat.postMessage({
      channel: 'C0BQ8GW78',
      blocks: [
        {
          type: 'section',
          text: {
            type: 'mrkdwn',
            text:
              "Hello, Assistant to the Regional Manager Dwight! *Michael Scott* wants to know where you'd like to take the Paper Company investors to dinner tonight.\n\n *Please select a restaurant:*",
          },
        },
      ],
    })
    ```

2. run tsc
3. got error

    ```
[...]

PR #1173, opened 2021-02-10
fix type definition of ChatPostMessageArguments

kentac55:
###  Summary

Made `text` property of `ChatPostMessageArguments` optional.

close: #1172

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -155,7 +155,7 @@ export class WebClient extends Methods {
     this.logger.debug(`apiCall('${method}') start`);
 
     warnDeprecations(method, this.logger);
-    warnMissingTextArgument(method, this.logger, options);
+    warnIfFallbackIsMissing(method, this.logger, options);
 
     if (typeof options === 'string' || typeof options === 'number' || typeof options === 'boolean') {
       throw new TypeError(`Expected an options argument but instead received a ${typeof options}`);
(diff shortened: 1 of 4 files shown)


kentac55:
Note: 

If we do `postMessage` without specifying either `text` or `blocks` attribute, there will be no problem with the type check, but the API will return `no_text` error.

To prevent this, we should use generic type like [that](https://github.com/slackapi/node-slack-sdk/issues/496#issuecomment-373735256) I think.

seratch:
Thanks a lot for taking the time to make this pull request 🙇 Can you check my comment in the issue? https://github.com/slackapi/node-slack-sdk/issues/1172#issuecomment-779559834

kentac55:
@seratch added warning logic :hugs: 

Can you check this?

packages/web-api/src/WebClient.ts
```diff
@@ -613,3 +614,21 @@ function warnDeprecations(method: string, logger: Logger): void {
...
   }
 }
+
+/**
+ * Log a warning when using chat.postMessage without text argument
+ * @param method api method being called
+ * @param logger instance of we clients logger
+ */
+function warnMissingTextArgument(method: string, logger: Logger, options?: WebAPICallOptions): void {
+  const methodsWithOptionalText = ['chat.postMessage'];
```
seratch, line 624: Can you add other chat.* methods here like I implemented in Python SDK? https://github.com/slackapi/python-slack-sdk/pull/910/files

seratch (changes requested):
Thanks for updating! Please check my comment.

packages/web-api/src/WebClient.ts
kentac55: Thank you for revieing!

added warning logic for other methods and updated test :smile:

kentac55:
@seratch Sorry for late response. I forgot to mention you 🙇 

Can you check the new changes?

packages/web-api/src/WebClient.ts
```diff
@@ -613,3 +614,25 @@ function warnDeprecations(method: string, logger: Logger): void {
...
+/**
+ * Log a warning when using chat.postMessage without text argument
+ * @param method api method being called
+ * @param logger instance of we clients logger
+ */
+function warnMissingTextArgument(method: string, logger: Logger, options?: WebAPICallOptions): void {
+  const methodsWithOptionalText = ['chat.postEphemeral', 'chat.postMessage', 'chat.scheduleMessage', 'chat.update'];
+  const isNeededToCheckMethod = methodsWithOptionalText.includes(method);
+
+  const isEmptyText = (args: WebAPICallOptions) =>
```
seratch, line 627: Sorry, the implementation I mentioned had lack of consideration for `attachments`. Can you update this logic to check `fallback` in the case of `attachments`? Refer to https://github.com/slackapi/python-slack-sdk/pull/966 for the logic.

seratch (changes requested):

seratch:
@kentac55 Thanks for being patient here and updating the code for the attachments pattern! If you don't mind, could you rename the local variables like this? https://github.com/seratch/node-slack-sdk/commit/b963985c52c0598f2091667620b76d010255e5cc

kentac55:
@seratch Sorry for late response :bow: 

cherry-picked your commit :smiley:

seratch (approved):
Thanks for working on this! The changes look great to me but I would like to have other maintainers' eyes before merging it.

seratch:
OK, there is no concerns about this. Let me merge this PR. Thanks for your contribution! @kentac55

Issue #1180, opened 2021-02-25
promiseTimeout keeps Node process waiting even if the input promise resolves faster than timeout


oshchyhol:
### Description

[Current implementation of promise timeout in the package](https://github.com/slackapi/node-slack-sdk/blob/cc32ba7a4ee8b21cfe553d2c490a700110e118e0/packages/interactive-messages/src/util.ts#L17) doesn't clear the timer upon successful input promise resolution, so Node process keeps waiting for a scheduled timer.
I discovered this when writing an integration test for a Slack action handler using Jest. Test runner would keep waiting on scheduled timer for a default timeout duration of 2.5 sec and issue a warning.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [ ] `@slack/events-api`
- [x] `@slack/interactive-messages`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

#### Reproducible in:

package version: 2.0.0

node version: 14.5.4

OS version(s): Mac OS 11.2.1

#### Steps to reproduce:

1. Run any script in Node that uses `util.promiseTimeout` implementation from `@slack/interactive-messages` with a long enough timeout and immediately resolving promise.

E.g. use this script or clone and run the attached sample project
```
mkdir test-promise-timeout && cd test-promise-timeout
npm install @slack/interactive-messages
node -e "const { promiseTimeout } = require('@slack/interactive-messages/dist/util'); promiseTimeout(5000, Promise.resolve(42).then(() => console.log('Stealing 5 sec of your life :)')))"
```

#### Expected result:

Node process exits immediately

#### Actual result:

Node process keeps waiting for the timeout to pass.

[...]

PR #1181, opened 2021-02-25
Clear timer if input promise resolves before time runs out

oshchyhol:
###  Summary

Issue: https://github.com/slackapi/node-slack-sdk/issues/1180
Fixes a bug in the [implementation of promiseTimeout](https://github.com/slackapi/node-slack-sdk/blob/cc32ba7a4ee8b21cfe553d2c490a700110e118e0/packages/interactive-messages/src/util.ts#L17) where an uncleared timer would keep Node process waiting before exiting even if the input promise resolves faster than the timeout.
I discovered this when writing an integration test for a Slack action handler using Jest. Test runner would keep waiting on scheduled timer for a default timeout duration of 2.5 sec and issue a warning.
The solution is to clear the timeout on `Promise.race` resolution.

### DIFF
--- a/packages/interactive-messages/src/util.spec.js
+++ b/packages/interactive-messages/src/util.spec.js
@@ -2,6 +2,7 @@ require('mocha');
 const { assert } = require('chai');
 const { promiseTimeout, errorCodes } = require('./util');
 const { delayed } = require('../test/helpers');
+const sinon = require('sinon');
 
 // test suite
 describe('promiseTimeout', function () {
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for improving this! Although [this package is already deprecated](https://slack.dev/node-slack-sdk/tutorials/migrating-to-v6#deprecating-slackevents-api--slackinteractive-messages), I personally think this improvement should be merged.

seratch (approved):

mwbrooks (approved):
Thanks a lot for this improvement @oshchyhol! I agree with @seratch that we should definitely apply this patch to the next release.

Tests look great and pass for me! 🚀

PR #1188, opened 2021-03-11
Add API response types to web-api methods

seratch:
###  Summary

This pull request adds types to `@slack/web-api` API method call results. **It does not bring any breaking changes**. The generated types are all subtypes of `WebAPICallResult` interface.

Although the pull request #1078 applies similar changes, the changes were done by hand and the coverage is not complete. The approach in this PR utilizes [the official Java SDK's integration test results](https://github.com/slackapi/java-slack-sdk) for automatic code generation. Thus, we don't need to manually update the interface. Also, 100% of the API methods are covered.

We can assume the already generated fields are accurate as they surely exist in the actual API responses. If the Java SDK's tests do not cover some patterns, the fields that may be included only with a particular condition can be missing.

You can update the type definitions by running the following command at any time. The command checks out the Java SDK git repository and run [quicktype](https://quicktype.io/) code generator to update the TypeScript source code under `packages/web-api/src/response/` package. 

```bash
./scripts/generate-web-api-types.sh
```

We can add the execution to the CI builds but manually running it would be enough so far.

Lastly, here are pros/cons of this pull request's approach. I'm sure it's reasonable enough to go with this way though.

#### Pros

* 100% automatic code generation - zero manual updates for maintaining existing types
* zero breaking changes to the existing apps
* developers still can call missing fields, thanks to `[key: string]: unknown;` in `WebAPICallResult`
* benefits for both TS and JS (+VS Code) users

#### Cons

* dependency to the slack-java-sdk project
* hundreds of new files
* depends on quicktype

If we have to stop using quicktype in the future, we can just give up automatic generation. We are still able to manually maintain the types afterward.


packages/web-api/src/WebClient.spec.js
```diff
@@ -1,39 +1,38 @@
...
-const Busboy = require('busboy');
-const sinon = require('sinon');
-
-const token = 'xoxb-faketoken';
-
-describe('WebClient', function () {
-
-  describe('constructor()', function () {
-    it('should build a default client given a token', function () {
+require("mocha");
```
seratch, line 1: I just applied prettier to all the files in web-api project

packages/web-api/src/methods.ts
```diff
@@ -1,17 +1,206 @@
...
+  CallUser,
+} from "@slack/types";
+import {
+  WebAPICallOptions,
+  WebAPICallResult,
+  WebClient,
+  WebClientEvent,
+} from "./WebClient";
+import { EventEmitter } from "eventemitter3";
+import { AdminAppsApproveResponse } from "./response/AdminAppsApproveResponse";
```
seratch, line 18: These files under `./response` are auto generated.

packages/web-api/src/methods.ts
```diff
@@ -33,412 +222,987 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
...
-  public abstract apiCall(method: string, options?: WebAPICallOptions): Promise<WebAPICallResult>;
+  public abstract apiCall(
+    method: string,
+    options?: WebAPICallOptions
+  ): Promise<WebAPICallResult>;
 
   public readonly admin = {
     apps: {
-      approve: bindApiCall<AdminAppsApproveArguments, WebAPICallResult>(this, 'admin.apps.approve'),
+      approve: bindApiCall<AdminAppsApproveArguments, AdminAppsApproveResponse>(
```
seratch, line 242: `AdminAppsApproveResponse` is a subtype of `WebAPICallResult`

packages/web-api/src/response/ChatPostMessageResponse.ts
```diff
@@ -0,0 +1,363 @@
+import { WebAPICallResult } from "../WebClient";
+export type ChatPostMessageResponse = WebAPICallResult & {
```
seratch, line 2: Probably, most of you are already familiar with this type 😄

seratch (commented):
comments for reviewers

packages/web-api/src/WebClient.spec.js
seratch: oh, lint fails. i will update the whole changes.

seratch: reverted

seratch:
codecov does not return the results but the builds passed!

packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts
```diff
@@ -0,0 +1,8 @@
+/* tslint:disable */
```
seratch, line 1: Some parts of auto-generated code are not compatible with the current tslint settings. I tried to settle all of them [this way](https://github.com/slackapi/node-slack-sdk/blob/94eb35ad1361a52f65ac43a289b99e703810827f/scripts/code_generator.rb#L24-L26) but a few patterns are inevitable. You can see the result by modifying the code generator.

seratch:
Let me add a few more comments for some questions about this:

Q: What's the dependency to the Java SDK project?
To be clear, this code generation depends on only JSON data produced by the integration tests in the Java SDK. There is no dependency to Java runtime and Java SDK code. These JSON files are automatically generated from the actual JSON response from our server-side. The test utilities in the project save and normalize the data in [this directory](https://github.com/slackapi/java-slack-sdk/tree/main/json-logs/samples/api). If we do the same in the Node project as part of integration tests, we can do this only in the node-slack-sdk project. But it’s obviously duplicated work.

Q: Why are all fields optional?
Being required or optional is not easily deterministic for machine at this point. The code generator uses normalized JSON data as source of truth and there is no metadata about optional vs non-optional. Also, it can often be complex. Some fields can be basically optional while it may be non-optional under a particular condition. It may be possible to try to implement custom rules but we are unable to use existing tools like quicktype for it. We have to invent our own tool or manually maintain types. In my opinion, these ways are not sustainable.

mwbrooks (approved):
I like this approach. Well done @seratch!

While auto-generated types may not be as rich as handwritten, it's better than nothing and easier to maintain.

The dependency on the Java SDK is not ideal, but okay in my mind. If it's an issue, we could create separate project to generate the types and send a PR into this project.

@seratch is it possible to switch the Ruby `quicktype` for [Node.js dev dependency](https://github.com/quicktype/quicktype#installation)?

```zsh
npm install --save-dev quicktype
````

seratch:
@mwbrooks Thanks for the suggestion - I've resolved it 👍

TonalidadeHidrica:
I'm very excited to use the fully-typed version of API client, but it's not released yet.  The issues in 6.2 milestone has all been closed or merge, so isn't it already possible to release the new version?

seratch:
@TonalidadeHidrica Thanks for your interest! Yes, the last task in the list was completed yesterday. We are going to release a new version soon 🚀

PR #1208, opened 2021-04-13
Resolve inconsistent package dependencies across slack packages

seratch:
###  Summary

This pull request resolves the build settings of `@slack/web-api` and `@slack/oauth` packages. The major issue this pull request is resolving is that only `@slack/web-api`'s requires older major version of `@slack/logger`.

### DIFF
--- a/packages/oauth/package.json
+++ b/packages/oauth/package.json
@@ -38,8 +38,8 @@
     "watch": "npx nodemon --watch 'src' --ext 'ts' --exec npm run build"
   },
   "dependencies": {
-    "@slack/logger": "^2.0.0",
-    "@slack/web-api": "^5.7.0",
+    "@slack/logger": "^3.0.0",
+    "@slack/web-api": "^6.0.0",
     "@types/jsonwebtoken": "^8.3.7",
     "@types/node": ">=12",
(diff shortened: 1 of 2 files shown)


packages/web-api/package.json
```diff
@@ -45,8 +45,8 @@
     "watch": "npx nodemon --watch 'src' --ext 'ts' --exec npm run build"
   },
   "dependencies": {
-    "@slack/logger": ">=1.0.0 <3.0.0",
```
seratch, line 48: at least we should update this

packages/oauth/package.json
```diff
@@ -38,8 +38,8 @@
     "watch": "npx nodemon --watch 'src' --ext 'ts' --exec npm run build"
   },
   "dependencies": {
-    "@slack/logger": "^2.0.0",
```
seratch, line 41: these are optional - I'm happy to revert these for now if others think we should do so.

mwbrooks: I think it's a great time to bump everything, so go for it!

mwbrooks (approved):

seratch:
@mwbrooks Thanks for checking this!

Issue #1196, opened 2021-03-25
Warn logging if thread_ts in chat.postMessage arguments is a float


seratch:
This is a nice-to-have feature but we may want to have this for better developer experience.

---
As mentioned above, any `ts` value in message objects needs to be a string.

When you write apps in TypeScript, the TS compiler accepts only string https://github.com/slackapi/node-slack-sdk/blob/%40slack/web-api%406.1.0/packages/web-api/src/methods.ts#L974 but the validation does not work in JavaScript. 

We may want to have warn logging to let developers know the necessity to pass a string value. I will create a new task issue for it. Let me close this issue then.

_Originally posted by @seratch in https://github.com/slackapi/node-slack-sdk/issues/780#issuecomment-806263857_

PR #1223, opened 2021-05-04
Warn logging if thread_ts is not a string

hi1280:
###  Summary

Warn logging if thread_ts argument of chat.postMessage, chat.postEphemeral, chat.scheduleMessage and files.upload is not a string.

fix: #1196 

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -620,6 +620,7 @@ function warnDeprecations(method: string, logger: Logger): void {
  * Log a warning when using chat.postMessage without text argument or attachments with fallback argument
  * @param method api method being called
  * @param logger instance of we clients logger
+ * @param options arguments for the Web API method
  */
 function warnIfFallbackIsMissing(method: string, logger: Logger, options?: WebAPICallOptions): void {
   const targetMethods = ['chat.postEphemeral', 'chat.postMessage', 'chat.scheduleMessage', 'chat.update'];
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -645,3 +647,18 @@ function warnIfFallbackIsMissing(method: string, logger: Logger, options?: WebAP
...
+ * @param method api method being called
+ * @param logger instance of web clients logger
+ * @param options arguments for the Web API method
+ */
+function warnIfThreadTsIsNotString(method: string, logger: Logger, options?: WebAPICallOptions): void {
+  const targetMethods = ['chat.postEphemeral', 'chat.postMessage', 'chat.scheduleMessage', 'files.upload'];
+  const isTargetMethod = targetMethods.includes(method);
+
+  if (isTargetMethod && options?.thread_ts !== undefined && typeof options?.thread_ts !== 'string') {
+    logger.warn('The thread_ts argument need to be a string.');
```
seratch, line 662: Can you include the method name in the log message like we do in `warnIfFallbackIsMissing` method?

```suggestion
    logger.warn(`The given \`thread_ts\` value in the request payload for a ${method} call is a float value. We highly recommend using a string value instead.');
```

seratch (changes requested):
Thanks for taking the time to make this contribution! It's pretty nice to have the tests covering all the patterns 👍  One thing, I have a suggestion about the log message to make it more informative and consistent with the similar `warnIfFallbackIsMissing` method.

stevengill:
Thanks for sending this in @hi1280! I'd say just make the suggestion by @seratch and this looks good to merge!

packages/web-api/src/WebClient.ts
hi1280: @seratch Thank you for your review. I fixed the thing you suggested.

seratch (approved):
Thanks!

PR #1226, opened 2021-05-06
Add Timepicker to @slack/types

raycharius:
###  Summary

Noticed that @slack/types does not include an interface for the time picker block element. Added it and included it in all of the corresponding union types.

Also, I think there are some opportunities for refactoring that would make this package even more useful than it already is. While developing v.2 of [slack-block-builder](https://github.com/raycharius/slack-block-builder), I've noticed I'm doing a lot of typing on my end, since a lot of the types aren't available through this package. 

- Moving union types to their own exported types. For example, elements supported within the `ActionsBlock` having a type of `ActionsElements`. For `SectionBlock`, `SectionElements`. Etc.
- Moving enum values to their own exported enums.

Happy to take a shot at that, too, and send over a PR, if that's something you could see in the scope here.

Cheers!

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -1,4 +1,3 @@
-
 /*
  * Reusable shapes for argument values
  */
@@ -292,7 +291,16 @@ export interface SectionBlock extends Block {
   type: 'section';
   text?: PlainTextElement | MrkdwnElement; // either this or fields must be defined
   fields?: (PlainTextElement | MrkdwnElement)[]; // either this or text must be defined
-  accessory?: Button | Overflow | Datepicker | Timepicker | Select | MultiSelect | Action | ImageElement | RadioButtons | Checkboxes;
+  accessory?: Button
+    | Overflow
+    | Datepicker
+    | Timepicker
+    | Select
+    | MultiSelect
+    | Action
+    | ImageElement
+    | RadioButtons
+    | Checkboxes;
 }
 
 export interface FileBlock extends Block {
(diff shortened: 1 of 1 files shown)


seratch:
Hi @raycharius, thanks for taking the time to make this PR! The current revision seems to have a lint error. Can you fix the error?

```
> @slack/types@2.0.0 lint /home/runner/work/node-slack-sdk/node-slack-sdk/packages/types
> tslint --project .

ERROR: /home/runner/work/node-slack-sdk/node-slack-sdk/packages/types/src/index.ts:295:1 - Exceeds maximum line length of 120

lerna ERR! npm run lint stderr:
```

Apart from that, the changes look good to me. For the timing to merge this, I have been holding off adding timepicker in the Python SDK as it's still in a developer beta. See https://github.com/slackapi/python-slack-sdk/pull/876 for the context. However, the beta has been taking longer than I expect and even now, I cannot tell when the GA release will come yet. Therefore, I'm leaning toward merging the changes. If the CI failure is fixed, we are happy to merge the change.

Also, thanks for bringing the further suggestions! As I would like to make this PR's scope as small as possible, I will create a new issue for discussing other improvements. Let's discuss them separately!

seratch (changes requested):
the CI build is failing

raycharius:
Oops! Fixed. It's a bit ugly and hard to read with that union type on each line in the actual interface, but easily fixed if proposals for moving union types to their own exported type

seratch (approved):

stevengill (approved):

Issue #1227, opened 2021-05-06
Improvement suggestions on @slack/types for better TypeScript support


seratch:
### Description

This issue is for the discussion about  @raycharius's suggestions described at https://github.com/slackapi/node-slack-sdk/pull/1226 :

>Also, I think there are some opportunities for refactoring that would make this package even more useful than it already is. While developing v.2 of [slack-block-builder](https://github.com/raycharius/slack-block-builder), I've noticed I'm doing a lot of typing on my end, since a lot of the types aren't available through this package. 

- Moving union types to their own exported types. For example, elements supported within the `ActionsBlock` having a type of `ActionsElements`. For `SectionBlock`, `SectionElements`. Etc.
- Moving enum values to their own exported enums.

@raycharius
As long as the changes never bring breaking changes to existing apps, we are happy to discuss the changes. I haven't checked your library in detail yet but can you elaborate a bit more about the plans in your mind? If sharing a draft pull request with us is easier for you, that's also helpful to learn your intention.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [x] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

raycharius:
Cheers!

So, as one example, I've created my own enums for things like [button styles](https://github.com/raycharius/slack-block-builder/blob/initial-release-v2/src/constants/button-styles.ts), [filters](https://github.com/raycharius/slack-block-builder/blob/initial-release-v2/src/constants/filters.ts), etc. By putting those in their own enums in @slack/types, those could just be imported for use other than type checking.

And overall, some ideas on how to make it a bit more maintainable and flexible.

I'll put together a draft PR for this, as there's not much work involved, and it will provide us with some concrete context.

More soon!

raycharius:
Hi!

I've created a draft pull request demonstrating improvements to @slack/types.

- Moving all string values to exported enums. By doing so, these enums can be imported and used as values and for typing in projects.
[...]

PR #1228, opened 2021-05-07
Draft: Improve types

raycharius:
###  Summary

Draft pull request to discussion in [this issue](https://github.com/slackapi/node-slack-sdk/issues/1227) around improving @slack/types. See discussion [here](https://github.com/slackapi/node-slack-sdk/issues/1227).

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -1,11 +1,32 @@
 /*
  * Reusable shapes for argument values
  */
+
+export enum DialogElementType {
+  Text = 'text',
+  TextArea = 'textarea',
+  Select = 'select',
+}
+
+export enum DialogElementSubType {
+  Email = 'email',
+  Number = 'number',
+  Telephone = 'tel',
+  Url = 'url',
+}
+
+export enum DataSource {
+  Users = 'users',
+  Channels = 'channels',
+  Conversations = 'conversations',
+  External = 'external',
+}
+
 export interface Dialog {
   title: string;
   callback_id: string;
   elements: {
-    type: 'text' | 'textarea' | 'select';
+    type: DialogElementType;
     name: string; // shown to user
     label: string; // shown to user
     optional?: boolean;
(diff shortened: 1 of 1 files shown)


packages/types/src/index.ts
```diff
@@ -1,11 +1,32 @@
...
+  Channels = 'channels',
+  Conversations = 'conversations',
+  External = 'external',
+}
+
 export interface Dialog {
   title: string;
   callback_id: string;
   elements: {
-    type: 'text' | 'textarea' | 'select';
```
seratch, line 8: I'm afraid that changing union type to enum may be fully backward-compatible. I cannot tell what the possible troubles are by this but at least, unions and enums are not the same. To me, just naming the union types would be a safer refactoring. What do you think?

rr-codes: Just my two cents, but I really love enums more than unions aha

raycharius:
I've been playing around with this a bit, and you are totally right. For some reason, I always though of enums as a subset of union types, but it will not work with simple strings. 

Naming the union types would definitely make the code more structured, but that would be a change that's more internal for your team as opposed to offering value for developers using the SDK. 

There is one other possibility, to have those values available as imports – export an object literal instead of an enum and create union types that reference the values in those objects. Advantage would be having those available as constants instead of working with strings for those using the package, but not sure if that's in scope.

raycharius:
What are thoughts on the discriminating union types for more strict type checking? 

Advantages: 

- A much better and more type-safe experience with the package and SDK, such as with the `WebClient`.

Disadvantages: 

- Would require more swift updates to the package if API validation changes.
- It could be a breaking change, depending on how breaking is defined. It could throw compiler errors for objects that don't comply with the API validation, but unless there are some cases that I'm not taking into consideration, it would readonly only be showing issues with current code that would probably be rejected by the API anyway.

filmaj:
I am closing this pull request - but have referenced the commits in it extensively! - in favour of having the different discussions about each type of change in this PR separately. They are:

- https://github.com/slackapi/node-slack-sdk/issues/1816
- https://github.com/slackapi/node-slack-sdk/issues/1817
- https://github.com/slackapi/node-slack-sdk/issues/1818

Thank you very much for your work in this PR. I know it has dragged on for literal years, but I think we will actually incorporate at least some of the suggestions from this work into the project!

PR #1236, opened 2021-05-12
Fix #1235 ViewsOpenResponse type is incorrect

### LINKED ISSUE #1235: `ViewsOpenResponse` type is incorrect
### Description

the type is defined as 

```ts
export declare type ViewsOpenResponse = WebAPICallResult & {
    ok?: boolean;
    error?: string;
    response_metadata?: ResponseMetadata;
    needed?: string;
    provided?: string;
};
```

however in actuality, it has an additional `view` property. Here is an example opened view response:

```json
{
   "ok":true,
   "view":{
      "id":"V021U522M6G",
      "team_id":"OMITTED",
      "type":"modal",
      "blocks":[
         {
            "type":"section",
            "block_id":"JtAZ+",
            "text":{
               "type":"mrkdwn",
               "text":"Loading...",
               "verbatim":false
            }
         }
      ],
      "private_metadata":"",
      "callback_id":"CONFIRM_MESSAGE_SELECTION",
      "state":{
         "values":{
            
         }
      },
      "hash":"1620800476.cy2CiRSe",
      "title":{
         "type":"plain_text",
         "text":"Create a new issue",
         "emoji":true
      },
      "clear_on_close":false,
      "notify_on_close":false,
      "close":{
         "type":"plain_text",
         "text":"Cancel",
         "emoji":true
      },
      "submit":null,
      "previous_view_id":null,
      "root_view_id":"V021U522M6G",
      "app_id":"OMITTED",
      "external_id":"",
      "app_installed_team_id":"OMITTED",
      "bot_id":"OMITTED"
   },
   "response_metadata":{
      "scopes":[
      ]
   }
}
```

I suspect the same bug is present in other view related response types.

- [x] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

---

### Bug Report

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

#### Reproducible in:

package version: 6.2.3

node version: 12.19.0

OS version(s): N/A

#### Steps to reproduce:

1. Invoke `client.views.open(options)`

#### Expected result:

The response type corresponds to what properties actually exist. As far as I can tell, it is just missing the `view` property, which is quite important.

#### Actual result:

There is no such property defined on the interface.

seratch:
###  Summary

This pull request fixes #1235 by updating the API response types. see also: https://github.com/slackapi/java-slack-sdk/pull/743

### DIFF
--- a/packages/web-api/src/response/ViewsOpenResponse.ts
+++ b/packages/web-api/src/response/ViewsOpenResponse.ts
@@ -2,12 +2,141 @@
 import { WebAPICallResult } from '../WebClient';
 export type ViewsOpenResponse = WebAPICallResult & {
   ok?:                boolean;
+  warning?:           string;
   error?:             string;
-  response_metadata?: ResponseMetadata;
   needed?:            string;
   provided?:          string;
+  view?:              View;
+  response_metadata?: ResponseMetadata;
 };
 
 export interface ResponseMetadata {
   messages?: string[];
+}
+
+export interface View {
+  id?:                    string;
+  team_id?:               string;
+  type?:                  string;
+  title?:                 Close;
+  submit?:                Close;
+  close?:                 Close;
+  blocks?:                Block[];
+  private_metadata?:      string;
+  callback_id?:           string;
+  external_id?:           string;
+  state?:                 State;
+  hash?:                  string;
+  clear_on_close?:        boolean;
+  notify_on_close?:       boolean;
+  submit_disabled?:       boolean;
+  root_view_id?:          string;
+  previous_view_id?:      string;
+  app_id?:                string;
+  app_installed_team_id?: string;
+  bot_id?:                string;
+}
+
+export interface Block {
+  type?:            string;
+  block_id?:        string;
+  label?:           Close;
+  element?:         Element;
+  dispatch_action?: boolean;
+  hint?:            Close;
+  optional?:        boolean;
+  elements?:        Element[];
+  fallback?:        string;
+  image_url?:       string;
+  image_width?:     number;
+  image_height?:    number;
+  image_bytes?:     number;
+  alt_text?:        string;
+  title?:           Close;
+  text?:            Close;
+  fields?:          Close[];
+  accessory?:       Accessory;
+}
+
+export interface Accessory {
+  type?:         string;
+  image_url?:    string;
+  alt_text?:     string;
+  fallback?:     string;
+  image_width?:  number;
+  image_height?: number;
+  image_bytes?:  number;
+}
+
+export interface Element {
+  type?:                            string;
+  action_id?:                       string;
+  placeholder?:                     Close;
+  initial_value?:                   string;
+  multiline?:                       boolean;
+  min_length?:                      number;
+  max_length?:                      number;
+  dispatch_action_config?:          DispatchActionConfig;
+  options?:                         Option[];
+  initial_option?:                  Option;
+  confirm?:                         Confirm;
+  text?:                            Close;
+  url?:                             string;
+  value?:                           string;
+  style?:                           string;
+  initial_channel?:                 string;
+  response_url_enabled?:            boolean;
+  initial_conversation?:            string;
+  default_to_current_conversation?: boolean;
+  filter?:                          Filter;
+  initial_date?:                    string;
+  initial_time?:                    string;
+  min_query_length?:                number;
+  image_url?:                       string;
+  alt_text?:                        string;
+  fallback?:                        string;
+  image_width?:                     number;
+  image_height?:                    number;
+  image_bytes?:                     number;
+  initial_user?:                    string;
+}
+
+export interface Confirm {
+  title?:   Close;
+  text?:    Close;
+  confirm?: Close;
+  deny?:    Close;
+  style?:   string;
+}
+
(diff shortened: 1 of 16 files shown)


packages/web-api/src/response/ViewsOpenResponse.ts
```diff
@@ -2,12 +2,141 @@
...
+
+export interface Close {
+  type?:     Type;
+  text?:     string;
+  emoji?:    boolean;
+  verbatim?: boolean;
+}
+
+export enum Type {
+  Empty = '',
```
seratch, line 120: This looks invalid - I will look into this later on.

packages/web-api/src/response/ViewsPublishResponse.ts
```diff
@@ -2,12 +2,141 @@
...
 export interface ResponseMetadata {
   messages?: string[];
 }
+
+export interface View {
+  id?:                    string;
+  team_id?:               string;
+  type?:                  string;
+  title?:                 Close;
+  submit?:                Close;
```
rr-codes, line 22: I understand type wise that `title`, `submit`, and `close` are the same (I think so at least), but semantically, it seems really weird to have the types of `title` and `submit` be `Close`

packages/web-api/src/response/ViewsPublishResponse.ts
```diff
@@ -2,12 +2,141 @@
...
+  view?:              View;
+  response_metadata?: ResponseMetadata;
 };
 
 export interface ResponseMetadata {
   messages?: string[];
 }
+
+export interface View {
+  id?:                    string;
```
rr-codes, line 18: also, is there a reason why every single property is optional, even though this is not the case? `id` for example should always be present

packages/web-api/src/response/ViewsPushResponse.ts
```diff
@@ -2,12 +2,141 @@
...
+  trigger_actions_on?: string[];
+}
+
+export interface Filter {
+  exclude_external_shared_channels?: boolean;
+  exclude_bot_users?:                boolean;
+}
+
+export interface Option {
+  text?:        Close;
```
rr-codes, line 135: ya again this type semantically shouldn't be `Close`

packages/web-api/src/response/ViewsOpenResponse.ts
```diff
@@ -2,12 +2,141 @@
...
+  image_url?:    string;
+  alt_text?:     string;
+  fallback?:     string;
+  image_width?:  number;
+  image_height?: number;
+  image_bytes?:  number;
+}
+
+export interface Element {
+  type?:                            string;
```
rr-codes, line 72: `type` property values are always defined as far as I'm aware

seratch: @rr-codes Thanks for the comment. You're right. This `type` and some others are not optional but as we generate the types using code generator (quicktype), this is not customizable.

packages/web-api/src/response/ViewsPublishResponse.ts
seratch: the same as my above comment

packages/web-api/src/response/ViewsPublishResponse.ts
seratch: This is automatically determined by the generator tool. Unfortunately, there is no way to customize the name,

packages/web-api/src/response/ViewsPushResponse.ts
seratch: same as above

seratch (commented):
@rr-codes Thanks for your comment (I really appreciate it). However, as we use code generator, all the points you mentioned are not customizable. We've decided to go with these limitations.

packages/web-api/src/response/ViewsOpenResponse.ts
seratch: I've looked into this and updating this is not so simple. Let us leave it as-is this time.

rr-codes:
@seratch out of curiosity, what is the generator that's being used to generate the types? Is it open source?

seratch:
@rr-codes As I mentioned above, we are using https://github.com/quicktype/quicktype for it.

rr-codes:
@seratch oh whoops completely missed that aha, thanks so much!!

rr-codes:
Also sorry @seratch one last question, whats the actual workflow of using Prototype for the types? Does it use [this schema ](https://raw.githubusercontent.com/slackapi/slack-api-specs/master/web-api/slack_web_openapi_v2_without_examples.json) with the json schema option?

seratch:
@rr-codes Ideally, we should use the slack-api-specs but we don't use it. The main reason is that the outputs in the project are incomplete and are not up-to-date. We use the JSON data generated by the official Java SDK. Refer to https://github.com/slackapi/node-slack-sdk/pull/1188 for details.

misscoded (approved):

PR #1242, opened 2021-05-20
Fix #1240 by enabling to pass integer as post_at in chat.scheduleMessage API calls

### LINKED ISSUE #1240: post_at as option in schedule message has the wrong type
### Description
I think `ChatScheduleMessageArguments` interface has wrong type for `post_at` property. In my opinion, it should be `number` instead of `string`.

### What type of issue is this?
- [x] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements

---

### Bug Report

As in [documentation in "code"](https://api.slack.com/methods/chat.scheduleMessage/code) section we see `post_at` as number. While `@slack/web-api` package tells me that property should be a string.
As I try to call `webClient.chat.scheduleMessage({..., post_at: <Unix Epoch timestamp as string>})` I encountered a problem in response:
```
[ERROR]  web-api:WebClient:0 must provide a number [json-pointer:/post_at]
error Error: An API error occurred: invalid_arguments
    at Object.platformErrorFromResult (C:\<project-path>\node_modules\@slack\web-api\src\errors.ts:94:5)
    at WebClient.apiCall (C:\<project-path>\node_modules\@slack\web-api\src\WebClient.ts:196:13)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
{
  code: 'slack_webapi_platform_error',
  data: {
    ok: false,
    error: 'invalid_arguments',
    response_metadata: { messages: [Array], scopes: [Array], acceptedScopes: [Array] }
  }
}
```

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

#### Reproducible in:

package version: 6.2.3

node version: 12.18.3

OS version(s): Windows 10

seratch:
###  Summary

This pull request fixes #1240 by correcting the argument type for `post_at` in `ChatScheduleMessageArguments`. The argument should accept either `number` (specifically integer value) or `string` value for it. Although both `number` and `string` actually works, the API document https://api.slack.com/methods/chat.scheduleMessage recommends `number` over `string`.

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -1225,7 +1225,7 @@ export interface ChatPostMessageArguments extends WebAPICallOptions, TokenOverri
 export interface ChatScheduleMessageArguments extends WebAPICallOptions, TokenOverridable {
   channel: string;
   text?: string;
-  post_at: string;
+  post_at: string | number;
   as_user?: boolean;
   attachments?: MessageAttachment[];
   blocks?: (KnownBlock | Block)[];
--- a/packages/web-api/test/types/webclient-named-method-types.test-d.ts
+++ b/packages/web-api/test/types/webclient-named-method-types.test-d.ts
@@ -65,7 +65,13 @@ expectType<Promise<ChatScheduleMessageResponse>>(web.chat.scheduleMessage({
 }));
 expectType<Promise<ChatScheduleMessageResponse>>(web.chat.scheduleMessage({
   channel: 'C111',
-  post_at: '11111',
+  post_at: '1621497568',
+  text: 'Hi there!',
+  blocks: [],
+}));
+expectType<Promise<ChatScheduleMessageResponse>>(web.chat.scheduleMessage({
+  channel: 'C111',
+  post_at: 1621497568,
   text: 'Hi there!',
   blocks: [],
 }));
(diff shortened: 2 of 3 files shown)


packages/web-api/src/methods.ts
```diff
@@ -1225,7 +1225,7 @@ export interface ChatPostMessageArguments extends WebAPICallOptions, TokenOverri
 export interface ChatScheduleMessageArguments extends WebAPICallOptions, TokenOverridable {
   channel: string;
   text?: string;
-  post_at: string;
+  post_at: string | number;
```
misscoded, line 1228: One thought: by not limiting it to only `number`, there's a greater possibility that folks will continue to interpret `string` as possibly being a timestamp, as the dev in this case did. By switching it to `number`, we reduce the chance of that happening.

seratch: @misscoded As I mentioned in this PR's description and my response to the developer, using `string` here is totally fine. It works. The issue the developer had was that they used non-integer value. For this reason, we still can have `string` as an acceptable type for it. If we remove `string` from here, it's a breaking change.

stevengill (approved):
LGTM

seratch:
Thanks for your reviews!

Issue #1241, opened 2021-05-20
Add Support for Token Rotation


misscoded:
### Description

Developers will soon have the ability to opt their apps into advanced token security via token rotation. We should introduce logic to incorporate these new tokens, as well as refresh behavior.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

PR #1244, opened 2021-05-21
Add Support for Token Rotation

misscoded:
#1241

Although the original consensus was to tie the expiry check and subsequent refresh to making a call to the API, it was at some point decided that this should occur when events are received. The current state of this draft reflects that decision.

### HOW IT WORKS
On OAuth installation (the only installation flow allowed when token rotation is enabled), new access tokens are stored as they have been, but now include an accompanying `refresh_token` and `expires_in`. The `refresh_token` (does not expire) is used to "trade-in" for a new `access_token` and `refresh_token` pair. 

For each token (bot|user), we use the `expires_in` value to calculate and store a corresponding future expiry timestamp (UTC, seconds).

Each time an event is received, `authorize` gets run. Here we determine if token rotation is enabled. If so, we check the expiry of the token(s) and refresh those that have expired or will expire within 2 hours. We then update the returned `authResult` and the `Installation`.

### NOTABLE CHANGES
1. `AuthorizationResult` has four new keys: `botRefreshToken`, `botTokenExpiresAt`, `userRefreshToken`, and `userTokenExpiresAt`. This was to handle apps that have both User and Bot tokens while still keeping the shape flat.

2. `Installation` has four new keys: a `refreshToken` and `expiresAt` for both the `user` and `bot` objects.

### TO CONSIDER
2 hours is the current window of time we try to get ahead of an expiry. If customization of the TTL is introduced later on, this is likely too inflexible and will need to be revisited.

Refreshing on incoming events still feels strange, but it appears to work as expected thus far. Logic for the expiry checks are run on every incoming event, which is no problem with single user testing, but I'm unsure if results would be different with events constantly pouring in across many channels.

### TODO / OUTSTANDING
- [x] Address TypeScript issues
- [ ] Write tests - _will write during beta_
- [ ] ~Retry on failure to refresh (discussion)~
- [x] Revisit possibility of concurrent calls (discussion)
- [ ] ~Introduce `deleteInstallation` (discussion)~
- [x] Verify non-token rotation-enabled apps still work as expected
- [x] Stress testing

### DIFF
--- a/packages/oauth/src/index.ts
+++ b/packages/oauth/src/index.ts
@@ -91,7 +91,7 @@ export class InstallProvider {
   }
 
   /**
-   * Fetches data from the installationStore for non Org Installations.
+   * Fetches data from the installationStore
    */
   public async authorize(source: InstallationQuery<boolean>): Promise<AuthorizeResult> {
     try {
@@ -129,6 +129,87 @@ export class InstallProvider {
         authResult.botToken = queryResult.bot.token;
         authResult.botId = queryResult.bot.id;
         authResult.botUserId = queryResult.bot.userId;
+
+        // Token Rotation Enabled (Bot Token)
+        if (queryResult.bot.refreshToken !== undefined) {
+          authResult.botRefreshToken = queryResult.bot.refreshToken;
+          authResult.botTokenExpiresAt = queryResult.bot.expiresAt; // utc, seconds
+        }
+      }
+
+      // Token Rotation Enabled (User Token)
+      if (queryResult.user.refreshToken !== undefined) {
+        authResult.userRefreshToken = queryResult.user.refreshToken;
+        authResult.userTokenExpiresAt = queryResult.user.expiresAt; // utc, seconds
+      }
+
+      /*
+      * Token Rotation (Expiry Check + Refresh)
+      * The presence of `(bot|user)TokenExpiresAt` indicates having opted into token rotation.
+      * If the token has expired, or will expire within 2 hours, the token is refreshed and
+      * the `authResult` and `Installation` are updated with the new values.
+      */
+      if (authResult.botRefreshToken !== undefined || authResult.userRefreshToken !== undefined) {
+        const currentUTCSec = Math.floor(Date.now() / 1000); // seconds
+        const tokensToRefresh: string[] = [];
+        let client: WebClient;
+
+        if (authResult.botRefreshToken !== undefined && authResult.botTokenExpiresAt !== undefined) {
+          const botTokenExpiresIn = authResult.botTokenExpiresAt - currentUTCSec;
+          if (botTokenExpiresIn <= 7200) { // 2 hours
+            tokensToRefresh.push(authResult.botRefreshToken);
+          }
+        }
+
+        if (authResult.userRefreshToken !== undefined && authResult.userTokenExpiresAt !== undefined) {
+          const userTokenExpiresIn = authResult.userTokenExpiresAt - currentUTCSec;
+          if (userTokenExpiresIn <= 7200) { // 2 hours
+            tokensToRefresh.push(authResult.userRefreshToken);
+          }
+        }
+
+        if (tokensToRefresh.length > 0) {
+          const installationUpdates: any = { ...queryResult }; // TODO :: TS
+          client = new WebClient(undefined, this.clientOptions);
+
+          // TODO :: Concurrent Calls?
+          // TODO :: Retry Strategy
+          for (const refreshToken of tokensToRefresh) {
+
+            const refreshResp = await client.oauth.v2.access({
+              client_id: this.clientId,
+              client_secret: this.clientSecret,
+              grant_type: 'refresh_token',
+              refresh_token: refreshToken,
+            }) as OAuthV2Response;
+
+            // TODO :: Sort out TS issues
+            const tokenType: 'bot' | 'user' = refreshResp.token_type!; // TODO :: TS
+
+            if (tokenType === 'bot') {
+              authResult.botToken = refreshResp.access_token;
+              authResult.botRefreshToken = refreshResp.refresh_token;
+              authResult.botTokenExpiresAt = currentUTCSec + refreshResp.expires_in!; // TODO :: TS
+            }
+
+            if (tokenType === 'user') {
+              authResult.userToken = refreshResp.access_token;
+              authResult.userRefreshToken = refreshResp.refresh_token;
+              authResult.userTokenExpiresAt = currentUTCSec + refreshResp.expires_in!; // TODO :: TS
+            }
+
+            installationUpdates[tokenType].token = refreshResp.access_token;
+            installationUpdates[tokenType].refreshToken = refreshResp.refresh_token;
+            installationUpdates[tokenType].expiresAt = currentUTCSec + refreshResp.expires_in!; // TODO :: TS
+
+            const updatedInstallation = {
+              ...installationUpdates,
+              [tokenType]: { ...queryResult[tokenType], ...installationUpdates[tokenType] }, // bot | user
+            };
+
+            await this.installationStore.storeInstallation(updatedInstallation);
+          }
+        }
       }
 
       return authResult;
@@ -227,6 +308,7 @@ export class InstallProvider {
       // Start: Build the installation object
       let installation: Installation;
       let resp: OAuthV1Response | OAuthV2Response;
+
       if (this.authVersion === 'v1') {
         // convert response type from WebApiCallResult to OAuthResponse
         const v1Resp = await client.oauth.access({
@@ -269,6 +351,7 @@ export class InstallProvider {
         resp = v1Resp;
         installation = v1Installation;
       } else {
+
         // convert response type from WebApiCallResult to OAuthResponse
         const v2Resp = await client.oauth.v2.access({
           code,
@@ -294,9 +377,13 @@ export class InstallProvider {
           authVersion: 'v2',
         };
 
+        const currentUTC = Math.floor(Date.now() / 1000); // utc, seconds
+
+        // Installation has Bot Token
         if (v2Resp.access_token !== undefined && v2Resp.scope !== undefined && v2Resp.bot_user_id !== undefined) {
-          // A bot user/scope was requested
+
           const authResult = await runAuthTest(v2Resp.access_token, this.clientOptions);
+
           v2Installation.bot = {
             scopes: v2Resp.scope.split(','),
             token: v2Resp.access_token,
@@ -305,20 +392,31 @@ export class InstallProvider {
           };
 
           if (v2Resp.is_enterprise_install) {
-            // if it is an org enterprise install, add the enterprise url
             v2Installation.enterpriseUrl = authResult.url;
           }
 
-        } else if (v2Resp.authed_user.access_token !== undefined) {
-          // Only user scopes were requested
+          // Token Rotation is Enabled
+          if (v2Resp.refresh_token !== undefined && v2Resp.expires_in !== undefined) {
+            v2Installation.bot.refreshToken = v2Resp.refresh_token;
+            v2Installation.bot.expiresAt = currentUTC + v2Resp.expires_in; // utc, seconds
+          }
+        }
+
+        // Installation has User Token
+        if (v2Resp.authed_user !== undefined && v2Resp.authed_user.access_token !== undefined) {
+
           // TODO: confirm if it is possible to do an org enterprise install without a bot user
           const authResult = await runAuthTest(v2Resp.authed_user.access_token, this.clientOptions);
-          if (v2Resp.is_enterprise_install) {
+
+          if (v2Resp.is_enterprise_install && v2Installation.enterpriseUrl === undefined) {
             v2Installation.enterpriseUrl = authResult.url;
           }
-        } else {
-          // TODO: make this a coded error
-          throw new Error('The response from the authorization URL contained inconsistent information. Please file a bug.');
+
+          // Token Rotation is Enabled
+          if (v2Resp.authed_user.refresh_token !== undefined && v2Resp.authed_user.expires_in !== undefined) {
+            v2Installation.user.refreshToken = v2Resp.authed_user.refresh_token;
+            v2Installation.user.expiresAt = currentUTC + v2Resp.authed_user.expires_in; // utc, seconds
+          }
         }
 
         resp = v2Resp;
@@ -333,6 +431,7 @@ export class InstallProvider {
           configurationUrl: resp.incoming_webhook.configuration_url,
         };
       }
+
       if (installOptions !== undefined && installOptions.metadata !== undefined) {
         // Pass the metadata in state parameter if exists.
         // Developers can use the value for additional/custom data associated with the installation.
@@ -455,7 +554,7 @@ export interface InstallationStore {
     installation: Installation<AuthVersion, boolean>,
     logger?: Logger): Promise<void>;
   fetchInstallation:
-    (query: InstallationQuery<boolean>, logger?: Logger) => Promise<Installation<'v1' | 'v2', boolean>>;
+  (query: InstallationQuery<boolean>, logger?: Logger) => Promise<Installation<'v1' | 'v2', boolean>>;
 }
 
 // using a javascript object as a makeshift database for development
@@ -565,12 +664,16 @@ export interface Installation<AuthVersion extends ('v1' | 'v2') = ('v1' | 'v2'),
 
   user: {
     token: AuthVersion extends 'v1' ? string : (string | undefined);
+    refreshToken?: AuthVersion extends 'v1' ? never : (string | undefined);
+    expiresAt?: AuthVersion extends 'v1' ? never : (number | undefined);
     scopes: AuthVersion extends 'v1' ? string[] : (string[] | undefined);
     id: string;
   };
 
   bot?: {
     token: string;
+    refreshToken?: string;
+    expiresAt?: number;
     scopes: string[];
     id: string; // retrieved from auth.test
     userId: string;
@@ -636,7 +739,11 @@ export type OrgInstallationQuery = InstallationQuery<true>;
 // of Bolt.
 export interface AuthorizeResult {
   botToken?: string;
+  botRefreshToken?: string;
+  botTokenExpiresAt?: number;
   userToken?: string;
+  userRefreshToken?: string;
+  userTokenExpiresAt?: number;
   botId?: string;
   botUserId?: string;
   teamId?: string;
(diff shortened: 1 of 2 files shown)


packages/oauth/src/index.ts
```diff
@@ -91,7 +91,7 @@ export class InstallProvider {
   }
 
   /**
-   * Fetches data from the installationStore for non Org Installations.
```
seratch, line 94: Nice improvements! As this pull request takes a bit long before merging into main branch, can you submit another pull request including some token rotation unrelated changes?

packages/oauth/src/index.ts
```diff
@@ -636,7 +739,11 @@ export type OrgInstallationQuery = InstallationQuery<true>;
 // of Bolt.
 export interface AuthorizeResult {
   botToken?: string;
+  botRefreshToken?: string;
+  botTokenExpiresAt?: number;
```
seratch, line 743: How about having a comment like `// UTC time in seconds` for the expiresAt properties?

packages/oauth/src/index.ts
```diff
@@ -129,6 +129,87 @@ export class InstallProvider {
...
+          if (userTokenExpiresIn <= 7200) { // 2 hours
+            tokensToRefresh.push(authResult.userRefreshToken);
+          }
+        }
+
+        if (tokensToRefresh.length > 0) {
+          const installationUpdates: any = { ...queryResult }; // TODO :: TS
+          client = new WebClient(undefined, this.clientOptions);
+
+          // TODO :: Concurrent Calls?
```
seratch, line 175: Concurrent calls are fine as long as no call won't be skipped. Specifically, we cannot have delayed calls here as the calls can be forcibly terminated when running on a FaaS runtime (AWS Lambda, Google Cloud Functions, Azure Functions). We may want to have a flag option like processBeforeResponse in Bolt.

packages/oauth/src/index.ts
```diff
@@ -129,6 +129,87 @@ export class InstallProvider {
...
+        if (authResult.userRefreshToken !== undefined && authResult.userTokenExpiresAt !== undefined) {
+          const userTokenExpiresIn = authResult.userTokenExpiresAt - currentUTCSec;
+          if (userTokenExpiresIn <= 7200) { // 2 hours
+            tokensToRefresh.push(authResult.userRefreshToken);
+          }
+        }
+
+        if (tokensToRefresh.length > 0) {
+          const installationUpdates: any = { ...queryResult }; // TODO :: TS
+          client = new WebClient(undefined, this.clientOptions);
```
seratch, line 173: This `client` can be `const client` as it's used only inside this code block (scope).

packages/oauth/src/index.ts
```diff
@@ -129,6 +129,87 @@ export class InstallProvider {
...
+          for (const refreshToken of tokensToRefresh) {
+
+            const refreshResp = await client.oauth.v2.access({
+              client_id: this.clientId,
+              client_secret: this.clientSecret,
+              grant_type: 'refresh_token',
+              refresh_token: refreshToken,
+            }) as OAuthV2Response;
+
+            // TODO :: Sort out TS issues
```
seratch, line 186: Indeed, as we use automatic code generator for the types, we are unable to have non-optional properties here. 
When we can assume the property exists 100%, using `!` can be okay but if I wrote this part, I would have if statements verifying the existence. Doing this satisfies the TS compiler plus may detect really unexpected situations (say, Slack server-side regressions / incidents).

```typescript
            const tokenType = refreshResp.token_type;
            if (tokenType === undefined) {
              throw new Error("unexpectedly token_type is missing here!");
            }
            const expireIn = refreshResp.expires_in;
            if (expireIn === undefined) {
              throw new Error("unexpectedly expire_in is missing here!");
            }
            const expiresAt: number = currentUTCSec + expireIn;

            if (tokenType === 'bot') {
              authResult.botToken = refreshResp.access_token;
              authResult.botRefreshToken = refreshResp.refresh_token;
              authResult.botTokenExpiresAt = expiresAt;
            }
```

packages/oauth/src/index.ts
```diff
@@ -129,6 +129,87 @@ export class InstallProvider {
...
+      * the `authResult` and `Installation` are updated with the new values.
+      */
+      if (authResult.botRefreshToken !== undefined || authResult.userRefreshToken !== undefined) {
+        const currentUTCSec = Math.floor(Date.now() / 1000); // seconds
+        const tokensToRefresh: string[] = [];
+        let client: WebClient;
+
+        if (authResult.botRefreshToken !== undefined && authResult.botTokenExpiresAt !== undefined) {
+          const botTokenExpiresIn = authResult.botTokenExpiresAt - currentUTCSec;
+          if (botTokenExpiresIn <= 7200) { // 2 hours
```
seratch, line 159: We are doing the same twice here. How about extracting the logic as a private function `function isExpiring(expiresIn, currentUTCSeconds): boolean`?

packages/oauth/src/index.ts
misscoded: Good catch! This is a remnant from a change I made during cleanup. Will fix.

packages/oauth/src/index.ts
misscoded: Great point. Since it's only ever two calls, do you think implementing a concurrent strategy is necessary? Also, thoughts on how to move forward with the retry logic if a call fails?

packages/oauth/src/index.ts
misscoded: Thanks! The correction to the comment above is the only non-token rotation change included here. Since it's not time-sensitive and it's just a comment, I think it's fine to just wait until this PR goes in to see it fixed.

packages/oauth/src/index.ts
misscoded: This type is actually not using the generated code version of the response, but it brings up the greater question: should it? Are we throwing out all of our hand-rolled types and opting for the generated ones at this point?

Thanks for your suggestion for a resolution the current state of things -- I'll plan to incorporate it.

packages/oauth/src/index.ts
seratch: >do you think implementing a concurrent strategy is necessary?

Yes, I think that it's valuable for end-user experience and less 3 second timeouts. To avoid having an extra overhead, we would like to asynchronously do these refreshing API calls as it's not necessary for **this** incoming request handling. If the `oauth.v2.access` here takes 300-500 milliseconds in average, calling twice may take 1 second in a worst case scenario. This may not be a small overhead.

I haven't checked if this works but one possible idea would be having a private function like `async function refreshExpiringToken(tokensToRefresh)` and calling the method asynchronously (no `await`) by default. In addition to that, we can consider FaaS applications by calling the same method with `await` if `processBeforeResponse` is true.

>Also, thoughts on how to move forward with the retry logic if a call fails?

It depends on the error. If the error is recoverable (e.g., connection failure, timeouts), we can retry with the same set of parameters. If the response has an error code like invalid parameters or invalid auth, of course, we don't need to retry the call anymore.

packages/oauth/src/index.ts
seratch: >This type is actually not using the generated code version of the response, but it brings up the greater question: should it?

Yes, the type will be updated once the properties are GAed (or at least in public beta). With the current approach, tall the properties in responses are optional. There is no way to exclude specific properties at this point. During the beta period, you can go with (=internally cast the type) an internal type in this source file.

>Are we throwing out all of our hand-rolled types and opting for the generated ones at this point?

No, we don't have the bandwidth to guarantee the quality of it for all API responses and I have to say that the situation can be the same in the future. That's why we are auto generating the code.

stevengill: For OAuth, we do have hand rolled types right now. See https://github.com/slackapi/node-slack-sdk/blob/main/packages/oauth/src/index.ts#L705-L726

But maybe we shouldn't have these and instead be switching over to using the generated types

packages/oauth/src/index.ts
misscoded: After circling back to this, I've decided to pull out the bad use of a magic number into a `const` above. I don't think that creating a new function for a comparison operator is a good choice here, as it causes even more legwork for the reader.

mwbrooks (commented):
Nice work @misscoded and thanks for the detailed write up about Token Rotation!

Two questions that come to mind are:

**1. Tokens refreshed on events:** When an event is received, the `authorize` is run. What happens when an app doesn't subscribe to any events? For example, the app may only want to use `channels.list` and `chat.postMessage`. Would token rotation not work in that situation?

**2. Refreshing a token:** When we refresh a token, is the previous token immediately invalidated?

I can imagine a scenario where multiple events trigger at once, causing the same refresh token to be refreshed by multiple event handlers at the same time. Each one will be written to the database and there's no guarantee it would be the most "fresh" token. 

In this scenario, would any of the refreshed tokens be valid? Or is there a chance that we would save an "old" token to the database?

packages/oauth/src/index.ts
seratch: Thanks. That's okay. Another factor to generally consider is that naming a chunk of logic may help us easily understand what the lines of code do but this time, having a method or not is not a big deal at all.

On another note, while checking this pull request again, I noticed that we can have two groups of tokens here.

1. the tokens already expired
2. the tokens that are going to be expired in 2 hours

In the case of 1. , we always must immediately call oauth.v2.access API with a refresh token for following usage. 

On the other hand, in the case of 2., we basically don't want to have the refresh API call in the synchronous operation for an incoming request as it's not yet expired. As I mentioned in my comment, we may want to have a new flag option similar to `processBeforeResponse` for FaaS users. If the flag is true, we have to call the refresh API synchronously.

I don't know having 1. and 2. in the `tokensToRefresh` array is intentional but a bit more consideration for 2. improve app's response time and it would be valuable for user experience and less timeouts. Thoughts? If you want to see example code, I can come up with Python draft PR in a few days.

misscoded: I completely agree re: the chunks of logic and its role in aiding readability, so I do understand your point. I'll take another pass as it and see if I can improve upon what's there.

Regarding the second part, you bring up excellent concerns. The use of `tokensToRefresh` was definitely intentional when I wrote it because we considered the situation either refresh vs. no refresh, but I agree with you that we can do better by drilling down. Given my lack of first-hand experience with the FaaS experience, I will rely on you heavily to ensure the incorporation of an additional open is been incorporated properly.

I'll work on the above over the next day or so and would like to compare notes with your implementation so that we can end up with the best solution. Thanks for your continued input!

packages/oauth/src/index.ts
misscoded: Every time we remove types that reliably indicate what properties are present I die a little inside, but I do acknowledge the motivation outlined above. 🙂  

Just let me know if you two would like me to remove it when the generated types are updated to reflect the payload changes!

packages/oauth/src/index.ts
stevengill: Regarding 2, I think for the beta we stick with the current approach (make the listener code wait until token refresh is done, even if token isn't expired). I'd like to see if people in the real world start getting 3 second timeout issues with this change. If the answer is no, than our solution is sufficient and we don't need to optimize this for now. If timeouts do increase, then we may want to look at moving token refresh code into web-api or possibly or even adding an entry point to run token refresh after the listener is completed.

seratch: Thanks for sharing thoughts, you both! 

I agree that we can go with a simple approach first and see how it goes during the beta. The discussion here should be helpful when we optimize the part when it's needed. Let's go with this way for its beta releases.

packages/oauth/src/index.ts
stevengill: haha I agree. Lets not remove anything right now. Maybe in the future we can get more reliable with the generated response types and then we can revist this.

misscoded:
@mwbrooks

> **1. Tokens refreshed on events:** When an event is received, the `authorize` is run. What happens when an app doesn't subscribe to any events? For example, the app may only want to use `channels.list` and `chat.postMessage`. Would token rotation not work in that situation?

This is a _very_ valid point, IMO, and was my biggest push back when Steve/Kaz hit me with the redirection to event-based refresh rather than API-call-based. I can't speak in depth about the rationale beyond what I understand to be considering it an edge case and unlikely to happen, but I could be misremembering that so I'll tap @stevengill to fill in the details of the conversation where it was decided.

> **2. Refreshing a token:** When we refresh a token, is the previous token immediately invalidated?

So, refresh tokens actually expire 5 minutes after they're first used, meaning that the same refresh token could be used multiple times back-to-back. That said, only two `access_tokens` can exists at any given point, so a dev will have -- at most -- two unique, _valid_ `access_token`s and a single `refresh_token`. Once those two exist, any more refreshes beyond that would invalidate the older of the two access tokens.

> I can imagine a scenario where multiple events trigger at once, causing the same refresh token to be refreshed by multiple event handlers at the same time. Each one will be written to the database and there's no guarantee it would be the most "fresh" token. **In this scenario, would any of the refreshed tokens be valid? Or is there a chance that we would save an "old" token to the database?**

In the case of multiple events coming in at once, assuming each is written to the DB, I've been assured it's been deemed safe to assume that the last one (or two) would be valid.

stevengill:
> This is a very valid point, IMO, and was my biggest push back when Steve/Kaz hit me with the redirection to event-based refresh rather than API-call-based. I can't speak in depth about the rationale beyond what I understand to be considering it an edge case and unlikely to happen, but I could be misremembering that so I'll tap @stevengill to fill in the details of the conversation where it was decided.

I think the thinking here is that the main way to interact with client in bolt is to use a listener. Before any listener gets executed, authorize will run (which means the token can get refreshed). 

If they use client outside of the listener, that client will potentially have the wrong token attached. We may want to see if we can update the token on the top level app.client property (or remove this property?). Issue is the way webclient initializes, it takes in a token. We don't have a mechanism to change the token on a web client instance as far as I know. 

I wouldn't be surprised if people file an issue for this. Lets see how it goes. We very well may have to do some refactoring to web-api.

packages/web-api/src/methods.ts
```diff
@@ -506,6 +506,7 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
     access: bindApiCall<OAuthAccessArguments, OauthAccessResponse>(this, 'oauth.access'),
     v2: {
       access: bindApiCall<OAuthV2AccessArguments, OauthV2AccessResponse>(this, 'oauth.v2.access'),
+      exchange: bindApiCall<OAuthV2AccessArguments, OauthV2AccessResponse>(this, 'oauth.v2.exchange'),
```
seratch, line 509: TODO: Need to have `OAuthV2ExchangeArguments` for this. The arguments are not the same with `oauth.v2.access`. It has `token`. Regarding the response, it's fine to use `WebAPICallResult` for now.

packages/oauth/src/index.ts
```diff
@@ -137,6 +195,52 @@ export class InstallProvider {
...
+    const refreshPromises = tokensToRefresh.map(async (refreshToken) => {
+      return await client.oauth.v2.access({
+        client_id: this.clientId,
+        client_secret: this.clientSecret,
+        grant_type: 'refresh_token',
+        refresh_token: refreshToken,
+      }).catch(e => e) as OAuthV2ExchangeResponse;
+    });
+
+    return Promise.all(refreshPromises);
```
misscoded, line 241: I spent a good amount of time trying to best solution this, but would love input. 

With our current tsconfig, we don't have the ability to use `Promise.allSettled` (we use `es2017` and would need `es2020`), which is what we want in this case (to wait until all Promises are settled, despite outcome of each).

`Promise.all` will fail fast if a rejection occurs, and I think under normal circumstances that would be handled fine with the additional `.catch(e => e)` that's included for each individual Promise, but in the case of a failure, the WebAPI actually throws an error before that catch can happen (I _think_ that's what I've been observing while testing. Feel free to challenge that).

packages/oauth/src/index.ts
```diff
@@ -129,6 +129,64 @@ export class InstallProvider {
...
+      * Token Rotation (Expiry Check + Refresh)
+      * The presence of `(bot|user)TokenExpiresAt` indicates having opted into token rotation.
+      * If the token has expired, or will expire within 2 hours, the token is refreshed and
+      * the `authResult` and `Installation` are updated with the new values.
+      */
+      if (authResult.botRefreshToken !== undefined || authResult.userRefreshToken !== undefined) {
+        const currentUTCSec = Math.floor(Date.now() / 1000); // seconds
+        const tokensToRefresh: string[] = this.checkTokenExpiry(authResult, currentUTCSec);
+
+        if (tokensToRefresh.length > 0) {
```
misscoded, line 156: I pulled out some logic (`checkTokenExpiry` and `refreshExpiringTokens`) in an attempt to clean up and separate concerns, but stopped short of going beyond this point because I couldn't figure out a clean way to do so without also mutating `authResult` within the confines of the helper function. Happy to take someone else's attempt and incorporate it.

packages/oauth/src/index.ts
```diff
@@ -725,6 +859,20 @@ interface OAuthV2Response extends WebAPICallResult {
   };
 }
 
+export interface OAuthV2ExchangeResponse extends WebAPICallResult {
```
misscoded, line 862: I had to add this because of conflicts created in other, pre-existing parts of the code. If someone else wants to take a stab at how to resolve this in another way, again, more than happy to accept a better solution!

packages/oauth/src/index.ts
```diff
@@ -725,6 +859,20 @@ interface OAuthV2Response extends WebAPICallResult {
...
+  scope: string;
+  token_type: 'bot' | 'user';
+  access_token: string;
+  refresh_token: string;
+  expires_in: number;
+  bot_user_id?: string;
+  team: { id: string, name: string };
+  enterprise: { name: string, id: string } | null;
+  is_enterprise_install: boolean;
+  response_metadata: {}; // TODO
```
misscoded, line 873: I couldn't find what the actual shape of this property was, neither in the docs nor in conversations. If anyone happens to know, would like to update this.

packages/oauth/src/index.ts
```diff
@@ -137,6 +195,52 @@ export class InstallProvider {
     }
   }
 
+  /**
+   * checkTokenExpiry determines access tokens' eligibility for refresh.
+   *
+   * The return value is an Array of expired or soon-to-expire access tokens.
+   */
+  private checkTokenExpiry(authResult: AuthorizeResult, currentUTCSec: number): string[] {
```
seratch, line 203: I think `check` is generally a bit unclear naming. How about naming the method using `detect` or `extract`? The method's responsibility is to return expired/expiring tokens from the auth result.

```suggestion
  private detectExpiredOrExpiringTokens(authResult: AuthorizeResult, currentUTCSec: number): string[] {
```

seratch: On another note, these new private methods are stateless (=they don't access the internal state of this object and probably they should not). You may want to place them at top level. You can call them without `self.`.

packages/oauth/src/index.ts
seratch: As we dropped Node 10 support, perhaps, `Promise.allSettled` should be available in all supported Node runtimes.

seratch (commented):
I haven't run the code yet but left a few comments

packages/oauth/src/index.ts
misscoded: Both great suggestions, I'll make the changes!

misscoded: When I went to actually make this change, it became apparent that the second function _does_ rely on the instance (`client`, `clientId` and `clientSecret`). If there is a way to do as you've indicated that I'm not seeing, happy to hear it and adjust, but for the time being, I'll just make the name change and move the detection function to the top level.

stevengill (approved):

PR #1264, opened 2021-06-09
Fix #1250: Add Support for Axios timeout Customization

### LINKED ISSUE #1250: Expose axios timeout parameter
### Description

By default axios has no timeout, this is a bad default for production systems and currently there is no way to override it.
You can override the socket timeout with the agent but it's not the same thing.
I would like to be able to change that via the WebClient options.

- [ ] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

---

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

misscoded:
###  Summary

Fixes #1250.

Adds additional `timeout` prop to `WebClientOptions` to support custom Axios timeout value.
 
### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -117,7 +117,7 @@ describe('WebClient', function () {
 
   describe('has an option to override the Axios timeout value', function () {
     it('should log warning and throw error if timeout exceeded', function (done) {
-      const timeoutOverride = 15; // ms, guaranteed failure
+      const timeoutOverride = 1; // ms, guaranteed failure
       
       const logger = {
         debug: sinon.spy(),
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -115,6 +115,46 @@ describe('WebClient', function () {
     });
   });
 
+  describe('has an option to override the Axios timeout value', function () {
+    it('should log warning and throw error if timeout exceeded', function (done) {
+      const timeoutOverride = 15; // ms, guaranteed failure
+      
+      const logger = {
+        debug: sinon.spy(),
+        info: sinon.spy(),
+        warn: sinon.spy(),
+        error: sinon.spy(),
+        setLevel: sinon.spy(),
+        setName: sinon.spy(),
+      };
+      
+      const client = new WebClient(undefined, { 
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.spec.js
```diff
@@ -115,6 +115,46 @@ describe('WebClient', function () {
...
+        logLevel: LogLevel.WARN, 
+        logger 
+      });  
+      
+      client.apiCall('users.list')
+        .then(_ => {
+          done(new Error("expected timeout to throw error"));
+        })
+        .catch(error => {
+          try {
```
stevengill, line 143: Great job figuring out the nested try/catch to make the assert.equal failures actually display proper error messages!

stevengill (approved):
Looks great

packages/web-api/src/WebClient.spec.js
misscoded: So happy we found a solution to that issue. Couldn't leave us in the dark with those!

packages/web-api/src/WebClient.spec.js
```diff
@@ -115,6 +115,46 @@ describe('WebClient', function () {
     });
   });
 
+  describe('has an option to override the Axios timeout value', function () {
+    it('should log warning and throw error if timeout exceeded', function (done) {
+      const timeoutOverride = 15; // ms, guaranteed failure
```
seratch, line 120: If there is no reason to choose 15 milliseconds here, making it as small as possible (say, one millisecond, I don't know it works, though) may be safer. Perhaps, 15 also works in your environment and may work with the current builds. But I have been observing that GitHub Actions runtime sometimes behaves differently.

seratch (approved):
Good work (happy to have unit tests for the change)! If the CI builds pass, you can merge the change.

packages/web-api/src/WebClient.spec.js
misscoded: You were right -- I wouldn't have thought that it would make a difference but reducing it to 1ms eliminating the failing tests. Thanks for sharing this insight!

PR #1273, opened 2021-06-18
Make deleteInstallation optional

misscoded:
###  Summary

Post-merge feedback pointed out that `deleteInstallation` not being optional will cause issues for folks that have introduced their own stores already. Designating it as optional until next major version (v3.0).


stevengill (approved):
LGTM! Thanks for the quick fix

PR #1274, opened 2021-06-19
Add `SlashCommand` type to @slack/types package

samchungy:
###  Summary

- Adds the `SlashCommand` type from the [bolt-js](https://github.com/slackapi/bolt-js/blob/22f938e53ccef59e5a4acdd7b4d2a271430dd957/src/types/command/index.ts) project to `@slack/types` package.

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -401,3 +401,22 @@ export interface CallUserExternal {
   display_name: string;
   avatar_url: string;
 }
+
+export interface SlashCommand extends Record<string, any> {
+  token: string;
+  command: string;
+  text: string;
+  response_url: string;
+  trigger_id: string;
+  user_id: string;
+  user_name: string;
+  team_id: string;
+  team_domain: string;
+  channel_id: string;
+  channel_name: string;
+  api_app_id: string;
+  enterprise_id?: string;
+  enterprise_name?: string;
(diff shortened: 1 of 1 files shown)


packages/types/src/index.ts
```diff
@@ -401,3 +401,22 @@ export interface CallUserExternal {
   display_name: string;
   avatar_url: string;
 }
+
+export interface SlashCommand extends Record<string, any> {
```
samchungy, line 405:
```suggestion
export interface SlashCommand extends Record<string, unknown> {
```
Unknown may be better, but this is just keeping it the same as https://github.com/slackapi/bolt-js/blob/22f938e53ccef59e5a4acdd7b4d2a271430dd957/src/types/command/index.ts

stevengill:
Hey @samchungy!

Thanks for sending a PR.

This looks like you just copied the type from the [bolt-js library](https://github.com/slackapi/bolt-js/blob/main/src/types/command/index.ts). This `@slack/types` library doesn't currently contain any payload types from slack. It instead contains type definitions of block kit elements and other things. 

Is there a specific reason you want the command payload type in this package instead of just referencing it from the bolt package?

samchungy:
Hmm interesting, I guess I thought that all `types` would live in a `types` library, but I'm fine with importing that too

PR #1279, opened 2021-07-02
Add Support for FileInstallationStore

### LINKED ISSUE #1278: Add support for FileInstallationStore
### Description

Currently, the default (and only) `InstallationStore` made available to developers out of the box is the `MemoryInstallationStore`. This store isn't ideal for all development scenarios, as data is stored in memory and thus cleared out when the server is restarted.

We should introduce a `FileInstallationStore`, reasonably following the existing pattern(s) found on the Python and Java SDK side of things.

- [ ] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

misscoded:
###  Summary

Fixes #1278:

> Currently, the default (and only) InstallationStore made available to developers out of the box is the MemoryInstallationStore. This store isn't ideal for all development scenarios, as data is stored in memory and thus cleared out when the server is restarted.
>
> We should introduce a FileInstallationStore, reasonably following the existing pattern(s) found on the Python and Java SDK side of things.

___

#### Exportability for Choice
Though `MemoryInstallationStore` remains the default if not otherwise specified, to allow developers the choice between using `MemoryInstallationStore` and `FileInstallationStore`, both are now exported for direct use in their projects (set to the `installationStore` constructor option):

```
const { InstallProvider, MemoryInstallationStore, FileInstallationStore } = require('@slack/oauth');

const installer = new InstallProvider({
  clientId: process.env.SLACK_CLIENT_ID,
  clientSecret: process.env.SLACK_CLIENT_SECRET,
  authVersion: 'v2',
  stateSecret: 'my-state-secret',
  // installationStore: new MemoryInstallationStore(),
  installationStore: new FileInstallationStore({ baseDir: '.', historicalDataEnabled: false }),
});
```

Options include: `baseDir`, `clientId`, and `historicalDataEnabled`. Developers can go with the defaults, or override to suit their needs.

#### Installation Directory + File Pattern
For the most part, implementation has stayed true to what is found in Python, but some differences are detailed below:
1. The `enterprise_id` and `team_id` are used to create the installation directory. If both present, the pattern is `enterprise_id-team_id`. If only one of the two are present, _only_ that ID is used for the directory. This differs slightly from Python's use of `none` in the directory name if one is not passed.
2. When `storeInstallation` is called, if `historicalDataEnabled` is enabled (which it is by default), 4 records will be created: an `app-latest`, `user-[user_id]-latest`, `app-[currentUTC]` and `user-[user_id]-[currentUTC]`. Every subsequent install will overwrite the data in latest and save two additional `currentUTC` records. If set to `false` by the developer, only `*-latest` records will be saved.

 > Screen Shot 2021-06-30 at 6 40 03 PM

3. Python goes a step further by saving the same information into both `installer-latest` / `installer-[currentUTC]` _and_ the `user-[user_id]-*` records. I omitted the former records because it wasn't clear to me how the duplication was value-adding, but am open to introducing it if folks feel it's important or if I've misunderstood its purpose.
4. Python also makes use of a `to_bot()` method to shape the data prior to writing it to file, but for a similar reason to the above, I forwent that. Also equally happy to introduce it if it's the better (or necessary) approach.

**Directory/File Restructuring**
1. **New `/stores` directory.**  Since we now have two stores, I've pulled out the `MemoryInstallationStore` class into a `stores` directory, alongside the new `FileInstallationStore`. These two are then made available in a dedicated `index.js` so that the `import` is cleaner (both importable directly from `/stores`).

> Screen Shot 2021-07-01 at 5 12 57 PM

> Screen Shot 2021-07-01 at 5 14 01 PM

### DIFF
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -204,7 +204,7 @@ const installer = new InstallProvider({
         // org wide app installation deletion
         return await myDB.delete(installQuery.enterpriseId);
       }
-      if (query.teamId !== undefined) {
+      if (installQuery.teamId !== undefined) {
         // single team app installation deletion
         return await myDB.delete(installQuery.teamId);
       }
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -185,11 +185,11 @@ const installer = new InstallProvider({
     // returns installation object from database
     fetchInstallation: async (installQuery) => {
       // replace myDB.get with your own database or OEM getter
-      if (query.isEnterpriseInstall && query.enterpriseId !== undefined) {
+      if (installQuery.isEnterpriseInstall && installQuery.enterpriseId !== undefined) {
         // org wide app installation lookup
         return await myDB.get(installQuery.enterpriseId);
       }
-      if (query.teamId !== undefined) {
+      if (installQuery.teamId !== undefined) {
         // single team app installation lookup
         return await myDB.get(installQuery.teamId);
       }
@@ -200,7 +200,7 @@ const installer = new InstallProvider({
     // returns nothing
     deleteInstallation: async (installQuery) => {
       // replace myDB.get with your own database or OEM getter
-      if (query.isEnterpriseInstall && query.enterpriseId !== undefined) {
+      if (installQuery.isEnterpriseInstall && installQuery.enterpriseId !== undefined) {
         // org wide app installation deletion
         return await myDB.delete(installQuery.enterpriseId);
       }
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -153,9 +153,11 @@ app.get('/slack/oauth_redirect', (req, res) => {
 
 ### Storing installations in a database
 
-Although this package uses a default `MemoryInstallationStore`, it isn't recommended for production use since the access tokens it stores would be lost when the process terminates or restarts. Instead, `InstallProvider` has an option for supplying your own installation store, which is used to save and retrieve install information (like tokens) to your own database.
+Although this package uses a default `MemoryInstallationStore`, it isn't recommended for production purposes since the access tokens it stores are lost when the process terminates or restarts. To override this default, `InstallProvider` allows for supplying your own `installationStore`, which is then used to save and retrieve installation information (like tokens) to your own database. 
 
-An installation store is an object that provides two methods: `storeInstallation`, and `fetchInstallation`. `storeInstallation` takes an `installation` as an argument, which is an object that contains all installation related data (like tokens, teamIds, enterpriseIds, etc). `fetchInstallation` takes in a `installQuery`, which is used to query the database. The `installQuery` can contain `teamId`, `enterpriseId`, `userId`, `conversationId` and `isEnterpriseInstall`. 
+For more persistent storage during development, `FileInstallationStore` is a provided alternative to `MemoryInstallationStore` and is available for import and use directly from the package. Customizable options for this store include specifying the `baseDir`, `clientId`, and `historicalDataEnabled`.
+
+An installation store is an object that provides three methods: `storeInstallation`, `fetchInstallation`, and `deleteInstallation`. `storeInstallation` takes an `installation` as an argument, which is an object that contains all installation related data (like tokens, teamIds, enterpriseIds, etc). `fetchInstallation` and `deleteInstallation` both take in an `installQuery`, which is used to query the database. The `installQuery` can contain `teamId`, `enterpriseId`, `userId`, `conversationId` and `isEnterpriseInstall`. 
 
 In the following example, the `installationStore` option is used and the object is defined in line. The methods are implemented by calling an example database library with simple get and set operations.
 
@@ -192,6 +194,21 @@ const installer = new InstallProvider({
         return await myDB.get(installQuery.teamId);
       }
       throw new Error('Failed fetching installation');
+    },
+    // takes in an installQuery as an argument
+    // installQuery = {teamId: 'string', enterpriseId: 'string', userId: 'string', conversationId: 'string', isEnterpriseInstall: boolean};
+    // returns nothing
+    deleteInstallation: async (installQuery) => {
+      // replace myDB.get with your own database or OEM getter
+      if (query.isEnterpriseInstall && query.enterpriseId !== undefined) {
+        // org wide app installation deletion
+        return await myDB.delete(installQuery.enterpriseId);
+      }
+      if (query.teamId !== undefined) {
+        // single team app installation deletion
+        return await myDB.delete(installQuery.teamId);
+      }
+      throw new Error('Failed to delete installation');
     },
   },
 });
--- a/packages/oauth/src/index.ts
+++ b/packages/oauth/src/index.ts
@@ -11,6 +11,7 @@ import {
 } from './errors';
 import { parse as parseUrl, URLSearchParams, URL } from 'url';
 import { Logger, LogLevel, getLogger } from './logger';
+import { MemoryInstallationStore } from './stores';
 
 /**
  * InstallProvider Class.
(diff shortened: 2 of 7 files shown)


packages/oauth/src/index.ts
```diff
@@ -455,93 +456,9 @@ export interface InstallationStore {
     installation: Installation<AuthVersion, boolean>,
     logger?: Logger): Promise<void>;
   fetchInstallation:
-    (query: InstallationQuery<boolean>, logger?: Logger) => Promise<Installation<'v1' | 'v2', boolean>>;
+  (query: InstallationQuery<boolean>, logger?: Logger) => Promise<Installation<'v1' | 'v2', boolean>>;
   deleteInstallation:
-    (query: InstallationQuery<boolean>, logger?: Logger) => Promise<void>;
-}
-
-// using a javascript object as a makeshift database for development
```
misscoded, line 463: Never fear, this code has just been relocated to `/stores/memory-store.ts` below. 👇🏼

packages/oauth/src/stores/file-store.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (logger !== undefined) { logger.warn('Storing installation in FileInstallationStore'); }
+
+    // Create Installation Directory
+    fs.mkdir(installationDir, { recursive: true }, (err) => {
+      if (err !== null) { return console.error(err); }
+    });
+
+    try {
+      writeToFile(`${installationDir}/app-latest`, installationData);
+      writeToFile(`${installationDir}/user-${user.id}-latest`, installationData);
```
misscoded, line 41: Is there ever a case where `user` would not be present in the installation payload?

packages/oauth/src/stores/file-store.ts
```diff
@@ -0,0 +1,116 @@
...
+    const installationDir = this.getInstallationDir(enterprise?.id, team?.id);
+
+    if (logger !== undefined) { logger.warn('Storing installation in FileInstallationStore'); }
+
+    // Create Installation Directory
+    fs.mkdir(installationDir, { recursive: true }, (err) => {
+      if (err !== null) { return console.error(err); }
+    });
+
+    try {
```
misscoded, line 39: I mentioned this in the summary, but this is where it's applicable, so pasting:

> The Python FileStoreInstallation goes a step further by saving the same information into both installer-latest / installer-[currentUTC] and the user-[user_id]-* records. I omitted the former records because it wasn't clear to me how the duplication was value-adding, but am open to introducing it if folks feel it's important or if I've misunderstood its purpose.

packages/oauth/src/stores/file-store.ts
```diff
@@ -0,0 +1,116 @@
...
+      writeToFile(`${installationDir}/app-latest`, installationData);
+      writeToFile(`${installationDir}/user-${user.id}-latest`, installationData);
+
+      if (this.historicalDataEnabled) {
+        const currentUTC = Date.now();
+        writeToFile(`${installationDir}/app-${currentUTC}`, installationData);
+        writeToFile(`${installationDir}/user-${user.id}-${currentUTC}`, installationData);
```
misscoded, line 46: I mentioned this in the summary, but this is where it's applicable, so carrying it over:

> The Python FileStoreInstallation goes a step further by saving the same information into both installer-latest / installer-[currentUTC] and the user-[user_id]-* records. I omitted the former records because it wasn't clear to me how the duplication was value-adding, but am open to introducing it if folks feel it's important or if I've misunderstood its purpose.

packages/oauth/src/stores/file-store.ts
```diff
@@ -0,0 +1,116 @@
...
+      }
+    } catch (err) {
+      throw new Error('Failed to save installation to FileInstallationStore');
+    }
+  }
+
+  public async fetchInstallation(query: InstallationQuery<boolean>, logger?: Logger): Promise<Installation> {
+    const { enterpriseId, teamId } = query;
```
misscoded, line 54: `MemoryInstallationStore` doesn't take into consideration `user_id`, so I omitted it here, but now that I'm reading over it in this context.. 

Given that `user.id` has been introduced in `storeInstallation` above, my gut feeling is that it needs to be introduced here, as well, but confirmation of this from others would be useful.

packages/oauth/src/stores/file-store.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  private getInstallationDir(enterpriseId = '', teamId = ''): string {
+    let installDir = `${this.baseDir}/${enterpriseId}`;
+
+    if (teamId !== '') {
+      installDir += (enterpriseId !== '') ? `-${teamId}` : `${teamId}`;
+    }
+
+    return installDir;
```
misscoded, line 98: Different to the Python implementation, installation directories are either a `enterprise_id-team_id`, `enterprise_id` or `team_id` format. I don't think this has any unintended consequences, but another point I specifically would like confirmation on.

packages/oauth/src/stores/index.ts
```diff
@@ -0,0 +1,2 @@
+export { MemoryInstallationStore } from './memory-store';
```
misscoded, line 1: Both exported here to clean up the import, ie:

`import { MemoryInstallationStore, FileInstallationStore } from './stores';`

packages/oauth/src/index.ts
```diff
@@ -782,3 +699,4 @@ interface AuthTestResult extends WebAPICallResult {
 }
 
 export { Logger, LogLevel } from './logger';
+export { MemoryInstallationStore, FileInstallationStore } from './stores';
```
misscoded, line 702: Both now exported for use directly in projects so that the developer might import that which they prefer to work with.

packages/oauth/src/stores/file-store.ts
seratch: Specifically talking about this file based implementation, not having all the historical data is also fine as most developers won't use it for production.

The reason why we store all installations in Python / Java is that having the data can be helpful for troubleshooting / system operations in production. When you do investigation about an issue related to a user who installed the app multiple times, checking each installation's timestamp, the difference on its scopes, and the existence of enterprise_id may help.

If we think this file based one is a kind of reference implementation, having this consideration may be worth considering. Not a strong opinion here, though. Just having a comment about this context there is also fine.

packages/oauth/src/stores/file-store.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.baseDir = clientId !== undefined ? `${baseDir}/${clientId}` : baseDir;
+    this.historicalDataEnabled = historicalDataEnabled;
+  }
+
+  public async storeInstallation(installation: Installation, logger?: Logger): Promise<void> {
+    const { enterprise, team, user } = installation;
+    const installationData = JSON.stringify(installation);
+    const installationDir = this.getInstallationDir(enterprise?.id, team?.id);
+
+    if (logger !== undefined) { logger.warn('Storing installation in FileInstallationStore'); }
```
seratch, line 32: To me, info level would be more reasonable here as this is not an error or failure. Also, having useful data like enterprise_id / team_id / user_id in the message would be a good addition (needless to say, we should not have tokens / webhook URLs). Just having this message does not add so much value.

packages/oauth/src/stores/file-store.ts
seratch: How about having TODO comments on user token scope installation support?

packages/oauth/src/stores/file-store.ts
seratch: That's fine but the naming rules are not the same, I would recommend using a different base directory to avoid mixing different patterns in the same place. In other words, instead of `.bolt-app-installation`, you can use `.bolt-js-app-installation` or something similar. 

FYI, the Java SDK uses [`.slack-app`](https://github.com/slackapi/java-slack-sdk/blob/v1.8.1/bolt/src/main/java/com/slack/api/bolt/service/builtin/FileInstallationService.java#L24). We can revisit these naming if we would like to do so but for now, I think using different names would be fine.

packages/oauth/src/stores/index.ts
seratch: hm, do we need this file? Just importing the two in `src/index.ts` sounds good enough to me.

packages/oauth/src/index.ts
seratch: How about doing this instead? No need to modify when we add more builtin store implementations.
```typescript
export * from './stores';
```

seratch (commented):
I haven't verified if it works fine but at first glance, the changes look good to me. Let a few comments.

seratch:
This pull request seems to hav conflicts with the latest revision in main branch. Resolving it before you finalize this work would make things easier.

packages/oauth/src/stores/file-store.ts
misscoded: Totally get it and am with you regarding the importance. 👍🏼  

I think what I'm not understanding is the difference between [these two records](https://github.com/slackapi/python-slack-sdk/blob/2a4487c81fa95d1cb07a3d916887ac10d67d79ab/slack_sdk/oauth/installation_store/file/__init__.py#L52-L63), other than the extra formatting done to the first, and why having both is necessary or beneficial.

docs/_packages/oauth.md
```diff
@@ -183,16 +185,31 @@ const installer = new InstallProvider({
...
+    // takes in an installQuery as an argument
+    // installQuery = {teamId: 'string', enterpriseId: 'string', userId: 'string', conversationId: 'string', isEnterpriseInstall: boolean};
+    // returns nothing
+    deleteInstallation: async (installQuery) => {
+      // replace myDB.get with your own database or OEM getter
+      if (installQuery.isEnterpriseInstall && installQuery.enterpriseId !== undefined) {
+        // org wide app installation deletion
+        return await myDB.delete(installQuery.enterpriseId);
+      }
+      if (query.teamId !== undefined) {
```
misscoded, line 207: Missed this one. Note to myself to not forget to resolve this.

packages/oauth/src/stores/file-store.ts
misscoded: All good points. I did carry this over from the memory store, but did away with the true 'warning' part of it (recommending that they use a _real_ store for production purposes).

If using the `FileInstallationStore` is not something we'd recommend for production, it may be worth adding that back into the message and keeping it as a warning to remind them. Thoughts?

misscoded: Perhaps something like the following:

```
if (logger !== undefined) {
  logger.info(`Deleting installations in ${installationDir} using the following query: ${query}`);
  logger.warn('FileInstallationStore is not intended for production purposes.');
}
```

packages/oauth/src/stores/file-store.ts
stevengill: Should we consider setting saving historical data default to false? Users could set it to true if they want to save historical data.

packages/oauth/src/stores/file-store.ts
seratch: How about having the warning log only for this store instance's initialization? I think repeatedly seeing this warning when local development is not a great developer experience.

packages/oauth/src/stores/file-store.ts
stevengill: @seratch so to support user tokens, it seems we would want to fetch `/user-{user_id}-latest` from the installationDirectory. How do you determine if you want to return this over `app-latest`? Do you look for the existence of `user_id` in the `query` and  `/user-{user_id}-latest` on the file system and return it over `app-latest`? How about if you were explicitly looking for a user token existing, in that scenario if `/user-{user_id}-latest` didn't exist, you would want to throw an error instead of getting `app-latest` which might contain a user token belonging to a different user.

seratch: @stevengill Performing an additional query would be necessary for it. Here are the ways we do in Python / Java.

* https://github.com/slackapi/bolt-python/blob/v1.6.1/slack_bolt/authorization/authorize.py#L137-L167
* https://github.com/slackapi/java-slack-sdk/blob/v1.8.1/bolt/src/main/java/com/slack/api/bolt/middleware/builtin/MultiTeamsAuthorization.java#L128-L138

stevengill: @seratch this is very helpful! So the idea being do two fetchinstallation calls in authorize (or do this logic in fetchInstallatino itself). Roughly, in this JS implementation, the first fetchInstallation call would grab `app-latest`. We would want to strip out `user_id` from query so it doesn't try to grab a `user-{user_id}-latest` instead. The second call to fetchInstallation we would pass in the userId and it would look for `user-{user_id}-latest`. It would pull out the user portion of the installation Object and attach it to the original installationObject we get back from `app-latest`. It covers some of the edge cases I was worried about. Nice solution.

misscoded: To limit the scope, I'm in favor of leaving a TODO to address the user token support at a later time if others are.

packages/oauth/src/stores/file-store.ts
misscoded: I don't have a strong opinion on this. Happy to go with whatever others deem the best default.

seratch: >I think what I'm not understanding is the difference between these two records, other than the extra formatting done to the first, and why having both is necessary or beneficial.

Even after revoking all the user installations, bot token must still exist. This is one of the major improvements in GBP (Granular Bot Permissions). This is the reason why Python / Java SDK have bots table separately. 

If there is a better alternative way, we can go with a different approach in Node SDK first and implement the same in others, though. Having `is_user_token_revoked` flag in the installation table plus trying to fetch a bot token even when all installations are marked as deleted may work. Do we think this is a better approach? 🤔 

>Should we consider setting saving historical data default to false?

In Python / Java SDK, the historical data support is enabled by default. For consistency and encouraging a similar approach for production app design, I prefer turning it on but I am open to change this default for Node SDK (and possibly for others in the future major versions).

seratch: @misscoded @stevengill For enabling the historical data support by default or not, I'm happy to leave the decision to you all. I've already shared my thoughts on it in the above comment. If we want to change the defaults for Python / Java, we may want to hold off doing so until the next major releases (we won't do at least this year for both).

For having bots and user installations, as I mentioned above, having something similar (or user_token_revocation flag) should be necessary to keep a bot token even after the user(s) who installed the app are deactivated or all the user tokens are revoked. For these two store implementations, I think it's okay not to support the use case.

seratch:
@misscoded 
I don't have any further comments apart from https://github.com/slackapi/node-slack-sdk/pull/1279#discussion_r668610400 

If we are comfortable to have this change in the next minor (v2.2 - the main update is the token rotation support), you can merge this pull request. Otherwise, you can move this PR and its issue to v2.3 milestone!

packages/oauth/src/stores/file-store.ts
misscoded: I think it makes sense to have historical data enabled as the default, especially to maintain parity between the three languages. So we're good on that point! I'll keep it as is.

I'd like to revisit the other point you bring up when I circle back around to sort out the lack of support for user tokens as a whole, unless you think we shouldn't move forward with introducing this until that's sorted. I think I'm still fuzzy on the details.

misscoded:
> If we are comfortable to have this change in the next minor (v2.2 - the main update is the token rotation support), you can merge this pull request. Otherwise, you can move this PR and its issue to v2.3 milestone!

I think we're good, just waiting on Steve to give the final go ahead on what's here.

stevengill (approved):

PR #1287, opened 2021-07-14
Add Support for Token Rotation

### LINKED ISSUE #1241: Add Support for Token Rotation

misscoded:
Fixes #1241

> Developers will soon have the ability to opt their apps into advanced token security via token rotation. We should introduce logic to incorporate these new tokens, as well as refresh behavior.

### HOW IT WORKS
On OAuth installation (the only installation flow allowed when token rotation is enabled), new access tokens are stored as they have been, but now include an accompanying `refresh_token` and `expires_in`. The `refresh_token` (does not expire) is used to "trade-in" for a new `access_token` and `refresh_token` pair. 

For each token (`bot`|`user`), we use the `expires_in` value to calculate and store a corresponding future expiry timestamp (UTC, seconds).

Each time an event is received, `authorize` gets run. Here we determine if token rotation is enabled. If so, we check the expiry of the token(s) and refresh those that have expired or will expire within 2 hours. We then update the returned `authResult` and the `Installation`.

### NOTABLE CHANGES
1. `AuthorizationResult` has four new keys: `botRefreshToken`, `botTokenExpiresAt`, `userRefreshToken`, and `userTokenExpiresAt`. This was to handle apps that have both User and Bot tokens while still keeping the shape flat.

2. `Installation` has four new keys: a `refreshToken` and `expiresAt` for both the `user` and `bot` objects.


packages/web-api/src/methods.ts
```diff
@@ -521,6 +521,7 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
     access: bindApiCall<OAuthAccessArguments, OauthAccessResponse>(this, 'oauth.access'),
     v2: {
       access: bindApiCall<OAuthV2AccessArguments, OauthV2AccessResponse>(this, 'oauth.v2.access'),
+      exchange: bindApiCall<OAuthV2ExchangeArguments, OauthV2AccessResponse>(this, 'oauth.v2.exchange'),
```
seratch, line 524: @misscoded Can you run `./scripts/generate-web-api-types.sh`? The command generates the response type for this endpoint. Also, it adds a few missing fields in other types.

seratch (approved):
Apart from the type definition re-generation and using the new one for `oauth.v2.exchange` API method, this pull request should be ready for merging

PR #1325, opened 2021-09-01
Replace TSLint with ESLint

filmaj:
Work In Progress!

###  Summary

This is a first go at introducing the ESLint configs we converged on in the bolt-js project to this project. Did you all think you would be free of thousands of lines of changes moving commas around and introducing newline characters? Well think again! 😈 

The initial pull request includes changes that will be core and applicable to all published packages making up this repository as well as changes to `packages/web-api/src` code to have the `web-api` package adhere to the new lint configuration. As a result, the most important changes to review include:

- `scripts/code_generator.rb`: changes to the type generator script (replacing `tslint:disable` with `eslint-disable`)
- Copies ESLint config files from bolt-js into a new `lint-configs/` directory
- What do people think of the approach of [symlinking](https://en.wikipedia.org/wiki/Symbolic_link) the eslint config files into each published package's directory? Any alternatives or better approaches that we should consider?
  - Related: because the eslint config files defining the rules pull in certain dependencies, this requires putting some of the eslint dependencies (like eslint itself and the jsdoc rules) into the top-level package.

Summary of changes so far:

- Updated type/code generator (`scripts/code_generator.rb`) to drop in an `eslint-disable` directive into generated code instead of a `tslint:disable` one.
- Factored out eslint-related configs into a top-level `lint-configs` directory.
- Set up symlinks to eslint configuration files between `packages/web-api` and `lint-configs` directories.

### Questions

- [x] Symlink approach good enough?
- [x] Should the top-level `package.json` differentiate between linting vs. testing each sub-project? i.e. `lerna run lint` and `lerna run test` being separate? Alternatively, can include linting in each sub-project's `test` npm run script, and drop the execution of `lerna run lint`. IMO the latter makes sense as passing the linter should be considered necessary - and thus linting is implicitly part of the test suite.

### TODO

- [x] Do an audit / search for `tslint:disable` directives
- [x] Review `.npmignore` files to make sure published packages don't include any of the new eslint config files
- [x] Instructions need to include running `npm i` from the root directory to ensure you have eslint-plugin-jsdoc installed


seratch:
Thanks for working on this! Looks good so far!

packages/oauth/src/index.ts
```diff
@@ -9,10 +10,31 @@ import {
   GenerateInstallUrlError,
   AuthorizationError,
 } from './errors';
-import { parse as parseUrl, URLSearchParams, URL } from 'url';
 import { Logger, LogLevel, getLogger } from './logger';
 import { MemoryInstallationStore } from './stores';
 
+// default implementation of StateStore
```
srajiang, line 16: Out of curiosity, what's the reasoning behind moving this up? Was this related to one of the eslint rules?

packages/oauth/src/index.ts
```diff
@@ -291,10 +320,10 @@ export class InstallProvider {
 
     try {
       if (req.url !== undefined) {
-        parsedUrl = parseUrl(req.url, true);
-        code = parsedUrl.query.code as string;
-        state = parsedUrl.query.state as string;
-        if (state === undefined || state === '' || code === undefined) {
```
srajiang, line 297: Wondering whether this was written this way to be a bit more explicit for people unfamiliar with the JS type coercion.

packages/oauth/src/index.ts
```diff
@@ -817,7 +823,7 @@ export interface OAuthV2TokenRefreshResponse extends WebAPICallResult {
   team: { id: string, name: string };
   enterprise: { name: string, id: string } | null;
   is_enterprise_install: boolean;
-  response_metadata: {}; // TODO
+  response_metadata: Record<string, never>; // TODO
```
srajiang, line 826: Is there a reason to keep this comment in this line now?

srajiang (approved):
Looking good! Left a couple of questions/ comments! Also, are you planning on committing up a `package-lock.json` with these changes?

packages/oauth/src/index.ts
filmaj: Yes, the 'no use before defined' rule. While functions are hoisted to the top by the runtime, classes, and I guess types as well, are not.

packages/oauth/src/index.ts
filmaj: I rewrote this part as use of `parseUrl` is deprecated and that triggered an eslint rule! So I did it the 'approved' newer-node-version way of creating a `new URL`.

packages/oauth/src/index.ts
filmaj: Good question! I think so but am not sure - `Record` seems like a barely-more-acceptable type than `{}`?

srajiang:
```suggestion
  response_metadata: Record<string, never>; // TODO revisit this type to see whether there are better alternatives
```

seratch (approved):
No other comments! Thanks for the massive efforts here 👍

packages/oauth/src/index.spec.js
```diff
@@ -398,7 +398,7 @@ describe('OAuth', async () => {
     });
 
     it('should call the failure callback due to missing state query parameter on the URL', async () => {
-      const req = { url: 'someUrl' };
+      const req = { url: 'http://someUrl.com' };
```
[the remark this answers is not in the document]
filmaj: Yes, good idea. In fact, this change is related to another change (see [my related review comment here](https://github.com/slackapi/node-slack-sdk/pull/1325#discussion_r705565947)). Because of the removal of the use of `parseUrl` (which is deprecated) and replacing it with `new URL()`, if the protocol and/or TLD is not provided in the URL string, it will cause an error. I wonder if this has farther-reaching consequences and as a result need to add more tests for this? For example, what if someone has custom domain mappings in their e.g. `hosts` file, to map e.g. `someUrl` to `127.0.0.1`. In the current implementation, perhaps it is perfectly OK to use a technically malformed URL (no protocol and/or TLD), whereas due to this change, that would not be acceptable.

What do you think?

packages/oauth/src/index.ts
filmaj: Sounds good, I have removed it in the most recent commit!

packages/oauth/src/index.spec.js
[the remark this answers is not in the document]
seratch: Thanks for sharing the context. My comment was not clear enough but I wanted to suggest `http://example.com` or `https://example.com` instead. If you click the URL in your editor, you will be navigated to the URL. Some people may unintentionally access the URLs. 

Although I don't have any thoughts on the domain itself, I think that we can always consider using the safest domains / URLs in test data. I'm not stick with example.com at all but it's commonly used. For this reason, it should be good enough.

PR #1329, opened 2021-09-09
Add optional InstallProvider#stateValidation flag

srajiang:
###  Summary
This PR is first part of addressing [bolt-js/#1101](https://github.com/slackapi/bolt-js/issues/1101). It adds the option to disable state verification during OAuth flow via a `stateVerification` flag in the InstallProvider. By default this will be true. 

Because state verification is a highly recommended security feature of Oauth, we opt to verify when state exists but when it doesn't fall back on whether stateVerification is on to determine what to do next. 

When `stateVerification` is set as false, in the event that state is not supplied in request params, the client will ignore the absence of state which is our desired behavior. But if state is supplied, the client presumes that is because this state param was generated on our side, and will attempt to verify as usual.

Another change here is that `handleCallback` will now optionally accept installOptions per Kaz's suggestion (there are some changes also on the receiver side that need to happen). 

Notes
- [x] I have not touched the URL generation / state generation flow.`stateSecret` is still required even with stateValidation disabled. So when Bolt is in charge of generating the URL it will still do so regardless of what the flag is set as. 
- [x] `package.json` isn't being found causing build to fail, need to come back to this. 
- [x] This should land after #1325 
- [ ] [bolt-js/pull/1116](https://github.com/slackapi/bolt-js/pull/1116) has a draft of the related changes needed receiver side once this lands

### DIFF
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -290,9 +290,9 @@ const installer = new InstallProvider({
 
 ### State validation
 
-By default, this package handles generating and validating a `state` parameter during OAuth installation. The state parameter helps to mitigate the risk of [CSRF attacks](https://auth0.com/docs/configure/attack-protection/state-parameters#csrf-attacks) and is strongly recommended. 
+By default, this package handles generating and validating a `state` parameter during OAuth installation. The state parameter helps to mitigate the risk of [Cross-Site Request Forgery](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12) and is strongly recommended. 
 
-In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state validation cannot be completed. In this case, you can disable validation via setting the `InstallProvider#stateValidation` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. It will attempt to validate `state` only if it is provided in the request url parameters. 
+In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state validation cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` validation via setting the `InstallProvider#stateValidation` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. However, it will attempt to validate `state` only if it is provided in the request url parameters. 
 
 ```javascript
 const installer = new InstallProvider({
@@ -303,6 +303,8 @@ const installer = new InstallProvider({
   },
 });
 ```
+
+Please note that while setting `stateValidation` false will allow installation to proceed when the `state` query parameter isn't provided, any metadata you provide via installation URL options will not be available.
 ---
 
 
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -288,11 +288,11 @@ const installer = new InstallProvider({
 ```
 ---
 
-### State validation
+### State verification
 
 By default, this package handles generating and verifying a `state` parameter during OAuth installation. The state parameter helps to mitigate the risk of [Cross-Site Request Forgery](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12) and is strongly recommended. 
 
-In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateValidation` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. However, it will attempt to verify `state` only if it is provided in the request url parameters. 
+In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. However, it will attempt to verify `state` only if it is provided in the request url parameters. 
 
 ```javascript
 const installer = new InstallProvider({
@@ -304,7 +304,7 @@ const installer = new InstallProvider({
 });
 ```
 
-Please note that while setting `stateValidation` false will allow installation to proceed when the `state` query parameter isn't provided, any metadata you provide via installation URL options will not be available.
+Please note that while setting `stateVerification` false will allow installation to proceed when the `state` query parameter isn't provided, any metadata you provide via installation URL options will not be available.
 ---
 
 
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -288,6 +288,24 @@ const installer = new InstallProvider({
 ```
 ---
 
+### State validation
+
+By default, this package handles generating and validating a `state` parameter during OAuth installation. The state parameter helps to mitigate the risk of [CSRF attacks](https://auth0.com/docs/configure/attack-protection/state-parameters#csrf-attacks) and is strongly recommended. 
+
+In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state validation cannot be completed. In this case, you can disable validation via setting the `InstallProvider#stateValidation` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. It will attempt to validate `state` only if it is provided in the request url parameters. 
+
+```javascript
+const installer = new InstallProvider({
+  clientId: process.env.SLACK_CLIENT_ID,
+  clientSecret: process.env.SLACK_CLIENT_SECRET,
+  stateSecret: 'my-state-secret',
+  stateValidation: false,
+  },
+});
+```
+---
+
+
 ### Setting the log level and using a custom logger
 
 The `InstallProvider` will log interesting information to the console by default. You can use the `logLevel` to decide how
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -288,26 +288,6 @@ const installer = new InstallProvider({
 ```
 ---
 
-### State verification
-
-By default, this package handles generating and verifying a `state` parameter during OAuth installation. The state parameter helps to mitigate the risk of [Cross-Site Request Forgery](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12) and is strongly recommended. 
-
-In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. However, it will attempt to verify `state` only if it is provided in the request url parameters. 
-
-```javascript
-const installer = new InstallProvider({
-  clientId: process.env.SLACK_CLIENT_ID,
-  clientSecret: process.env.SLACK_CLIENT_SECRET,
-  stateSecret: 'my-state-secret',
-  stateVerification: false,
-  },
-});
-```
-
-Please note that while setting `stateVerification` false will allow installation to proceed when the `state` query parameter isn't provided, any metadata you provide via installation URL options will not be available.
----
-
-
 ### Setting the log level and using a custom logger
 
 The `InstallProvider` will log interesting information to the console by default. You can use the `logLevel` to decide how
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -254,7 +254,7 @@ const result = await installer.installationStore.fetchInstallation({teamId:'my-t
 
 ### Using a custom state store
 
-A state store handles generating the OAuth `state` parameter in the installation URL for a given set of options, and verifying the `state` in the OAuth callback and returning those same options.
+A state store handles generating the OAuth `state` parameter in the installation URL for a given set of options, and validating the `state` in the OAuth callback and returning those same options.
 
 The default state store, `ClearStateStore`, does not use any storage. Instead, it signs the options (using the `stateSecret`) and encodes them along with a signature into `state`. Later during the OAuth callback, it verifies the signature.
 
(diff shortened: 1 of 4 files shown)


packages/oauth/src/errors.ts
```diff
@@ -25,6 +26,10 @@ export class MissingStateError extends Error implements CodedError {
   public code = ErrorCode.MissingStateError;
 }
 
+export class MissingCodeError extends Error implements CodedError {
```
srajiang, line 29: Added a new specific error for case when `code` param is omitted. Previously this was covered under `MissingStateError`, but I felt it made sense to split these failure states up.

packages/oauth/src/index.ts
```diff
@@ -495,7 +508,7 @@ export interface CallbackOptions {
   // callbackRes.
   success?: (
     installation: Installation | OrgInstallation,
-    options: InstallURLOptions,
+    options: InstallURLOptions | undefined,
```
srajiang, line 511: Anyone see any issues will allowing options here to be undefined? It appears in the built-in success and failure callbacks these options are referenced but not used.

packages/oauth/src/index.spec.js
```diff
@@ -417,8 +417,26 @@ describe('OAuth', async () => {
       assert.isTrue(sent);
     });
 
+    it('should call success callback when state query param is missing but stateValidation disabled', async () => {
+      const req = { url: 'http://someUrl.com?code=1234' };
```
seratch, line 421: Let's use example.com for safety (I don't have any comments about the used domain but just in case)
```suggestion
      const req = { url: 'http://example.com?code=1234' };
```

packages/oauth/src/index.ts
```diff
@@ -60,7 +63,10 @@ export class InstallProvider {
     } else {
       this.logger = getLogger('OAuth:InstallProvider', logLevel ?? LogLevel.INFO, logger);
     }
-
+    this.stateValidation = stateValidation;
+    if (!stateValidation) {
+      this.logger.info('You\'ve disabled mandatory state validation during oauth. When state is provided as a query param we will attempt to validate, when it is omitted, with this flag set to true, we will ignore validation. Custom metadata options will not be available when state is not supplied.');
```
seratch, line 68: How about using double-quotes? With the way, you don' need to escape single-quotes.

seratch: s/oauth/OAuth/

seratch: >with this flag set to true, 

Perhaps, the log message readers do not understand this part (What's "this flag"???). We can be more specific. If I try to write the message, I will go with something like this:

You've disabled OAuth's state parameter validation by setting `InstallProvider#stateValidation` to false. The option is supposed to be used only for Org-Wide app installations from admin pages. If this is not the case, we highly recommend setting the flag to true and starting your OAuth flow from `/slack/install` or your own starting endpoint. The state parameter can protect your end-users from CSRF attacks. Refer to https://datatracker.ietf.org/doc/html/rfc6749#section-10.12 for more details.

I know this is too long. We can simplify the sentence, though.

packages/oauth/src/index.ts
```diff
@@ -314,7 +327,7 @@ export class InstallProvider {
           code,
           client_id: this.clientId,
           client_secret: this.clientSecret,
-          redirect_uri: installOptions.redirectUri,
+          redirect_uri: installOptions ? installOptions.redirectUri : undefined,
```
seratch, line 330: If an app disables state validation while using a redirect_uri, this API call may fail. right? We may want to add `installOptions` as an optional argument of `handleCallback` method for it (I haven't verified how it works well yet, though).

packages/oauth/src/index.ts
seratch: This should be fine but changing the type can be a breaking change for existing TypeScript app maintainers. Instead of passing `undefined`, passing an empty `InstallURLOptions` object may be an alternative. This PR does not change the existing app behavior. No existing apps don't receive the empty value.

packages/oauth/src/index.ts
srajiang: @seratch From my read of things, it appears that installOptions is an object encoded in the `state` param. So I'm not sure how it could be passed to `handleCallback` on the HTTPReceiver side. Is that your understanding as well?

packages/oauth/src/index.ts
srajiang: Good point, the wording is a bit confusing at the moment. Let me reword with your feedback in mind and see whether we can make it shorter. I will also add a section to SDK documentation.

seratch: Thanks!

packages/oauth/src/index.ts
seratch: @srajiang I haven't verified if this works yet but I was thinking that we can have the following change on the bolt-js side.

```diff
diff --git a/src/receivers/HTTPReceiver.ts b/src/receivers/HTTPReceiver.ts
index 71188f69..9114ec3c 100644
--- a/src/receivers/HTTPReceiver.ts
+++ b/src/receivers/HTTPReceiver.ts
@@ -435,12 +435,17 @@ export default class HTTPReceiver implements Receiver {
     // when installer is defined then installCallbackOptions is always defined too.
     const [installer, installCallbackOptions] = [this.installer!, this.installCallbackOptions!];
 
-    installer.handleCallback(req, res, installCallbackOptions).catch((err) => {
+    const errorHandler = (err: Error) => {
       this.logger.error(
         'HTTPReceiver encountered an unexpected error while handling the OAuth install redirect. Please report to the maintainers.',
       );
       this.logger.debug(`Error details: ${err}`);
-    });
+    };
+    if (installer.stateValidation) {
+      installer.handleCallback(req, res, installCallbackOptions).catch(errorHandler);
+    } else {
+      installer.handleCallback(req, res, installCallbackOptions, this.installUrlOptions).catch(errorHandler);
+    }
   }
 }
 ```

File: packages/oauth/src/index.ts (hunk 2 above)
Comment (srajiang) [reply packages/oauth/src/index.ts]: That makes sense, I've updated this to pass an empty `InstallURLOptions` object instead of passing undefined.

File: docs/_packages/oauth.md [hunk 6]
```diff
@@ -288,6 +288,26 @@ const installer = new InstallProvider({
 ```
 ---
 
+### State validation
```
srajiang, line 291: I noticed that there's some inconsistency between other SDK which use the word(s) `validate, validation` and Bolt JS, where we use `verify, verifyStateParam()` etc. If others feel like being internally consistent with existing `verifyStateParam` method is more important than being consistent with Bolt Python etc, I can make modifications to change everything to `verify`

seratch: Good point. As the `verifyStateParam` method is public, I'm fine to use `verify` over `validate` only in bolt-js. To avoid having breaking changes, I prefer using validate for other SDKs.

seratch (changes requested):
@srajiang Can you move these document changes to a new pull request? With this way, it's easier for us to update the documents after releasing a new version including the change.

packages/oauth/src/index.ts
srajiang: I noticed that
1) `installUrlOptions` [doesn't contain ](https://github.com/slackapi/bolt-js/blob/ee18b1c7d7b5c80a0fcce66b5c48a8235f2417ec/src/receivers/HTTPReceiver.ts#L170-L174)`redirectUrl` field. This is the same object being [passed to](https://github.com/slackapi/bolt-js/blob/ee18b1c7d7b5c80a0fcce66b5c48a8235f2417ec/src/receivers/HTTPReceiver.ts#L464) `generateInstallUrl` and encoded as `state`. Then later on in handleCallback, we are trying to access redirectUri (for example [here](https://github.com/slackapi/node-slack-sdk/blob/66066ee4dfa069c4829bc239b33a98a87cb1c009/packages/oauth/src/index.ts#L351)) from the installOptions, but actually it's always undefined. This is a bug I can fix on the HTTPReceiver side.

2) `installUrlOptions` options object contains fields set during App initialization which are static (including even `metadata`) so they don't appear to change between installations. `installURLOptions` can always be available as a parameter to `handleCallback()`, and we could still verify `state` as planned (it throws an error if state is invalid). It's a different design than the original, but having options passed around as state seems to only make sense if options are dynamic. @stevengill re: our chat

packages/oauth/src/index.ts
```diff
@@ -312,30 +319,43 @@ export class InstallProvider {
...
     let parsedUrl;
     let code: string;
     let flowError: string;
     let state: string;
-    let installOptions: InstallURLOptions;
-
     try {
       if (req.url !== undefined) {
-        parsedUrl = new URL(req.url);
+        parsedUrl = new URL(req.url, `http://${req.headers.host}`);
```
srajiang, line 330: The new recommended way to url parse via URL constructor is less lenient about what it will accept than the previous url.parse(). So partial urls don't work any more.

packages/oauth/src/index.ts
srajiang: Alrighty, maybe 2. is too drastic of a change - I've made changes to support passing in installOptions to `handleCallback()`

packages/oauth/src/errors.ts
seratch: This is a nice improvement!

packages/oauth/src/index.ts
seratch: The URL should be `https://`. We don't allow using `http://` in production OAuth flows. Also, does this approach work well in `X-Forwarded-Host` header situations too?

packages/oauth/src/index.ts
seratch: @srajiang Thanks for checking these! 

Regarding the point 1., we should fix the bug for all the built-in receivers.

As for 2., perhaps, we can revisit the issues around the metadata. A static value for metadata is not useful at all. We should consider adding a way to set metadata per request. Also, some developers may want to switch redirect_uri per requests in some cases. 

For your information, in Java / Python SDKs, I have been encouraging to use browser cookies to tell the OAuth callback (=redirect_uri) the state/user tracking data. Perhaps, something similar rather than embedding value in the state parameter would be more robust for any use cases.

packages/oauth/src/errors.ts
filmaj: Can we add a test that checks for this new error? We used to have it but this PR seems to have re-used it for state validation - can we re-add it back?

filmaj (commented):
LGTM, though @seratch raises a good point re: proxy-originating behaviour via the `X-Forwarded-Host` header, and I believe we also lost some test coverage regarding the new `MissingCode` error - would be good to get that back.

packages/oauth/src/index.ts
```diff
@@ -88,7 +93,10 @@ export class InstallProvider {
     } else {
       this.logger = getLogger('OAuth:InstallProvider', logLevel ?? LogLevel.INFO, logger);
     }
-
+    this.stateVerification = stateVerification;
+    if (!stateVerification) {
+      this.logger.info("You've set InstallProvider#stateVerification to false. This flag is intended to enable org-wide app installations from admin pages. If this isn't your scenario, we recommend setting stateVerification to true and starting your OAuth flow from the provided `/slack/install` or your own starting endpoint.");
```
stevengill, line 98: nit: maybe we should use `this.logger.info` should be `this.logger.warn`

packages/oauth/src/index.ts
srajiang: Hey @seratch! OK is it fair to say that when the request contains x-forwarded-host, we would always prefer to take the host value from there?

packages/oauth/src/errors.ts
srajiang: Yep, added!

packages/oauth/src/index.ts
seratch: @srajiang If you would like to know the actual redirect URI here, you need to use `x-forwarded-` headers. Speaking of this part, whether this line of code works well also for the situation depends on how the Node's `URL` module behaves. Before going with this implementation, we should do testing with a reverse proxy. It seems that we are using only `searchParams` so far. As long as the URL initialization does not fail, it should work, though.

By the way, is this code taken from Node's official recommendation?

srajiang: I agree that we should do testing with reverse proxy; I would want to know how the `x-forwarded-` header is  accessed from the `req` object.  

Re: Is this code taken from Node's offical recommendation - I'm not sure exactly what you mean. I've been referencing [these docs](https://nodejs.org/api/url.html#url_new_url_input_base). Based on that, it seems that if we provide the correc original host, it should work.

Maybe I am missing something here?

seratch: Thanks. As `http://${req.headers.host}` is not a good-enough solution, I was curious about the reason why I chose the way. To properly build the `base` part of the `URL` constructor arguments, you can use the pair of `x-forwarded-proto` and `x-forwarded-host` in the case where they exist in `req`. Perhaps, we want to have an internal method that builds the `base` from `req` object.

seratch: @srajiang and I discussed the best way to implement this and we decided to have a private method that returns only searchParams from `req` as we don't use any other properties in `req` here.

packages/oauth/src/index.ts
```diff
@@ -312,30 +319,41 @@ export class InstallProvider {
...
     let parsedUrl;
     let code: string;
     let flowError: string;
     let state: string;
-    let installOptions: InstallURLOptions;
-
     try {
       if (req.url !== undefined) {
-        parsedUrl = new URL(req.url);
+        parsedUrl = new URL(req.url, `https://${req.headers.host}`);
```
seratch, line 330: @srajiang Can you move this to a private method like `const searchParams = extractSearchParams(req)` or have some comment saying that protocol/host in the `parsedUrl` may not be accurate here?

seratch (approved):
@srajiang One last comment about the remaining task. After resolving it, you can merge this PR but don't forget squashing commits into one 👋

packages/oauth/src/index.ts
filmaj: BTW for context on why the URL assembly changed: it was part of the move to ESLint. ESLint will warn (via the [`node/no-deprecated-api` rule](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-deprecated-api.md)) if you use the [now-deprecated node method `require('url').parse`](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost). ESLint instead recommends to construct the URL using the `new` keyword.

Issue #824, opened 2019-07-16
queue all calls to `respond()` by default


shaydewael:
Reopened from https://github.com/slackapi/node-slack-interactive-messages/issues/38

### Description

Currently, if you call `respond()` twice successively, there is no ordering guarantee on how the messages in Slack will update. If this is within an interactive message (button or menu), one call will overwrite the other but its a race condition. If its within a dialog submission (or a button or menu using the `replace_original: false` argument) then two messages will be created but in an arbitrary order.

The only way to rely on the ordering would be to wait on the first call's HTTP response to complete before sending the second call. This can be done by the developer once the return value of `respond()` is specified (https://github.com/slackapi/node-slack-interactive-messages/issues/37), but its still not intuitive.

We can implement a queue with a max concurrency of 1 in the adapter such that calls to `respond()` are serially sent to Slack. This would result in more intuitive behavior. One recommendation for implementing this queue is the [`p-queue`](https://www.npmjs.com/package/p-queue) package (it's also used in `@slack/client`).

It may also be useful to add an optional parameter that skips the queue and sends the request as soon as possible.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

misscoded:
We are closing this issue as `@slack/interactive-messages` officially reached its EOL on May 31st, 2021. As of that date, development has fully stopped for this package and all remaining open issues and pull requests are being closed. 

At this time, we recommend migrating to [Bolt for JavaScript](https://github.com/slackapi/bolt-js), a framework that offers all of the functionality available in this package (and more!). To help with that process, we've also provided some [migration samples](https://slack.dev/node-slack-sdk/tutorials/migrating-to-v6) for those looking to convert their existing apps.

PR #1333, opened 2021-09-13
Explicitly catch an error querystring with value access_denied

### LINKED ISSUE #1186: Improve error messaging of OAuth flow when user cancels
### Description

Brought to our attention in [#824](https://github.com/slackapi/bolt-js/issues/824) of the `bolt-js` repo, the error message surfaced when a user selects **Cancel** during the OAuth flow does not adequately indicate what has occurred and instead cites `redirect url is missing state or code query parameters` with an error code of `slack_oauth_missing_state`.

The [logic](https://github.com/slackapi/node-slack-sdk/blob/%40slack/oauth%402.0.1/packages/oauth/src/index.ts#L217) in the `@slack/oauth` package should be adjusted to accommodate, as the `MissingStateError` is not reflective of this use case. 

- [ ] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

filmaj:
###  Summary

An `error` querystring on the oauth redirect URL with a value of `access_denied` signifies that the user canceled out of the installation flow.

This PR fixes #1186 / #824

### DIFF
--- a/packages/oauth/src/index.spec.js
+++ b/packages/oauth/src/index.spec.js
@@ -417,6 +417,26 @@ describe('OAuth', async () => {
       assert.isTrue(sent);
     });
 
+    it('should call the failure callback due to missing code query parameter on the URL', async () => {
+      const req = { url: 'http://example.com' };
+      let sent = false;
+      const res = { send: () => { sent = true; } };
+      const callbackOptions = {
+        success: async (installation, installOptions, req, res) => {
+          res.send('successful!');
+          assert.fail('should have failed');
+        },
+        failure: async (error, installOptions, req, res) => {
+          assert.equal(error.code, ErrorCode.MissingStateError)
+          res.send('failure');
+        },
+      }
+      const installer = new InstallProvider({ clientId, clientSecret, stateSecret, installationStore, logger: noopLogger });
+      await installer.handleCallback(req, res, callbackOptions);
+
+      assert.isTrue(sent);
+    });
+
     it('should call the failure callback if an access_denied error query parameter was returned on the URL', async () => {
       const req = { url: 'http://example.com?error=access_denied' };
       let sent = false;
--- a/packages/oauth/src/index.spec.js
+++ b/packages/oauth/src/index.spec.js
@@ -417,8 +417,8 @@ describe('OAuth', async () => {
       assert.isTrue(sent);
     });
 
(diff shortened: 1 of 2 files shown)


packages/oauth/src/index.spec.js
```diff
@@ -417,8 +417,8 @@ describe('OAuth', async () => {
       assert.isTrue(sent);
     });
 
-    it('should call the failure callback due to missing code query parameter on the URL', async () => {
```
filmaj, line 420: I've essentially removed this test as it is line-for-line exactly the same as the previous test. While the intention of the two tests are different, because the callback handler needs _one of_ a `code` _or_ a `state` parameter returned, the result ends up a little off.

Alternatively, we could rewrite these tests a little to check that the callback handler passes if _either_ `state` _or_ `code` is provided.

packages/oauth/src/index.ts
```diff
@@ -315,12 +315,17 @@ export class InstallProvider {
...
+    let flowError: string;
     let state: string;
     let installOptions: InstallURLOptions;
 
     try {
       if (req.url !== undefined) {
         parsedUrl = new URL(req.url);
+        flowError = parsedUrl.searchParams.get('error') as string;
+        if (flowError === 'access_denied') {
+          throw new AuthorizationError('User cancelled the OAuth installation flow!');
```
filmaj, line 327: I decided to reuse an existing error, `AuthorizationError`, to describe this situation. Let me know if something else is desired.

packages/oauth/src/index.spec.js
srajiang: I've got a [branch open](https://github.com/slackapi/node-slack-sdk/pull/1329) with some updates to have these tests properly test `state` and `code` and throw different errors in each case. Related to making `state` optional depending on whether `stateValidation` flag is true/false.

filmaj: Ooooo nice!

srajiang: Perhaps just add your test separately for now, since it's testing something unrelated to either `state` or `code`?

filmaj: Sounds good, just pushed it back up!

stevengill (approved):

seratch (approved):

PR #1337, opened 2021-09-15
Move token sending to authorization header instead of body parameter

### LINKED ISSUE #1132: web-api token should be passed in header instead of body param
### Description

Slack's api is dropping support in Feb 2021 for `token` to be passed in as a body param. We should update web-api to always pass in token via header

---

#### Packages:

Select all that apply:

- [x] `@slack/web-api`

filmaj:
Fixes #1132

I also removed some trailing whitespace because that's the kind of nitpicky dude I am

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -25,6 +25,7 @@ describe('WebClient', function () {
       assert.instanceOf(client, WebClient);
       assert.equal(client.token, token);
       assert.equal(client.slackApiUrl, 'https://slack.com/api/')
+      assert.equal(client.axios.defaults.headers.Authorization, `Bearer ${token}`);
     });
 
     it('should build a client without a token', function () {
@@ -127,14 +128,14 @@ describe('WebClient', function () {
         setLevel: sinon.spy(),
         setName: sinon.spy(),
       };
- 
-      const client = new WebClient(undefined, { 
-        timeout: timeoutOverride, 
+
+      const client = new WebClient(undefined, {
+        timeout: timeoutOverride,
         retryConfig: { retries: 0 },
-        logLevel: LogLevel.WARN, 
-        logger 
-      });  
-      
+        logLevel: LogLevel.WARN,
+        logger
+      });
+
       client.apiCall('users.list')
         .then(_ => {
           done(new Error("expected timeout to throw error"));
@@ -151,7 +152,7 @@ describe('WebClient', function () {
           } catch (err) {
             done(err);
           }
-        }); 
+        });
     });
   });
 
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.spec.js
```diff
@@ -521,9 +522,6 @@ describe('WebClient', function () {
             const file = parts.files[0];
             // the filename is picked up from the the ReadableStream since it originates from fs
             assert.include(file, { fieldname: 'someBinaryField', filename: 'train.jpg' });
-
-            assert.lengthOf(parts.fields, 1);
-            assert.deepInclude(parts.fields, { fieldname: 'token', value: token });
```
filmaj, line 526: With the token no longer a part of the request body, this part needed removal.

packages/web-api/src/WebClient.ts
```diff
@@ -183,6 +183,9 @@ export class WebClient extends Methods {
       this.logger = getLogger(WebClient.loggerName, logLevel ?? LogLevel.INFO, logger);
     }
 
+    // eslint-disable-next-line no-param-reassign
+    if (!headers.Authorization) headers.Authorization = `Bearer ${this.token}`;
```
seratch, line 187: `this.token` can be absent. For instance, some API calls (such as oauth.v2.access with code parameter, openid.connect.token with code parameter, api.test) do not require token in the authorization header. Can you update the logic to check if `this.token` exists too?

packages/web-api/src/WebClient.spec.js
```diff
@@ -425,7 +426,7 @@ describe('WebClient', function () {
     it('should properly serialize simple API arguments', function () {
       const scope = nock('https://slack.com')
         // NOTE: this could create false negatives if the serialization order changes (it shouldn't matter)
-        .post(/api/, 'token=xoxb-faketoken&team_id=T12345678&foo=stringval&bar=42&baz=false')
+        .post(/api/, 'team_id=T12345678&foo=stringval&bar=42&baz=false')
         .reply(200, { ok: true });
       return this.client.apiCall('method', { foo: 'stringval', bar: 42, baz: false, team_id: 'T12345678' })
```
seratch, line 431: Can we override the token in the authorization header if a developer passes a token like `this.client.apiCall('method', { token: 'xoxb-' })`? Switching tokens is not a rare use case. For instance, the singleton `WebClient` does not hold any token and your app utilizes multiple user tokens for user related API calls.

seratch (changes requested):
I like the change! Can you check the two points I mentioned in comments?

filmaj:
@seratch I believe I have addressed your comments, and also added another test for the Authorization header overriding behaviour via `apiCall()` options parameter.

seratch (approved):
@filmaj LGTM! Having unit tests this way makes reviews easier 👍

stevengill (approved):
Looks great!

PR #1338, opened 2021-09-17
Add stateVerification option to documentation

srajiang:
###  Summary

Adds documentation for stateVerification option. Re: [node-slack-sdk/pull/1329](https://github.com/slackapi/node-slack-sdk/pull/1329).

### DIFF
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -292,7 +292,7 @@ const installer = new InstallProvider({
 
 By default, this package handles generating and verifying a `state` parameter during OAuth installation. This added measure helps to mitigate the risk of [Cross-Site Request Forgery](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12) and is strongly recommended. 
 
-In specific installation scenarios, such as when an Org-wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation.
+In specific installation scenarios with Enterprise Grid organizations, such as when an Org-wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. Now, the installer will no longer require that `state` be present to proceed with installation.
 
 ```javascript
 const installer = new InstallProvider({
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -299,7 +299,7 @@ const installer = new InstallProvider({
   clientId: process.env.SLACK_CLIENT_ID,
   clientSecret: process.env.SLACK_CLIENT_SECRET,
   stateVerification: false,
-  },
+  }
 });
 ```
 ---
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -292,13 +292,12 @@ const installer = new InstallProvider({
 
 By default, this package handles generating and verifying a `state` parameter during OAuth installation. This added measure helps to mitigate the risk of [Cross-Site Request Forgery](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12) and is strongly recommended. 
 
-In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. If `state` is provided in the request url parameters, it will still be verified.
+In specific installation scenarios, such as when an Org-wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation.
 
 ```javascript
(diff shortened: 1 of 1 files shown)


docs/_packages/oauth.md
```diff
@@ -288,6 +288,23 @@ const installer = new InstallProvider({
 ```
 ---
 
+### State verification
+
+By default, this package handles generating and verifying a `state` parameter during OAuth installation. This added measure helps to mitigate the risk of [Cross-Site Request Forgery](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12) and is strongly recommended. 
+
+In specific installation scenarios, such as when an Org-Wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. In this case, the installer will no longer require that `state` be present to proceed with installation. If `state` is provided in the request url parameters, it will still be verified.
```
seratch, line 295: A few comments:

* How about being more specific here> In specific installation scenarios -> In specific installation scenarios with Enterprise Grid organizations?
* `In this case,` is repeated in two consecutive sentences. Can we change the transition words to something else?
* I think that we don't need to do the verification in this case because it's disabled> If `state` is provided in the request url parameters, it will still be verified.

docs/_packages/oauth.md
```diff
@@ -288,6 +288,23 @@ const installer = new InstallProvider({
...
+
+```javascript
+const installer = new InstallProvider({
+  clientId: process.env.SLACK_CLIENT_ID,
+  clientSecret: process.env.SLACK_CLIENT_SECRET,
+  stateSecret: 'my-state-secret',
```
seratch, line 301: Should developers still set stateSecret here? 

In the above comment, you mentioned that even with stateVerification: false, state parameter can be generated. But I don't think that it's easy-to-understand behavior. If the flag is false, we can skip state parameter generation.

docs/_packages/oauth.md
srajiang: @seratch There's also another scenario, the Add to Slack button on the App config page also doesn't supply state. 
 
I'm thinking it's possible for an app to need to be installed from the`slack/install` page and from admin page / Add to Slack app configuration button. In that case a `state` param could sometimes be supplied and sometimes not. If an app is providing a state, wouldn't it be more secure to verify it?

seratch: @srajiang For the scenario, our recommendation would be using Direct Install URL. As you know, we've recently added the support for it in bolt-js (it's not yet released though): https://github.com/slackapi/bolt-js/issues/977 With this way, you can use `/slack/install` for Add to Slack in App Directory.

docs/_packages/oauth.md
srajiang: Great point! With the latest merged changes to the [oauth package](https://github.com/slackapi/node-slack-sdk/pull/1339), state store initialization is skipped when state verification is disabled (so implicitly state parameter generation and verification are both skipped as well). This also means that we can support removing the hard requirement for developers to set `stateSecret` in these specific cases. Hopefully that presents a more clear developer experience. 


```javascript
const installer = new InstallProvider({
  clientId: process.env.SLACK_CLIENT_ID,
  clientSecret: process.env.SLACK_CLIENT_SECRET,
  stateVerification: false,
  stateSecret: 'mysecret' // no longer required when stateVerification is false
  },
});
```

docs/_packages/oauth.md
```diff
@@ -288,6 +288,22 @@ const installer = new InstallProvider({
...
+By default, this package handles generating and verifying a `state` parameter during OAuth installation. This added measure helps to mitigate the risk of [Cross-Site Request Forgery](https://datatracker.ietf.org/doc/html/rfc6749#section-10.12) and is strongly recommended. 
+
+In specific installation scenarios with Enterprise Grid organizations, such as when an Org-wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. Now, the installer will no longer require that `state` be present to proceed with installation.
+
+```javascript
+const installer = new InstallProvider({
+  clientId: process.env.SLACK_CLIENT_ID,
+  clientSecret: process.env.SLACK_CLIENT_SECRET,
+  stateVerification: false,
+  },
```
seratch, line 302: This line is unnecessary
```suggestion
```

seratch (changes requested):

docs/_packages/oauth.md
srajiang: Thanks, fixed!

docs/_packages/oauth.md
```diff
@@ -288,6 +288,22 @@ const installer = new InstallProvider({
...
+
+In specific installation scenarios with Enterprise Grid organizations, such as when an Org-wide app is installed from an admin page, state verification cannot be completed because a `state` parameter isn't provided. In this case, you can disable `state` verification via setting the `InstallProvider#stateVerification` option to `false`. Now, the installer will no longer require that `state` be present to proceed with installation.
+
+```javascript
+const installer = new InstallProvider({
+  clientId: process.env.SLACK_CLIENT_ID,
+  clientSecret: process.env.SLACK_CLIENT_SECRET,
+  stateVerification: false,
+  }
+});
```
seratch, line 303: It looks still invalid 😄 
```suggestion
});
```

srajiang: Oh boy! Yes, fixed.

seratch (approved):
@srajiang Looks good to me! Once we release a new version including the changes, you can merge this pull request.

PR #1366, opened 2021-11-02
Tentatively disable some of the web-api tests (#1334)

seratch:
###  Summary

This pull request is a kind of workaround for #1334 -- with this change, we can make a new release of web-api package shortly.

### Requirements (place an `x` in each `[ ]`)

* [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #1369, opened 2021-11-03
Fixes #1334: rewrote admin.usergroups and admin.inviteRequests tests

### LINKED ISSUE #1334: Upgrade axios once nock issue is resolved
### Description

Once https://github.com/axios/axios/issues/4036 is resolved, we should bump up the axios dependency.

- [x] testing related

filmaj:
This rewrites some of the troublesome tests we had so that they work with latest nock + axios. 

Also upgraded axios to latest version. Removed `package-lock.json` and added it back to `.gitignore` since we no longer need to lock to specific versions of axios.

This PR fixes #1334 and should be merged in favour of #1366.


packages/web-api/package.json
```diff
@@ -49,7 +49,7 @@
     "@slack/types": "^2.0.0",
     "@types/is-stream": "^1.1.0",
     "@types/node": ">=12.0.0",
-    "axios": "0.21.1",
+    "axios": "^0.24.0",
     "eventemitter3": "^3.1.0",
     "form-data": "^2.5.0",
     "is-stream": "^1.1.0",
```
filmaj, line 55: If we will merge this PR in favour of #1366 , we may still want to include [some of the devDependency updates #1366 introduces](https://github.com/slackapi/node-slack-sdk/pull/1366/files#diff-c37f41fce6c9e7fe08ba94502179041088c29e04666e107c98265272f8391039R76) in this PR too.

packages/web-api/src/WebClient.spec.js
```diff
@@ -1148,132 +1148,79 @@ describe('WebClient', function () {
...
+    });
 
-    it('can call admin.inviteRequests.approve', function (done) {
-      verify(
-        client.admin.inviteRequests.approve({ team_id: 'T123', invite_request_id: 'I123' }),
-        'admin.inviteRequests.approve',
-        'team_id=T123&invite_request_id=I123',
-        done,
-      );
+    it('properly serializes arguments to admin.inviteRequests.approve', async function () {
```
filmaj, line 1162: I took the liberty of renaming the tests here as well to try to make them more representative of what we are actually testing here. Given we are controlling the response of the API to (via `nock`) to simply return the request body, in effect these tests only check that:

1. The relevant methods on the client exist, can be called and return a response, and 
2. That depending on the argument type, the arguments are serialized properly.

As a result I decided to rename these tests to be "check that arguments are serialized properly" since that is in effect what we are exercising in these tests.

packages/web-api/src/WebClient.ts
```diff
@@ -42,7 +42,7 @@ export interface WebClientOptions {
   tls?: TLSOptions;
   timeout?: number;
   rejectRateLimitedCalls?: boolean;
-  headers?: Record<string, unknown>;
+  headers?: Record<string, string>;
```
filmaj, line 45: This change is related to upgrading to the latest axios, which I believe has updated (added?) its types. I assumed this would be an innocent change, since HTTP headers are always strings, but what does everyone else think?

packages/web-api/src/WebClient.spec.js
srajiang: IMO this renaming makes it way more clear what is happening in these tests - thanks 👍 

For context (and later possible eyes on this thread), axios/axios#4036 is the related issue with a succinct summary.

packages/web-api/src/WebClient.ts
seratch: This looks fine.

seratch (approved):
Thanks for fixing this issue!

PR #1384, opened 2021-11-24
Tweaking user-agent instrumentation for deno compatibility

filmaj:
###  Summary

- Injected `process.title` and a (slightly more correct) `process.version` when building via `esbuild` for deno
- Use `process.title` when instrumenting the user-agent; this should yield a slightly more correct user-agent header string when using in deno.

Before this PR, user agent would be the following when executing in different runtimes:

- in node v16.4.2 on a mac running 11.6.1: `'User-Agent': '@slack:web-api/6.5.1 node/16.4.2 darwin/20.6.0',`
- in deno: `"User-Agent": "@slack:web-api/6.5.1 node/16.0.0 browser/undefined",`

After this PR:

- in node v16.4.2 on a mac running 11.6.1: `'User-Agent': '@slack:web-api/6.5.1 node/16.4.2 darwin/20.6.0',`
- in deno: `"User-Agent": "@slack:web-api/6.5.1 deno/1.15.2 browser/undefined",`

These are all temporary compatibility / shim polyfills. Once the compatibility layer in deno for node matures, we should be able to slowly drop all of these small tweaks.

### DIFF
--- a/packages/web-api/src/instrument.ts
+++ b/packages/web-api/src/instrument.ts
@@ -1,4 +1,5 @@
 import * as os from 'os';
+import { basename } from 'path';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-commonjs
 const packageJson = require('../package.json');
@@ -10,8 +11,14 @@ function replaceSlashes(s: string): string {
   return s.replace('/', ':');
 }
 
+// TODO: for the deno build (see the `npm run build:deno` npm run script), we could replace the `os-browserify` npm
+// module shim with our own shim leveraging the deno beta compatibility layer for node's `os` module (for more info
+// see https://deno.land/std@0.116.0/node/os.ts). At the time of writing this TODO (2021/11/25), this required deno
+// v1.16.2 and use of the --unstable flag. Once support for this exists without the --unstable flag, we can improve
+// the `os` module deno shim to correctly report operating system from a deno runtime. Until then, the below `os`-
+// based code will report "browser/undefined" from a deno runtime.
 const baseUserAgent = `${replaceSlashes(packageJson.name)}/${packageJson.version} ` +
-                      `node/${process.version.replace('v', '')} ` +
+                      `${basename(process.title)}/${process.version.replace('v', '')} ` +
                       `${os.platform()}/${os.release()}`;
 
 const appMetadata: { [key: string]: string } = {};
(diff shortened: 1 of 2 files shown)


packages/web-api/src/instrument.ts
```diff
@@ -11,7 +12,7 @@ function replaceSlashes(s: string): string {
 }
 
 const baseUserAgent = `${replaceSlashes(packageJson.name)}/${packageJson.version} ` +
-                      `node/${process.version.replace('v', '')} ` +
+                      `${basename(process.title)}/${process.version.replace('v', '')} ` +
                       `${os.platform()}/${os.release()}`;
```
seratch, line 16: The improvement on the title part looks great to me! Can we do anything for the OS info part, which is currently `browser/undefined` with Deno?

seratch (approved):

packages/web-api/src/instrument.ts
filmaj: I _believe_ `browser/undefined` comes from the [`os-browserify` NPM package](https://www.npmjs.com/package/os-browserify) - and of course, we can shim in whatever values we want! I can look into writing our own miniature `os` compatibility shim if the beta deno one is insufficient or not ready or the Deno OS APIs are not implemented.

filmaj: Just checked - we could use the [built-in compatibility layer for the `os` node module from deno](https://deno.land/std@0.116.0/node/os.ts) but we require to use the latest deno (v1.16.2) and we must use with the `--unstable` flag. If we are ok with increasing the deno requirements in this way, then this can be tweaked to use the same approach as the [`deno-shims/qs-shim.js` file](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/deno-shims/qs-shim.js), something as easy as:

```
export * from 'https://deno.land/std@0.116.0/node/os.ts'
```

seratch: Thanks for checking this! In the near future, our main users will use the latest Deno version for sure. If we have a reason to keep using a specific stable version, I think we don't need to rush for having the os module. In the case, we can have a TODO comment describing the necessity to update the part once the os module is available without unstable flag.

filmaj: @seratch I've added a TODO/comment explaining the situation, thanks for the suggestion! I will now resolve this conversation.

PR #1386, opened 2021-11-29
Add focus_on_load property to views.* API responses

seratch:
###  Summary

This pull request adds focus_on_load in the `views.*` API responses. Also, `url` was added to `team` data recently.

### DIFF
--- a/packages/web-api/src/response/AdminTeamsSettingsInfoResponse.ts
+++ b/packages/web-api/src/response/AdminTeamsSettingsInfoResponse.ts
@@ -28,6 +28,7 @@ export interface Team {
   default_channels?:  string[];
   is_verified?:       boolean;
   enterprise_domain?: string;
+  url?:               string;
 }
 
 export interface Icon {
--- a/packages/web-api/src/response/TeamInfoResponse.ts
+++ b/packages/web-api/src/response/TeamInfoResponse.ts
@@ -18,13 +18,17 @@ export type TeamInfoResponse = WebAPICallResult & {
 };
 
 export interface Team {
-  id?:           string;
-  name?:         string;
-  domain?:       string;
-  email_domain?: string;
-  icon?:         Icon;
-  is_verified?:  boolean;
-  url?:          string;
+  id?:                string;
+  name?:              string;
+  domain?:            string;
+  email_domain?:      string;
+  icon?:              Icon;
+  is_verified?:       boolean;
+  url?:               string;
+  enterprise_id?:     string;
+  enterprise_name?:   string;
+  enterprise_domain?: string;
+  discoverable?:      string;
 }
 
 export interface Icon {
--- a/packages/web-api/src/response/TeamInfoResponse.ts
+++ b/packages/web-api/src/response/TeamInfoResponse.ts
@@ -24,6 +24,7 @@ export interface Team {
   email_domain?: string;
   icon?:         Icon;
   is_verified?:  boolean;
+  url?:          string;
 }
 
 export interface Icon {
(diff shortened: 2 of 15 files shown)


packages/web-api/src/response/AdminTeamsSettingsInfoResponse.ts
```diff
@@ -28,6 +28,7 @@ export interface Team {
   default_channels?:  string[];
   is_verified?:       boolean;
   enterprise_domain?: string;
+  url?:               string;
```
filmaj, line 31: I don't see any of `enterprise_domain`, `is_verified` nor `url` described in the [`admin.teams.settings.info` API method](https://api.slack.com/methods/admin.teams.settings.info) example response. Should it be documented there too?

packages/web-api/src/response/TeamInfoResponse.ts
```diff
@@ -24,6 +24,7 @@ export interface Team {
   email_domain?: string;
   icon?:         Icon;
   is_verified?:  boolean;
+  url?:          string;
```
filmaj, line 27: I'm just comparing what we have in these responses to what's on the docs site, and for [the `teams.info` method](https://api.slack.com/methods/team.info), `is_verified` and `url` does not exist, but it also says that `enterprise_id` and `enterprise_name` exist. Should we add the `enterprise_*` properties here, too, then, and should the `is_verified` and `url` properties be documented on the public site?

packages/web-api/src/response/ViewsOpenResponse.ts
```diff
@@ -86,6 +86,7 @@ export interface Element {
   min_length?:                      number;
   max_length?:                      number;
   dispatch_action_config?:          DispatchActionConfig;
+  focus_on_load?:                   boolean;
```
filmaj, line 89: Good news, was able to find this on the docs site here: https://api.slack.com/reference/block-kit/block-elements
Bad news, not on the API methods example response for `views.open`: https://api.slack.com/methods/views.open - but I guess it would be too verbose to fully document all the possible Block elements on this page? What do you think?

filmaj (approved):
LGTM, left some questions about if/how what we have on the API Docs site needs to be updated as a result.

packages/web-api/src/response/TeamInfoResponse.ts
seratch: >Should we add the enterprise_* properties here, too, 

Yes, we should. The reason why enterprise_* properties are not available in this class is that the Java SDK integration tests with team.info endpoint do with only a non-Grid workspace. I'll update the tests and this PR later on.

>then, and should the is_verified and url properties be documented on the public site?

Yes, the example should be updated.

packages/web-api/src/response/AdminTeamsSettingsInfoResponse.ts
seratch: Yes, we can improve it. Not only the missing fields but also the invalid type data like `"default_channels": "array"`

packages/web-api/src/response/ViewsOpenResponse.ts
seratch: Yeah, I don't think the `views.open` API method page should cover all the patterns of blocks. If we would like to do so, it should mention not only focus_on_load but also many things such as response_url_enabled, and dispatch_action_config. As the page has a link to the modal document, I don't think we should repeat all the things there.

filmaj:
Thanks for the comments on my questions, @seratch ! If you wish I can try to update the api.slack.com/methods documentation I commented on. Let me know!

PR #1390, opened 2021-12-01
web-api deno build: tweaks and improvements for type hint support

filmaj:
###  Summary

- Reorganizing some of the package.json hax into `npm run` scripts for easier deno building locally.
- Removing the need to install additional dependencies and moving to custom-built deno+node shims for most built-in node modules.
- Using shim file approach for the `process` node module, rather than hard-coded strings and too-simplistic symbol substitution.
- Tweaking GitHub Actions workflow to concatenate all built web-api d.ts files into one for use with the deno build.

### Testing

To test this out, pull this branch in and run (NOTE: Linux/Mac only):

```
npm run prepare:deno # this command will mess with your package.json
npm install
npm run build:deno # use esbuild to build the deno artifact, which will be named mod.js
find dist -type f -name '*.d.ts' -exec cat {} \; > mod.d.ts # builds a mod.d.ts sidekick for mod.js
git checkout -- package.json # clean up your package.json from step 1
```

Then in vscode, create a new JS / Deno / TypeScript project and start with:

```
import { WebClient, /* whatever else from web-api */ } from './path/to/mod.js';
```

Then give the autocomplete bits a shot! It should be identical to the existing node build

### DIFF
--- a/.github/workflows/deno-build.yml
+++ b/.github/workflows/deno-build.yml
@@ -26,21 +26,22 @@ jobs:
         node-version: 16
     - name: Echo Tag Name
       run: echo "$GITHUB_REF_NAME"
-    - name: Inject browser field and add dependencies to package.json
-      run: |
-        perl -pi.bak -e 's/"tsd": {/"browser":{"os":"os-browserify","path":"path-browserify","querystring":".\/deno-shims\/qs-shim.js"},"tsd": {/g' package.json
-        perl -pi.bak -e 's/"axios": /"path-browserify":"1.0.1","os-browserify":"0.3.0","axios": /g' package.json
+    - name: Inject browser field to package.json
+      run: npm run prepare:deno
     - run: npm install
     - run: npm run build:deno
-    - run: mv mod.js ../../.
+    - name: Concatenate .d.ts files into one declaration file for use in Deno
+      run: |
+        find dist -type f -name '*.d.ts' -exec cat {} \; > mod.d.ts
+    - run: mv mod.js mod.d.ts ../../.
     - name: Commit and tag built artifact to separate branch
       run: |
         git config user.name slack-cicd
         git config user.email github-cicd@slack.com
         git checkout -- package.json
         git checkout deno-web-api
-        mv ../../mod.js .
-        git add mod.js
-        git commit -m "Adding transpiled deno ${GITHUB_REF_NAME} release artifact."
+        mv ../../mod.js ../../mod.d.ts .
+        git add mod.js mod.d.ts
+        git commit -m "Adding transpiled deno ${GITHUB_REF_NAME} release artifacts."
         git tag "deno/${GITHUB_REF_NAME}"
         git push --tags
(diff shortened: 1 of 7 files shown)


seratch:
@filmaj 
I just tried the steps in the GitHub Actions job but the find command fails because of the absence of `dist` directory. Is it only me? (macOS 11.6.1)
```bash
$ npm run build:deno

> @slack/web-api@6.5.1 build:deno /Users/ksera/github/node-slack-sdk/packages/web-api
> npm run build:clean && esbuild src/index.ts --bundle --define:Buffer=dummy_buffer --inject:./deno-shims/buffer-shim.js --inject:./deno-shims/xhr-shim.js --define:process=dummy_process --inject:./deno-shims/process-shim.js --target=esnext --format=esm --outfile=./mod.js

> @slack/web-api@6.5.1 build:clean /Users/ksera/github/node-slack-sdk/packages/web-api
> shx rm -rf ./dist ./coverage ./.nyc_output mod.js

  mod.js  151.2kb

⚡ Done in 32ms
$ find dist -type f -name '*.d.ts' -exec cat {} \; > mod.d.ts
find: dist: No such file or directory
$
```

seratch:
Thanks for the update! The build script does not fail anymore but I'm still unsure the types are correctly exported.

If you run the following script,

```typescript
import { WebClient } from '/path/to/mod.js';

(async () => {
  try {
    const response = await client.api.test({});
    console.log(response);
  } catch (e) {
    console.log(e);
  }
})();
```

the `response` object should be statically typed in TypeScript source code. However, the value is resolved as an any value as below:



Also, I may misunderstand something but the merging all the response types into one file may not work. The sub types in each response data are not universally unique (= there are so many duplicated names). For example, most of the response types has their own `ResponseMetadata` class. `Message` for `ChatPostMessageResponse` and `Message` for `ChatScheduleMessageResponse` are not shared. If we have to concatenate them into one file, a kind of namespace or so would be necessary.

filmaj:
Ohhhh.... nice catch. Yes that is a challenge then 😅  the other obstacle is that the `esbuild` output does sometimes do variable reassignment, so perhaps some of the names do not align properly. Thank you for pointing this out. I will take another look..

Perhaps the approach with `esbuild` is not a good one.. the only transpilation that would be required, I think, on the original source code would be converting `require()` to `import`. That said, the issue is bundling the dependencies so that they function in Deno...

filmaj:
> Also, I may misunderstand something but the merging all the response types into one file may not work. The sub types in each response data are not universally unique (= there are so many duplicated names). For example, most of the response types has their own ResponseMetadata class. Message for ChatPostMessageResponse and Message for ChatScheduleMessageResponse are not shared. If we have to concatenate them into one file, a kind of namespace or so would be necessary.

Yeah you are right here. Some of the generated `d.ts` files also have default exports, which would not be able to be combined into one file. Argh.

It's also difficult to debug VSCode autocompletion / typehint behaviour. It looks to me like, when using the node package, VSCode uses both the `d.ts` file definitions as well as the typescript source to infer some of the types and behaviours. With `esbuilt`, we lose all the source type information and rely purely on `d.ts` files.

Perhaps the `esbuild` approach is not the correct one to employ. The challenge is getting the dependencies of the project working in node. Maybe I try a different tactic of purely bundling the dependencies with `esbuild` and then using a different approach for the project source, to maintain its type information.

filmaj:
LOL maybe I should have paid attention to this tab in VSCode 😂  (still learning my way around this IDE!)

Screen Shot 2021-12-06 at 5 21 24 PM

PR #1392, opened 2021-12-03
Add external_id to WorkflowStepView

### LINKED ISSUE #1391: WorkflowStepView doesn't work if there are multiple `views.update` API calls
#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [x] `@slack/types`
- [ ] I don't know

### Reproducible in:
```
"slack/bolt": "^3.8.1",
"slack/logger": "^3.0.0",
"slack/socket-mode": "^1.2.0",
"slack/web-api": "^6.5.1",
"slack/oauth": "^2.3.0",
"slack/types": "^2.3.0",
v14.17.6
ProductName:	macOS
ProductVersion:	11.5.2
BuildVersion:	20G95
Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:31 PDT 2021; root:xnu-7195.141.2~5/RELEASE_X86_64
```

#### Steps to reproduce:
1. Make an API call to `views.update` using the new `WorkflowStepView` object. The first view update will be successful.
2. Make another API call to `views.update` to update the same view. This time the call will return an error `not_found`.

The reason is because `external_id` is not a property in `WorkflowStepView`.

### Expected result:
`views.update` API call should be successful and the view should continue to expand with new fields.
Screen Shot 2021-12-02 at 12 30 04 PM

### Actual result:
An error occurred and the view doesn't get updated.
Screen Shot 2021-12-02 at 10 55 06 AM
```
{"level":"ERROR","msg":"An API error occurred: not_found","line_number":23,"function":null,"file":"src/libs/logging/mw-bolt-ctx-logger.ts","error_name":"Error","error_msg":"An API error occurred: not_found"}
```

### Requirements

For general questions/issues about Slack API platform or its server-side, could you submit questions at https://my.slack.com/help/requests/new instead. :bow:

srajiang:
###  Summary

Fixes #1391 by adding `external_id` properties required by [views.update](https://api.slack.com/methods/views.update) method. 

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -58,7 +58,7 @@ export interface WorkflowStepView {
   private_metadata?: string;
   callback_id?: string;
   submit_disabled?: boolean; // defaults to false
-  external_id?: string; 
+  external_id?: string;
   view_id?: string;
 }
 
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -58,6 +58,8 @@ export interface WorkflowStepView {
   private_metadata?: string;
   callback_id?: string;
   submit_disabled?: boolean; // defaults to false
+  external_id?: string; 
+  view_id?: string;
 }
 
 export type View = HomeView | ModalView | WorkflowStepView;
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -59,7 +59,6 @@ export interface WorkflowStepView {
   callback_id?: string;
   submit_disabled?: boolean; // defaults to false
   external_id?: string;
-  view_id?: string;
 }
 
 export type View = HomeView | ModalView | WorkflowStepView;


seratch (approved):

filmaj (changes requested):
Hmm, now that I'm reviewing this, I'm wondering if these changes are correct. @seratch @srajiang can you double-check my assumptions below?

In digging into where these types are used, I can see that it is consumed in the `@slack/web-api` package: the super-type [`View` is imported in `src/methods.ts`](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/methods.ts#L2), a e.g. [`ViewsUpdateArguments` interface uses the `View` as a `view` property](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/methods.ts#L2048-L2053) and finally the e.g. [`views.update` method uses `ViewsUpdateArguments` as the arguments](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/methods.ts#L698-L703) to be passed in.

However, [the optional `external_id` and `view_id` properties are set as properties on this `ViewsUpdateArguments` interface inside `@slack/web-api`](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/methods.ts#L2048-L2053) - not on the `WorkflowStepView` type exported by `@slack/types`.

With that context, perhaps this isn't the right change? It seems like the various `*View` types exported by `@slack/types` are meant to model the `view` object to be passed into the various `views.*` API methods.

Perhaps there is nothing wrong here? Am I totally off-base?

filmaj (approved):
Nevermind! @srajiang and @bkeung astutely pointed out that the View object itself can also have the IDs as part of the responses coming back. My mistake!

srajiang:
I have removed `view_id` from the type for now as Fil's pointed out `view_id` is set at the `ViewsUpdateArguments` and is not listed in the [general views documentation](https://api.slack.com/reference/surfaces/views).

seratch:
Thanks for working on this! Not having `view_id` this time is fine, but we updated the view object in API response, not the API argument list here. As long as the actual payload has `view_id`, there is no reason not to have it. We may improve the incomplete views document page plus add the property too in the future.

PR #1396, opened 2021-12-05
Add more labels to exempt-issue-labels in stale GitHub action

seratch:
###  Summary

This pull request improves the stale GitHub Action job not to mark stale to issues with enhancement etc.

### DIFF
--- a/.github/workflows/triage-issues.yml
+++ b/.github/workflows/triage-issues.yml
@@ -26,7 +26,7 @@ jobs:
           stale-issue-label: auto-triage-stale
           stale-issue-message: 👋 It looks like this issue has been open for 30 days with no activity. We'll mark this as stale for now, and wait 10 days for an update or for further comment before closing this issue out.
           close-issue-message: As this issue has been inactive for more than one month, we will be closing it. Thank you to all the participants! If you would like to raise a related issue, please create a new issue which includes your specific details and references this issue number.
-          exempt-issue-labels: auto-triage-skip
+          exempt-issue-labels: auto-triage-skip,bug,discussion,docs,enhancement,security
           exempt-all-milestones: true
           remove-stale-when-updated: true
           enable-statistics: true


srajiang:
I have a concern that if we poke many holes in the rules for auto-triage, longterm we will lose out on some advantages that it brings to auto cleaning things up, and we will be manually managing most tickets. I am less familiar with the overall state of the issues in each repo, so I assume there's good reasons why we might want to add these exemptions, I just don't fully understand the reasoning yet: 

My thinking: If there's an `enhancement` which we have not added to a milestone (i.e, backlog) during initial triage, that means no one on triage on the team has yet prioritized this issue for work so when this is staled out at the 30-day mark, this will bring it back to the team's (and OPs attention). At this point, we can add this to a milestone again if it needs to be worked on, and closed out (if ignored or no action at that point). 

Same goes here for `bug`. If there is a bug that no one has taken action on (either through adding to a milestone or outright resolving) in 30 days, wouldn't we want the triage bot to surface bug that haven't been marked with this tag, so that we're more likely to action them? 

I don't want to negatively impact the original poster or participants' experience and cause additional thrash in the case where we do want to work on an issue, have it marked as `enhancement` or `bug`, but again these should already be assigned to a milestone (i.e. in the backlog (i.e. a `milestone`) at the time when it's staled.

One thing we may consider is modifying the language fo the stale message to include a note which makes it clear that maintainers are also responsible for issues getting to resolution. 

Something like:  
> 👋 It looks like this issue has been open for 30 days with no activity. We'll mark this as stale for now, and wait 10 days for an update or for further comment before closing this issue out. If you think this issue needs to be prioritized, please comment to get the thread going again! Maintainers also review issues marked as `stale` on a regular basis and comment or adjust status if the issue needs to be reprioritized. 

Are these proposed changes necessary? @stevengill brought this up in a previous discussion, so please chime in if you have thoughts.

seratch:
@srajiang All make sense. Can you work on the updates (updating the message to have the part for maintainers)? We can close my PRs like this

PR #1410, opened 2022-01-07
Fix #1409 installerOptions.failure() options can be undefined

seratch:
###  Summary

This pull request resolves #1409 

### Requirements (place an `x` in each `[ ]`)

* [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #1411, opened 2022-01-08
Refactor the internals of InstallProvider for better null safety

seratch:
###  Summary

This pull request improves the internals in `InstallProvider` for the patterns where `InstallationStore` returns data including null values. The fix is related to the discussion at https://github.com/slackapi/bolt-js/issues/1265 but the changes here may not resolve the situation the reporter is facing.

### DIFF
--- a/packages/oauth/src/index.ts
+++ b/packages/oauth/src/index.ts
@@ -134,6 +134,9 @@ export class InstallProvider {
    */
   public async authorize(source: InstallationQuery<boolean>): Promise<AuthorizeResult> {
     try {
+      // Note that `queryResult` may unexpectedly include null values for some properties.
+      // For example, MongoDB can often save properties as null for some reasons.
+      // Inside this method, we should alwayss check if a value is either undefined or null.
       let queryResult;
       if (source.isEnterpriseInstall) {
         queryResult = await this.installationStore.fetchInstallation(source as InstallationQuery<true>, this.logger);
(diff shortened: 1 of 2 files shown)


packages/oauth/src/index.ts
```diff
@@ -134,26 +134,32 @@ export class InstallProvider {
...
-      if (queryResult === undefined) {
+      if (queryResult === undefined || queryResult === null) {
         throw new Error('Failed fetching data from the Installation Store');
       }
 
       const authResult: AuthorizeResult = {};
-      authResult.userToken = queryResult.user.token;
 
-      if (queryResult.team !== undefined) {
+      if (queryResult.user) {
```
seratch, line 153: Basically the top-level `user` should not be either null or undefined. That being said, avoiding a runtime error due to the absence is safer than the current code.

packages/oauth/src/index.ts
```diff
@@ -134,26 +134,32 @@ export class InstallProvider {
...
+        authResult.userToken = queryResult.user.token;
+      }
+
+      if (queryResult?.team?.id) {
```
seratch, line 157: Changed from `queryResult.team !== undefined`. As it's fine to skip any falsy values such as 0 and an empty string here, we can check if a value is truthy.

packages/oauth/src/index.ts
```diff
@@ -802,14 +808,16 @@ function detectExpiredOrExpiringTokens(authResult: AuthorizeResult, currentUTCSe
   const tokensToRefresh: string[] = [];
   const EXPIRY_WINDOW: number = 7200; // 2 hours
 
-  if (authResult.botRefreshToken !== undefined && authResult.botTokenExpiresAt !== undefined) {
+  if (authResult.botRefreshToken &&
+    (authResult.botTokenExpiresAt !== undefined && authResult.botTokenExpiresAt !== null)) {
```
seratch, line 812: Although 0 value does not mean anything here, just in case, I avoided truthy value check for expiresAt values

misscoded (approved):
Love to see it. I greatly prefer the more concise truthy/falsy checks, but if my memory serves me right, previously we'd been required to include the explicit `=== undefined` for TS linting reasons. 

The changes here indicate that we no longer have that restriction. Is my understanding correct?

packages/oauth/src/index.ts
filmaj: Tiny little thing: do we need the optional chaining `?` operator on the `queryResult` variable here? Since we check for null or undefined on `queryResult` a few lines before?

filmaj (approved):
LGTM! Left one small comment. I had questions around behaviour with a missing or falsy `user` property on the installation record, but the tests do a good job of explaining expected behaviour - thanks!

packages/oauth/src/index.ts
seratch: Good point. will remove the `?`

seratch:
@misscoded Thanks for pointing it out! Indeed, as you mentioned, some of the changes in this PR caused new lint warnings (I overlooked them as there are a few other warnings already). I'll change those to eliminate the warnings.

seratch:
@misscoded Never mind. I just checked the warnings again and actually they are related to the issue I've resolved by https://github.com/slackapi/node-slack-sdk/pull/1410 (the linter had been detecting the error as warnings!)

```
Warning:   527:32  warning  Forbidden non-null assertion              @typescript-eslint/no-non-null-assertion
Warning:   530:32  warning  Forbidden non-null assertion              @typescript-eslint/no-non-null-assertion
```

We don't have any warnings due to the changes here.

Issue #1421, opened 2022-02-11
What can I do about `Retry header did not contain a valid timeout`?


meichstedt:
Hey there,

I've been running into 3 occurrences of this error: `Retry header did not contain a valid timeout` (from [here](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/WebClient.ts#L417))

The client was using the outdated 5.15.0 WebApi, so it didn't tell me the url.

My main questions are
- is there a way to prevent getting rate limit responses w/o retrySec information?
- are there any ideas about what causes this?
- Is there a best practice on how to handle these?

Thanks in advance

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

### Reproducible in:

```bash
npm ls | grep -o "\S\+@\S\+$" | tr @ ' ' | awk -v q='"' '{print q$1q": "q"^"$2q","}' | grep slack
node --version
sw_vers && uname -v # or `ver`
```

#### The Slack SDK version

@slack/web-api: 5.15.0

#### Python runtime version

v15.14.0

#### OS info

Chrome/98.0.4758.80
Macintosh; Intel Mac OS X 10_15_7

#### Steps to reproduce:

(Share the commands to run, source code, and project settings)

1. Send a slack message
2. Receive a 429
3. Miss the timeout

### Expected result:

I don't know, assume a default timeoutSec?

### Actual result:

(Tell what actually happened with logs, screenshots)

### Requirements

For general questions/issues about Slack API platform or its server-side, could you submit questions at https://my.slack.com/help/requests/new instead. :bow:

Please read the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct) before creating this issue or pull request. By submitting, you are agreeing to those rules.

filmaj:
Hey @meichstedt, sorry about the trouble you're experiencing. If at all possible, if you have some logs to share around when this happened and which URL/API you saw was returning no retry delay information in its header, that would be helpful to dig into this further. I can then dig into the relevant backend API code to see how this could come about.

At this point, I don't have any good answer for your first two questions, but with more information like logs and which APIs, we could take a deeper look.

[...]

PR #1426, opened 2022-02-15
Include Retry-After raw response header value when throwing an error due to invalid Retry-After header

filmaj:
###  Summary

This is an incremental improvement for #1421. When we raise an exception due to a seemingly invalid or unparseable `Retry-After` header value, we include the raw header value in the error thrown to help track down / narrow in a potential cause for #1421.


srajiang (approved):
LGTM 👍

mwbrooks (approved):
Nice!

seratch (approved):
nice improvement

PR #1425, opened 2022-02-15
Fix eslint warnings and TODOs in OAuth module

seratch:
###  Summary

As a preparation before further development on the OAuth module, this pull request resolves the following issues and TODOs:

* eslint warnings (the following ones plus disabled ones)
* TODOs such as user token retrieval in FileInstallationStore

```
  200:38  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  241:21  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  521:21  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  797:24  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
```

### Requirements (place an `x` in each `[ ]`)

* [x] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #1430, opened 2022-02-21
Organize the source files in the OAuth package

seratch:
###  Summary

As a preparation for my forthcoming pull requests, this patch splits the big `index.ts` file into smaller pieces of code. This change improves the maintainability of both the main code and test code.

The changes in this PR are:

* Moved many parts from `index.ts`
  * Moved `ClearStateState` from the file to `./src/state-stores/clear-state-store.ts`
  * Moved other public interface from the file to `./src/state-stores/interface.ts`
  * Moved `InstallationStore` from the file to `./src/stores/interface.ts`
  * Also, moved many tests from `./src/index.spec.js` to relevant spec files
  * Moved `InstallProvider`, `CallbackOptions`, `InstallURLOptions`, `InstallProviderOptions`, `Installation`, `InstallationQuery` to dedicated files
  * Added tests for all of the above
  * The exports in `src/index.ts` are fully backward-compatible
* Added a new test for `logger.ts`

I will rebase https://github.com/slackapi/node-slack-sdk/pull/1425 after merging this PR.


packages/oauth/package.json
```diff
@@ -33,7 +33,7 @@
     "build:clean": "shx rm -rf ./dist ./coverage ./.nyc_output",
     "lint": "eslint --ext .ts src",
     "test": "npm run lint && npm run test:mocha",
-    "test:mocha": "nyc mocha --config .mocharc.json src/*.spec.js",
+    "test:mocha": "nyc mocha --config .mocharc.json src/*.spec.js src/**/*.spec.js src/*.spec.ts src/**/*.spec.ts",
```
seratch, line 36: As I've added more tests including the ones in TypeScript, this change is necessary for running all

packages/oauth/package.json
```diff
@@ -61,7 +63,7 @@
     "eslint-plugin-node": "^11.1.0",
     "mocha": "^9.1.0",
     "nop": "^1.0.0",
-    "nyc": "^14.1.1",
+    "nyc": "^15.1.0",
```
seratch, line 66: for fixing `npm audit` warnings

packages/oauth/src/callback-options.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { assert } from 'chai';
+import { describe, it } from 'mocha';
+import sinon from 'sinon';
+import { IncomingMessage, ServerResponse } from 'http';
+
+import { CallbackOptions } from './callback-options';
+import { MissingStateError } from './errors';
+
+describe('CallbackOptions', async () => {
+  it('should have success and failure', async () => {
```
seratch, line 10: These tests are not so great but it's better than nothing

packages/oauth/src/install-provider.spec.js
```diff
@@ -631,232 +630,4 @@ describe('OAuth', async () => {
       assert.equal(fakeStateStore.verifyStateParam.callCount, 0);
     });
   });
-
-  describe('MemoryInstallationStore', async () => {
```
seratch, line 635: These deleted tests are moved to more relevant files.

packages/oauth/src/install-provider.ts
```diff
@@ -0,0 +1,654 @@
...
+        defaultCallbackFailure(error, installOptions, req, res);
+      }
+    }
+  }
+}
+
+// This is intentionally structurally identical to AuthorizeResult from App
+// It is redefined so that this class remains loosely coupled to the rest
+// of Bolt.
+export interface AuthorizeResult {
```
seratch, line 531: We may want to have dedicated files for these interface. However, I didn't do so at this moment because these are used only in `InstallProvider`. If we want to make it more granular, we can do so in future changes.

seratch: Also, these are exported in `index.ts` as with before

packages/oauth/src/logger.spec.js
```diff
@@ -0,0 +1,21 @@
...
+const { getLogger } = require('./logger');
+const { LogLevel } = require('@slack/logger');
+
+describe('Logger', async () => {
+  it('should create a new logger instance', async () => {
+    const newLogger = getLogger('test-logger-name', LogLevel.DEBUG);
+    assert.match(newLogger.name, /test-logger-name:\d+/);
+    assert.equal(newLogger.getLevel(), LogLevel.DEBUG);
+  });
+  it('should modify the passed logger', async () => {
```
seratch, line 12: To make this clear, I've added this test

packages/oauth/src/stores/file-store.spec.js
```diff
@@ -0,0 +1,262 @@
...
+    channel: 'someChannel',
+    channelId: 'someChannelID',
+    configurationUrl: 'someConfigURL',
+  },
+  appId: 'fakeAppId',
+  tokenType: 'tokenType',
+  isEnterpriseInstall: false,
+}
+
+// TODO: valid tests with org-wide installations
```
seratch, line 44: We should add more tests for org-wide installations and token rotation use cases

packages/oauth/tsconfig.json
```diff
@@ -30,6 +30,7 @@
   ],
   "exclude": [
     "src/**/*.spec.js",
+    "src/**/*.spec.ts",
```
seratch, line 33: Added a few TS tests. They should be excluded in the package file. I've verified if this works by running `npm pack`.

seratch (commented):
comments for reviewers

packages/oauth/src/stores/file-store.spec.js
```diff
@@ -0,0 +1,262 @@
...
+  afterEach(() => {
+    fsMakeDir.restore();
+    fsWriteFile.restore();
+    fsReadFileSync.restore();
+    fsUnlink.restore();
+    fsReaddirSync.restore();
+  });
+
+  it('should store the latest installation', async () => {
+    const installationStore = new FileInstallationStore({ baseDir: './tmp' });
```
seratch, line 93: Changed the base directory for outputs - previously the tests used to generate directories in current dir.

packages/oauth/src/install-provider.ts
seratch: Revisited this and I came to think that `AuthorizeResult` should be `authorize-result.ts` as external code relies on this interface.

packages/oauth/src/callback-options.spec.ts
filmaj: Indeed, only about ~62% of branches are tested in this package, so definitely room for improvement!

packages/oauth/src/stores/file-store.spec.js
filmaj: Not necessary as a change but another option is to use [`os.tmpDir()`](https://nodejs.org/api/os.html#ostmpdir)

filmaj (approved):
LGTM! Ran it locally. Also much easier now to figure out where the missing code coverage is (via `npm it && nyc report --reporter=text`) - thanks a lot for doing this.

packages/oauth/src/stores/file-store.spec.js
seratch: Oh, nice! will update this part.

seratch:
Thanks for your review @filmaj !

PR #1436, opened 2022-02-25
Fix #1435 Proper use of state parameter for the OAuth CSRF protection

### LINKED ISSUE #1435: Proper use of state parameter for the OAuth CSRF protection
As of Feb 2022, `@slack/oauth` (~ v2.4) and `@slack/bolt` (~ v3.10) do not use `state` parameter for the purpose of CSRF protection in the Slack OAuth flow. Refer to the RFC https://datatracker.ietf.org/doc/html/rfc6749#section-10.12 for the details of state parameter in general.

Currently, the only things that `InstallProvider` does are adding `state` parameter to the Slack authorize URL, and then decoding the underlying value from the string. There is no way for developers to check if the state parameter in query string is valid for the accessing user-agent (=the web browser).

To improve this, `@slack/oauth` module can provide a new way to serve the install path (`/slack/install`) in a way to manage user-agent state by setting cookies. With that, the Redirect URL (`/slack/oauth_redirect`) can check if the user-agent is the same person in the OAuth flow.

We will add a new method `handleInstallPath()`, which handles all the above. Also, the existing `handleCallback()` verifies both the query string and browser cookies. Newer versions of bolt-js (v3.11 or newer) will switch to this new way.

#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [x] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

seratch:
###  Summary

This pull request resolves #1435. Refer to the issue for the context and the reason of the changes.

### DIFF
--- a/packages/oauth/src/install-provider-options.ts
+++ b/packages/oauth/src/install-provider-options.ts
@@ -2,17 +2,99 @@ import { Logger, LogLevel } from '@slack/logger';
 import { WebClientOptions } from '@slack/web-api';
 import { StateStore } from './state-stores';
 import { InstallationStore } from './stores';
+import { InstallURLOptions } from './install-url-options';
 
 export interface InstallProviderOptions {
+
+  /**
+   * Client ID, which can be found under the Basic Information section of your application on https://api.slack.com/apps
+   */
   clientId: string;
+
+  /**
+   * Client Secret, which can be found under the Basic Information section of your application on https://api.slack.com/apps
+   */
   clientSecret: string;
-  stateStore?: StateStore; // default ClearStateStore
-  stateSecret?: string; // required with default ClearStateStore
-  stateVerification?: boolean; // default true, disables state verification when false
+
+  /**
+   * Manages installation data, which can be called by both the OAuth flow and authorize() in event handling
+   */
   installationStore?: InstallationStore; // default MemoryInstallationStore
-  authVersion?: 'v1' | 'v2'; // default 'v2'
-  logger?: Logger;
-  logLevel?: LogLevel;
-  clientOptions?: Omit<WebClientOptions, 'logLevel' | 'logger'>;
+
+  /**
+   * The slack.com authorize URL
+   */
   authorizationUrl?: string;
+
+  /**
+   * Stores state issued to authorization server
+   * and verifies the value returned at redirection during OAuth flow to prevent CSRF
+   */
+  stateStore?: StateStore; // default ClearStateStore
+
+  /**
+   * The secret value used for generating the state parameter value
+   */
+  stateSecret?: string; // required with default ClearStateStore
+
+  /**
+   * handleCallback() verifies the state parameter if true (default: true)
+   */
+  stateVerification?: boolean; // default true, disables state verification when false
+
+  /**
+   * handleCallback() skips checking browser cookies if true (default: false)
+   * Enabling this option is not recommended.
+   * This is supposed to be used only for backward-compatibility with v2.4 and olders.
+   */
+  legacyStateVerification?: boolean;
+
+  /**
+   * The cookie name used for setting state parameter value in cookies
+   */
+  stateCookieName?: string;
+
+  /**
+   * The expiration time in seconds for the state parameter value stored via cookies
+   */
+  stateCookieExpirationSeconds?: number;
+
+  /**
+   * The function for rendering the web page for the install path URL
+   */
+  renderHtmlForInstallPath?: (url: string) => string;
+
+  /**
+   * The install path web page rendering will be skipped if true (default: false)
+   */
+  directInstall?: boolean; // default false, disables rendering "Add to Slack" page for /slack/install when true
+
+  /**
+   * The default is "v2" (a.k.a. Granular Bot Permissions), different from "v1" (a.k.a. "Classic Apps").
+   * More details here:
+   * - https://medium.com/slack-developer-blog/more-precision-less-restrictions-a3550006f9c3
+   * - https://api.slack.com/authentication/migration
+   */
+  authVersion?: 'v1' | 'v2'; // default 'v2'
+
+  /**
+   * The initialization options for the OAuth flow
+   */
+  installUrlOptions?: InstallURLOptions;
+
+  /**
+   * @slack/logger logging used in this class
+   */
+  logger?: Logger;
+
+  /**
+   * @slack/logger logging level used in this class
+   */
+  logLevel?: LogLevel;
+
+  /**
+   * The customization options for WebClient
+   */
+  clientOptions?: Omit<WebClientOptions, 'logLevel' | 'logger'>;
+
 }
--- a/packages/oauth/src/install-provider.spec.js
+++ b/packages/oauth/src/install-provider.spec.js
@@ -245,6 +245,57 @@ describe('InstallProvider', async () => {
     });
   });
 
+  describe('installer.handleInstallPath', async () => {
+    it('should fail if installUrlOptions are not given', async () => {
+      const installer = new InstallProvider({
+        clientId,
+        clientSecret,
+        installUrlOptions: undefined,
+        stateStore: {
+          generateStateParam: sinon.fake.resolves('fakeState'),
+          verifyStateParam: sinon.fake.resolves({})
+        },
+        logger: noopLogger,
+      });
+      const req = {};
+      const headers = {};
+      const res = { setHeader(n, v) { headers[n] = v; }, writeHead: () => {}, end: () => {}, };
+      try {
+        await installer.handleInstallPath(req, res);
+        assert.fail('Exception should be thrown')
+      } catch (e) {
+        assert.equal(e.code, ErrorCode.GenerateInstallUrlError);
+      }
+    });
+    it('should redirect installers to valid authorize URL with state param', async () => {
+      const installer = new InstallProvider({
+        clientId,
+        clientSecret,
+        directInstall: true,
+        installUrlOptions: {
+          scopes: ['channels:read'],
+          metadata: 'some_metadata',
+          teamId: 'T12345',
+          redirectUri: 'https://mysite.com/slack/redirect',
+          userScopes: ['chat:write:user'],
+          metadata: 'foo',
+        },
+        stateStore: {
+          generateStateParam: sinon.fake.resolves('fakeState'),
+          verifyStateParam: sinon.fake.resolves({})
+        },
+        logger: noopLogger,
+      });
+      const req = {};
+      const headers = {};
+      const res = { setHeader(n, v) { headers[n] = v; }, writeHead: () => {}, end: () => {}, };
+      await installer.handleInstallPath(req, res);
+
+      assert.equal(headers['Location'], 'https://slack.com/oauth/v2/authorize?scope=channels%3Aread&state=fakeState&client_id=MY_ID&redirect_uri=https%3A%2F%2Fmysite.com%2Fslack%2Fredirect&team=T12345&user_scope=chat%3Awrite%3Auser');
+      assert.equal(headers['Set-Cookie'], 'slack-app-oauth-state=fakeState; Secure; HttpOnly; Path=/; Max-Age=600');
+    });
+  });
+
   describe('installer.generateInstallUrl', async () => {
     it('should return a generated v2 url', async () => {
       const fakeStateStore = {
@@ -633,6 +684,117 @@ describe('InstallProvider', async () => {
       assert.isTrue(sent);
       assert.equal(fakeStateStore.verifyStateParam.callCount, 0);
     });
+  });
+
+  it('should fail if the state value is not in cookies', async () => {
+    const fakeStateStore = {
+      generateStateParam: sinon.fake.resolves('fakeState'),
+      verifyStateParam: sinon.fake.resolves({})
+    };
+    let sent = false;
+    const res = {
+      send: () => { sent = true; },
+      setHeader: () => {},
+    };
+    const installer = new InstallProvider({ clientId, clientSecret, stateSecret, installationStore, stateStore: fakeStateStore, authVersion: 'v2', logger: noopLogger, clientOptions: fooClientOptions });
+    const fakeState = 'fakeState';
+    const fakeCode = 'fakeCode';
+    const req = {
+      headers: {
+        host: 'example.com',
+        // cookie: `slack-app-oauth-state=${fakeState}`,
+      },
+      url: `http://example.com?state=${fakeState}&code=${fakeCode}`,
+    };
+    await installer.handleCallback(req, res, failureExpectedCallbackOptions);
+    assert.isTrue(sent);
+  });
+  it('should fail if there is a different state value in cookies', async () => {
+    const fakeStateStore = {
+      generateStateParam: sinon.fake.resolves('fakeState'),
+      verifyStateParam: sinon.fake.resolves({})
+    };
+    let sent = false;
+    const res = {
+      send: () => { sent = true; },
+      setHeader: () => {},
+    };
+    const installer = new InstallProvider({ clientId, clientSecret, stateSecret, installationStore, stateStore: fakeStateStore, authVersion: 'v2', logger: noopLogger, clientOptions: fooClientOptions });
+    const fakeState = 'fakeState';
+    const fakeCode = 'fakeCode';
+    const req = {
+      headers: {
+        host: 'example.com',
+        cookie: `slack-app-oauth-state=something-different`,
+      },
+      url: `http://example.com?state=${fakeState}&code=${fakeCode}`,
+    };
+    await installer.handleCallback(req, res, failureExpectedCallbackOptions);
+    assert.isTrue(sent);
+  });
+  it('should not fail if no state cookie returned and legacyStateVerification is enabled', async () => {
+    const fakeStateStore = {
+      generateStateParam: sinon.fake.resolves('fakeState'),
+      verifyStateParam: sinon.fake.resolves({})
+    };
+    let sent = false;
+    const res = {
+      send: () => { sent = true; },
+      setHeader: () => {},
+    };
+    const installer = new InstallProvider({
+      clientId,
+      clientSecret,
+      stateSecret,
+      installationStore,
+      logger: noopLogger,
+      stateStore: fakeStateStore,
+      clientOptions: fooClientOptions,
+      legacyStateVerification: true, // this is the key configuration in this test
+    });
+    const fakeState = 'fakeState';
+    const fakeCode = 'fakeCode';
+    const req = {
+      headers: {
+        host: 'example.com',
+        // cookie: `slack-app-oauth-state=${fakeState}`,
+      },
+      url: `http://example.com?state=${fakeState}&code=${fakeCode}`,
+    };
+    await installer.handleCallback(req, res, successExpectedCallbackOptions);
+    assert.isTrue(sent);
+  });
+  it('should not fail if a different state cookie returned and legacyStateVerification is enabled', async () => {
+    const fakeStateStore = {
+      generateStateParam: sinon.fake.resolves('fakeState'),
+      verifyStateParam: sinon.fake.resolves({})
+    };
+    let sent = false;
+    const res = {
+      send: () => { sent = true; },
+      setHeader: () => {},
+    };
+    const installer = new InstallProvider({
+      clientId,
+      clientSecret,
+      stateSecret,
+      installationStore,
+      logger: noopLogger,
+      stateStore: fakeStateStore,
+      clientOptions: fooClientOptions,
+      legacyStateVerification: true, // this is the key configuration in this test
+    });
+    const fakeState = 'fakeState';
+    const fakeCode = 'fakeCode';
+    const req = {
+      headers: {
+        host: 'example.com',
+        cookie: `slack-app-oauth-state=something-different`,
+      },
+      url: `http://example.com?state=${fakeState}&code=${fakeCode}`,
+    };
+    await installer.handleCallback(req, res, successExpectedCallbackOptions);
+    assert.isTrue(sent);
   });
 });
 
--- a/packages/oauth/src/install-provider.ts
+++ b/packages/oauth/src/install-provider.ts
@@ -9,50 +9,79 @@ import {
   InstallerInitializationError,
   UnknownError,
   MissingStateError,
+  InvalidStateError,
   MissingCodeError,
   GenerateInstallUrlError,
   AuthorizationError,
   CodedError,
   ErrorCode,
 } from './errors';
-import { Installation, OrgInstallation } from './installation';
+import { Installation } from './installation';
 import { InstallationQuery } from './installation-query';
 import { InstallURLOptions } from './install-url-options';
 import { InstallProviderOptions } from './install-provider-options';
 import { Logger, LogLevel, getLogger } from './logger';
 import { ClearStateStore, StateStore } from './state-stores';
 import { InstallationStore, MemoryInstallationStore } from './stores';
+import defaultRenderHtmlForInstallPath from './default-render-html-for-install-path';
 
 /**
- * InstallProvider Class.
- * @param clientId - Your apps client ID
- * @param clientSecret - Your apps client Secret
- * @param stateSecret - Used to sign and verify the generated state when using the built-in `stateStore`
- * @param stateStore - Replacement function for the built-in `stateStore`
- * @param stateVerification - Pass in false to disable state parameter verification
- * @param installationStore - Interface to store and retrieve installation data from the database
- * @param authVersion - Can be either `v1` or `v2`. Determines which slack Oauth URL and method to use
- * @param logger - Pass in your own Logger if you don't want to use the built-in one
- * @param logLevel - Pass in the log level you want (ERROR, WARN, INFO, DEBUG). Default is INFO
+ * InstallProvider Class. Refer to InsallProviderOptions interface for the details of constructor arguments.
  */
 export class InstallProvider {
+  // Stores state issued to authorization server and
+  // verifies the value returned at redirection during OAuth flow to prevent CSRF
   public stateStore?: StateStore;
 
+  // Manages installation data, which can be called by both the OAuth flow and authorize() in event handling
   public installationStore: InstallationStore;
 
+  // Client ID, which can be found under the Basic Information section of your application on https://api.slack.com/apps
   private clientId: string;
 
+  // Client Secret, which can be found under the Basic Information section of your application on https://api.slack.com/apps
   private clientSecret: string;
 
+  // The default is "v2" (a.k.a. Granular Bot Permissions), different from "v1" (a.k.a. "Classic Apps").
+  // More details here:
+  // - https://medium.com/slack-developer-blog/more-precision-less-restrictions-a3550006f9c3
+  // - https://api.slack.com/authentication/migration
   private authVersion: string;
 
+  // @slack/logger logging used in this class
   private logger: Logger;
 
-  private clientOptions: WebClientOptions;
+  // The initialization options for the OAuth flow
+  private installUrlOptions?: InstallURLOptions;
 
+  // The install path web page rendering will be skipped if true (default: false)
+  private directInstall: boolean;
+
+  // The function for rendering the web page for the install path URL
+  private renderHtmlForInstallPath: (url: string) => string;
+
+  // The slack.com authorize URL
   private authorizationUrl: string;
 
+  // handleCallback() verifies the state parameter if true (default: true)
   private stateVerification: boolean;
+
+  // handleCallback() skips checking browser cookies if true (default: false)
+  // Enabling this option is not really recommended.
+  // This is supposed to be used only for backward-compatibility with v2.4 and olders.
+  private legacyStateVerification: boolean;
+
+  // The cookie name used for setting state parameter value in cookies
+  private stateCookieName: string;
+
+  // The expinary time for the state parameter value in cookies
+  private stateCookieExpirationSeconds: number;
+
+  // The customization options for WebClient
+  private clientOptions: WebClientOptions;
+
+  // The singleton WebClient instance, which is used in this class
+  private noTokenClient: WebClient;
 
   public constructor({
     clientId,
@@ -60,7 +89,15 @@ export class InstallProvider {
     stateSecret = undefined,
     stateStore = undefined,
     stateVerification = true,
+    // this option is only for the backward-compatibility with v2.4 and older
+    legacyStateVerification = false,
+    stateCookieName = 'slack-app-oauth-state',
+    stateCookieExpirationSeconds = 600, // 10 minutes
+    directInstall = false,
     installationStore = new MemoryInstallationStore(),
+    // If installURLOptions is undefined here, handleInstallPath() does not work for you
+    installUrlOptions = undefined,
+    renderHtmlForInstallPath = defaultRenderHtmlForInstallPath,
     authVersion = 'v2',
     logger = undefined,
     logLevel = undefined,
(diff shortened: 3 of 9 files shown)


packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+   */
   stateVerification?: boolean; // default true, disables state verification when false
-  installationStore?: InstallationStore; // default MemoryInstallationStore
+
+  /**
+   * handleCallback() skips checking browser cookies if true (default: false)
+   * Enabling this option is not really recommended.
+   * This is supposed to be used only for backward-compatibility with v2.4 and olders.
+   */
+  legacyStateVerification?: boolean;
```
seratch, line 49: Added for backward-compatibility (most developers won't use this option but just in case)

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+
+  /**
+   * The cookie name used for setting state parameter value in cookies
+   */
+  stateCookieName?: string;
```
seratch, line 54: for customization - the default ones work fine for most developers, though

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+
+  /**
+   * The expinary time for the state parameter value in cookies
+   */
+  stateCookieExpirationSeconds?: number;
+
+  /**
+   * The function for rendering the web page for the install path URL
+   */
+  renderHtmlForInstallPath?: (url: string) => string;
```
seratch, line 64: Added this for better customization of `handleInstallPath()`

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+
+  /**
+   * The install path web page rendering will be skipped if true (default: false)
+   */
+  directInstall?: boolean; // default false, disables rendering "Add to Slack" page for /slack/install when true
```
seratch, line 69: Added for customization of `handleInstallPath()`

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+
+  /**
+   * The defalt is "v2" ("v1" is for "classic" apps)
+   */
   authVersion?: 'v1' | 'v2'; // default 'v2'
+
+  /**
+   * The initialization options for the OAuth flow
+   */
+  installUrlOptions?: InstallURLOptions;
```
seratch, line 79: This one works as the default but you can pass overridden one to `handleInstallPath()`. `handleCallback()` uses the one decoded from state parameter

packages/oauth/src/install-provider.spec.js
```diff
@@ -238,21 +237,72 @@ describe('InstallProvider', async () => {
...
-        const installer = new InstallProvider({ clientId, clientSecret });
+        new InstallProvider({ clientId, clientSecret, logger: noopLogger, });
       } catch (error) {
         assert.equal(error.code, ErrorCode.InstallerInitializationError);
         assert.equal(error.message, 'To use the built-in state store you must provide a State Secret');
       }
     });
   });
 
+  describe('installer.handleInstallPath', async () => {
```
seratch, line 248: Added tests for handleInstallPath

packages/oauth/src/install-provider.spec.js
```diff
@@ -612,22 +670,118 @@ describe('InstallProvider', async () => {
...
+      headers: {
+        host: 'example.com',
+        cookie: `slack-app-oauth-state=something-different`,
+      },
+      url: `http://example.com?state=${fakeState}&code=${fakeCode}`,
+    };
+    await installer.handleCallback(req, res, failureExpectedCallbackOptions);
+    assert.isTrue(sent);
+  });
+  it('should not fail if legacyStateVerification is enabled', async () => {
```
seratch, line 735: This is the backward compatible mode

packages/oauth/src/install-provider.ts
```diff
@@ -127,12 +172,7 @@ export class InstallProvider {
       // Note that `queryResult` may unexpectedly include null values for some properties.
       // For example, MongoDB can often save properties as null for some reasons.
       // Inside this method, we should alwayss check if a value is either undefined or null.
-      let queryResult;
-      if (source.isEnterpriseInstall) {
```
seratch, line 131: This if/else can be safely removed

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +285,84 @@ export class InstallProvider {
...
    */
   private async refreshExpiringTokens(tokensToRefresh: string[]): Promise<OAuthV2TokenRefreshResponse[]> {
-    const client = new WebClient(undefined, this.clientOptions);
-
-    const refreshPromises = tokensToRefresh.map(async (refreshToken) => await client.oauth.v2.access({
-      client_id: this.clientId,
-      client_secret: this.clientSecret,
-      grant_type: 'refresh_token',
-      refresh_token: refreshToken,
-    }).catch((e) => e) as OAuthV2TokenRefreshResponse);
```
seratch, line 255: I don't think this `as OAuthV2TokenRefreshResponse` is correct.

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +285,84 @@ export class InstallProvider {
...
+    const refreshPromises = tokensToRefresh
```
seratch, line 288: This revised one should be correct but can you double-check? @misscoded

packages/oauth/src/install-provider.ts
seratch: `catch((e) => e)` is not necessary

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +285,84 @@ export class InstallProvider {
...
-   */
-  private static extractSearchParams(req: IncomingMessage): URLSearchParams {
-    const { searchParams } = new URL(req.url as string, `https://${req.headers.host}`);
-    return searchParams;
+   * Handles the install path (the default is /slack/install) requests from an app installer.
+  */
+  public async handleInstallPath(
+    _req: IncomingMessage,
+    res: ServerResponse,
+    installOptions?: InstallURLOptions,
```
seratch, line 309: You can override install options if you pass it here

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +285,84 @@ export class InstallProvider {
...
+    if (installOptions === undefined && this.installUrlOptions === undefined) {
+      const errorMessage = 'To enable the built-in install path handler, you need to pass InstallURLOptions to InstallProvider. ' +
+      "If you're using @slack/bolt, please upgrade the framework to the latest version.";
+      throw new GenerateInstallUrlError(errorMessage);
+    }
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const _installOptions: InstallURLOptions = installOptions! || this.installUrlOptions!;
+    this.logger.debug(`Running handleInstallPath() with ${_installOptions}`);
+
+    if (this.stateStore === undefined) {
```
seratch, line 320: This never happens in normal use cases but the type is `StateStore | undefined`. Thus, I added validation, just in case.

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +285,84 @@ export class InstallProvider {
...
   /**
-   * Returns search params from a URL and ignores protocol / hostname as those
-   * aren't guaranteed to be accurate e.g. in x-forwarded- scenarios
-   */
-  private static extractSearchParams(req: IncomingMessage): URLSearchParams {
-    const { searchParams } = new URL(req.url as string, `https://${req.headers.host}`);
-    return searchParams;
+   * Handles the install path (the default is /slack/install) requests from an app installer.
+  */
+  public async handleInstallPath(
```
seratch, line 306: Reused most of https://github.com/slackapi/bolt-js/blob/%40slack/bolt%403.10.0/src/receivers/HTTPReceiver.ts#L613-L678

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +285,84 @@ export class InstallProvider {
...
 
   /**
    * Returns a URL that is suitable for including in an Add to Slack button
    * Uses stateStore to generate a value for the state query param.
    */
-  public async generateInstallUrl(options: InstallURLOptions, stateVerification: boolean = true): Promise<string> {
+  public async generateInstallUrl(
+    options: InstallURLOptions,
+    stateVerification: boolean = true,
+    state?: string,
```
seratch, line 364: Added this as an optional argument - this should be backward-compatible

packages/oauth/src/install-provider.ts
```diff
@@ -336,48 +436,65 @@ export class InstallProvider {
...
+          res.setHeader('Set-Cookie', this.buildSetCookieHeaderForStateDeletion());
+        }
       }
       if (!installOptions) {
         const emptyInstallOptions: InstallURLOptions = { scopes: [] };
         // eslint-disable-next-line no-param-reassign
         installOptions = emptyInstallOptions;
       }
 
-      const client = new WebClient(undefined, this.clientOptions);
```
seratch, line 372: Removed this instance creation and changed the following code to use a singleton object.

packages/oauth/src/install-provider.ts
```diff
@@ -500,18 +617,14 @@ export class InstallProvider {
       // End: Build the installation object
 
       // Save installation object to installation store
-      if (installation.isEnterpriseInstall) {
```
seratch, line 503: This if/else can be safely removed (again)

packages/oauth/src/state-stores/clear-state-store.spec.js
```diff
@@ -11,4 +11,17 @@ describe('ClearStateStore', async () => {
     const returnedInstallUrlOptions = await stateStore.verifyStateParam(new Date(), state);
     assert.deepEqual(installUrlOptions, returnedInstallUrlOptions);
   });
+
+  it('should detect old state values', async () => {
```
seratch, line 15: I've added the validation this time.

packages/oauth/src/state-stores/interface.ts
```diff
@@ -1,18 +1,40 @@
...
-// State object structure
+/**
+ * The data structure represented by the state parameter.
+ */
 export interface StateObj {
+
+  /**
+   * The timestamp that the state object was generated.
+   */
   now: Date;
```
seratch, line 11: Wanted to change this property name but we cannot do so for backward-compatibility

packages/oauth/src/stores/file-store.ts
```diff
@@ -30,7 +30,13 @@ export default class FileInstallationStore implements InstallationStore {
     const installationDir = this.getInstallationDir(enterprise?.id, team?.id);
 
     if (logger !== undefined) {
-      logger.info(`Storing installation in ${installationDir} for ${JSON.stringify({ enterprise, team, user })}`);
+      const dataForLogging = {
+        enterprise,
+        team,
+        // user object can include token values
+        user: { id: user.id },
```
seratch, line 37: While doing tests this time, I found that this info-level logging can print user tokens

seratch (commented):
comments for reviewers

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
 import { WebClientOptions } from '@slack/web-api';
 import { StateStore } from './state-stores';
 import { InstallationStore } from './stores';
+import { InstallURLOptions } from './install-url-options';
 
 export interface InstallProviderOptions {
+
+  /**
+   * Client ID, which can be found at https://api.slack.com/apps
```
filmaj, line 10: For this and client secret, perhaps the comment could be "which can be found under the Basic Information section of your application on https://api.slack.com/apps" ?

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+
+  /**
+   * handleCallback() verifies the state parameter if true (default: true)
+   */
   stateVerification?: boolean; // default true, disables state verification when false
-  installationStore?: InstallationStore; // default MemoryInstallationStore
+
+  /**
+   * handleCallback() skips checking browser cookies if true (default: false)
+   * Enabling this option is not really recommended.
```
filmaj, line 46:
```suggestion
   * Enabling this option is not recommended.
```

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+   */
+  legacyStateVerification?: boolean;
+
+  /**
+   * The cookie name used for setting state parameter value in cookies
+   */
+  stateCookieName?: string;
+
+  /**
+   * The expinary time for the state parameter value in cookies
```
filmaj, line 57:
```suggestion
   * The expiration time in seconds for the state parameter value stored via cookies
```

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
+   */
+  renderHtmlForInstallPath?: (url: string) => string;
+
+  /**
+   * The install path web page rendering will be skipped if true (default: false)
+   */
+  directInstall?: boolean; // default false, disables rendering "Add to Slack" page for /slack/install when true
+
+  /**
+   * The defalt is "v2" ("v1" is for "classic" apps)
```
filmaj, line 72:
```suggestion
   * The default is "v2" ("v1" is for "classic" apps)
```

filmaj: What specifically does this auth version refer to? Not the OAuth version, right? Is it the authentication procedure prior to introducing this state parameter verification built in mechanism in this PR?

packages/oauth/src/install-provider.spec.js
filmaj, line 735:
```suggestion
  it('should not fail if no state cookie returned and legacyStateVerification is enabled', async () => {
```

packages/oauth/src/install-provider.spec.js
```diff
@@ -612,22 +670,118 @@ describe('InstallProvider', async () => {
...
+      stateStore: fakeStateStore,
+      clientOptions: fooClientOptions,
+      legacyStateVerification: true,
+    });
+    const fakeState = 'fakeState';
+    const fakeCode = 'fakeCode';
+    const req = {
+      headers: {
+        host: 'example.com',
+        // cookie: `slack-app-oauth-state=${fakeState}`,
```
filmaj, line 760: Do we also want to check the case where legacyStateVerification=true and state cookie != state stored in state store?

packages/oauth/src/install-provider.ts
```diff
@@ -9,58 +9,91 @@ import {
...
- * @param stateStore - Replacement function for the built-in `stateStore`
- * @param stateVerification - Pass in false to disable state parameter verification
- * @param installationStore - Interface to store and retrieve installation data from the database
- * @param authVersion - Can be either `v1` or `v2`. Determines which slack Oauth URL and method to use
- * @param logger - Pass in your own Logger if you don't want to use the built-in one
- * @param logLevel - Pass in the log level you want (ERROR, WARN, INFO, DEBUG). Default is INFO
+ * InstallProvider Class. Refer to InsallProviderOptions interface for the details of constructor arguments.
  */
 export class InstallProvider {
+  // Manages state parameter values in the OAuth flow
```
filmaj, line 32:
```suggestion
  // Stores state issued to authorization server and verifies the value returned at redirection during OAuth flow to prevent CSRF
```

packages/oauth/src/install-provider.ts
```diff
@@ -9,58 +9,91 @@ import {
...
   public stateStore?: StateStore;
 
+  // Manages installation data, which can be called by both the OAuth flow and authorize() in event handling
   public installationStore: InstallationStore;
 
+  // Client ID, which can be found at https://api.slack.com/apps
```
filmaj, line 38: Same comment as above, perhaps we can be more specific as to where to find these values on the App Config page. Granted, it may be too specific. I do not have a strong feeling about this comment.

packages/oauth/src/install-provider.ts
```diff
@@ -9,58 +9,91 @@ import {
...
   public constructor({
     clientId,
     clientSecret,
     stateSecret = undefined,
     stateStore = undefined,
     stateVerification = true,
+    // this option is only for the backward-compatibility with v2.4 and older
+    legacyStateVerification = false,
+    stateCookieName = 'slack-app-oauth-state',
+    stateCookieExpirationSeconds = 600, // 10 minutes
```
filmaj, line 91: Nice, the RFC 6749 recommended maximum value 😄

packages/oauth/src/install-provider.ts
```diff
@@ -336,48 +436,65 @@ export class InstallProvider {
...
       // If state verification is enabled, attempt to verify, otherwise ignore
-      if (this.stateVerification && this.stateStore) {
-        // eslint-disable-next-line no-param-reassign
-        installOptions = await this.stateStore.verifyStateParam(new Date(), state);
+      if (this.stateVerification) {
+        try {
+          if (this.legacyStateVerification) {
+            // This mode is not enabled by default
+            // This option is for some of the existing developers that need time for migration
+            this.logger.warn('Enabling legacyStateVerification is not recommended as it does not properly work for OAuth CSRF protection. Please consider migrating from directly using InstallProvider#generateInstallUrl() to InstallProvider#handleInstallPath() for serving the install path.');
```
filmaj, line 467: Do we perhaps need to provide a bit more guidance on what this migration looks like for developers? How would they know what migration looks like? Is there a doc we should link them to?

filmaj (approved):
Phew, epic work here! Took me a while to review as I had to read RHC 6749 😬 It looks good to me but I would feel better if someone else also took the time to review this too.

packages/oauth/src/install-provider-options.ts
misscoded:
```suggestion
   * Client ID, which can be found under the Basic Information section of your application on https://api.slack.com/apps
```

packages/oauth/src/install-provider-options.ts
```diff
@@ -2,17 +2,95 @@ import { Logger, LogLevel } from '@slack/logger';
...
 
 export interface InstallProviderOptions {
+
+  /**
+   * Client ID, which can be found at https://api.slack.com/apps
+   */
   clientId: string;
+
+  /**
+   * Client Secret, which can be found at https://api.slack.com/apps
```
misscoded, line 15:
```suggestion
   * Client Secret, which can be found under the Basic Information section of your application on https://api.slack.com/apps
```

packages/oauth/src/state-stores/interface.ts
```diff
@@ -1,18 +1,40 @@
...
+ */
 export interface StateStore {
-  // Returned Promise resolves for a string which can be used as an
-  // OAuth state param.
-  // TODO: Revisit design. Does installOptions need to be encoded in state if metadata is static?
+
+  /**
+   * Generates a valid state parameter value, which can be decoded as a StateObj object
+   * by the verifyStateParam() method. This value may be stored on the server-side with expiration.
+   * Also, the InstallProvider verifies if this value is set in the installer's browser session.
```
misscoded, line 29:
```suggestion
   * The InstallProvider verifies if this value is set in the installer's browser session.
```

packages/oauth/src/state-stores/interface.ts
```diff
@@ -1,18 +1,40 @@
...
+   */
   generateStateParam: (installOptions: InstallURLOptions, now: Date) => Promise<string>;
 
-  // Returned Promise resolves for InstallURLOptions that were stored in the state
-  // param. The Promise rejects with a CodedError when the state is invalid.
+  /**
+   * Verifies the given state string value by trying to decode the value and
+   * build the passed InstallURLOptions object from the data.
+   * Also, this method verifies if the state value is not too old to detect reply attacks.
```
misscoded, line 36:
```suggestion
   * This method verifies if the state value is not too old to detect reply attacks.
```

misscoded (approved):

packages/oauth/src/install-provider.ts
seratch: No, it is still necessary but the code can be improved more. updated in the latest revision

packages/oauth/src/install-provider.ts
seratch: @filmaj 
The reason why I added this is to cover the use cases where developers cannot immediately enable the cookie validation for some reason. For instance, when we receive questions in GitHub issues (or directly from customers), we may reactively suggest this option for the developers (as a workaround). Thus, when an app enables this option, the maintainers of the app should know what they are doing and why. The migration timing and remaining tasks should be managed on the developer side.

That being said, we can add this option somewhere in docs. I don't think that we should highlight this one, though.

Lastly, this is not a manual migration. As the current behavior of the `@slack/oauth` is not correct in terms of OAuth flow handling in general. Thus, this is a bug fix, not a breaking change that requires developers' action.

packages/oauth/src/install-provider.ts
seratch: Confirmed the new implementation works better for error patterns

packages/oauth/src/install-provider.ts
```diff
@@ -9,58 +9,92 @@ import {
...
+  // Manages installation data, which can be called by both the OAuth flow and authorize() in event handling
   public installationStore: InstallationStore;
 
+  // Client ID, which can be found under the Basic Information section of your application on https://api.slack.com/apps
   private clientId: string;
 
+  // Client Secret, which can be found under the Basic Information section of your application on https://api.slack.com/apps
   private clientSecret: string;
 
+  // The defalt is "v2" ("v1" is for "classic" apps)
```
seratch, line 45:
```suggestion
  // The default is "v2" ("v1" is for "classic" apps)
```

packages/oauth/src/state-stores/interface.ts
```diff
@@ -1,18 +1,41 @@
...
+   * The InstallProvider verifies if this value is set in the installer's browser session.
+   */
   generateStateParam: (installOptions: InstallURLOptions, now: Date) => Promise<string>;
 
-  // Returned Promise resolves for InstallURLOptions that were stored in the state
-  // param. The Promise rejects with a CodedError when the state is invalid.
+  /**
+   * Verifies the given state string value by trying to decode the value and
+   * build the passed InstallURLOptions object from the data.
+   * This method verifies if the state value is not too old to detect reply attacks.
```
seratch, line 37:
```suggestion
   * This method verifies if the state value is not too old to detect replay attacks.
```

packages/oauth/src/state-stores/interface.ts
```diff
@@ -1,18 +1,41 @@
...
+   */
   generateStateParam: (installOptions: InstallURLOptions, now: Date) => Promise<string>;
 
-  // Returned Promise resolves for InstallURLOptions that were stored in the state
-  // param. The Promise rejects with a CodedError when the state is invalid.
+  /**
+   * Verifies the given state string value by trying to decode the value and
+   * build the passed InstallURLOptions object from the data.
+   * This method verifies if the state value is not too old to detect replay attacks.
+   * If the value is invalid, this method can throw InvalidStateError execption.
```
yujinakayama, line 38:
```suggestion
   * If the value is invalid, this method can throw InvalidStateError exception.
```

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +286,93 @@ export class InstallProvider {
    * The return value is an Array of Promises made up of the resolution of each token refresh attempt.
    */
   private async refreshExpiringTokens(tokensToRefresh: string[]): Promise<OAuthV2TokenRefreshResponse[]> {
-    const client = new WebClient(undefined, this.clientOptions);
+    const refreshPromises = tokensToRefresh.map((token) => this.resfreshExpiringToken(token));
```
yujinakayama, line 289:
```suggestion
    const refreshPromises = tokensToRefresh.map((token) => this.refreshExpiringToken(token));
```

packages/oauth/src/install-provider.ts
```diff
@@ -245,32 +286,93 @@ export class InstallProvider {
...
+    return (await Promise.all(refreshPromises))
+      .filter((res) => !(res instanceof Error))
+      .map((res) => res as OAuthV2TokenRefreshResponse);
+  }
 
-    const refreshPromises = tokensToRefresh.map(async (refreshToken) => await client.oauth.v2.access({
+  private async resfreshExpiringToken(refreshToken: string): Promise<OAuthV2TokenRefreshResponse | Error> {
```
yujinakayama, line 295:
```suggestion
  private async refreshExpiringToken(refreshToken: string): Promise<OAuthV2TokenRefreshResponse | Error> {
```

packages/oauth/src/install-provider.ts
seratch: Good catch. Thanks!

seratch (commented):
Thanks for your review! @yujinakayama

packages/oauth/src/state-stores/interface.ts
```diff
@@ -1,18 +1,41 @@
...
-  // Returned Promise resolves for a string which can be used as an
-  // OAuth state param.
-  // TODO: Revisit design. Does installOptions need to be encoded in state if metadata is static?
+
+  /**
+   * Generates a valid state parameter value, which can be decoded as a StateObj object
+   * by the verifyStateParam() method. This value may be stored on the server-side with expiration.
+   * The InstallProvider verifies if this value is set in the installer's browser session.
+   */
   generateStateParam: (installOptions: InstallURLOptions, now: Date) => Promise<string>;
```
yujinakayama, line 32: Although not directly related to this change, as a bolt-js user using custom state store, I've been wondering if encoding/decoding the entire InstallURLOptions into the state is necessary.

Currently the decoded InstallURLOptions returned by `verifyStateParam()` is used for the following purposes:

* [Passing `redirect_uri` for `oauth.v2.access` method](https://github.com/slackapi/node-slack-sdk/blob/e1adad09d93991cff3a54897a6985de360faeea9/packages/oauth/src/install-provider.ts#L425)
  * Maybe we can use `req.url` instead in `handleCallback()`?
* [Assigning `metadata` to `Installation`](https://github.com/slackapi/node-slack-sdk/blob/e1adad09d93991cff3a54897a6985de360faeea9/packages/oauth/src/install-provider.ts#L498)
* [Passing entire InstallURLOptions to `CallbackOptions.success/failure`](https://github.com/slackapi/node-slack-sdk/blob/e1adad09d93991cff3a54897a6985de360faeea9/packages/oauth/src/install-provider.ts#L512)
  * Do we really need this in the callbacks when we have `Installation`?

I feel the essential is only `InstallURLOptions.metadata`.

Is there any underlying reason for the design?

That said, I understand we cannot immediately change the public API (`StateStore` and `CallbackOptions`) until next major version bump.

seratch: @yujinakayama Thanks for the comment.
>as a bolt-js user using custom state store, I've been wondering if encoding/decoding the entire InstallURLOptions into the state is necessary.

I do understand your point here. Actually, bolt-python and bolt-java (the ones I implemented as the main developer) do not have this requirement at all. This is specific to this Node SDK package and bolt-js.

>Maybe we can use req.url in handleCallback() instead?

This is off-topic a bit but `req.url` in application code does not always return the publicly available full URL (e.g., in the case of apps behind a load balancer). Thus, knowing rediect_uri may be useful in some situations. That being said, this is not a common need. Also, if a developer would like to do this, including the URL as part of state parameter is not the only way to do so.

>I feel the essential is only InstallURLOptions.metadata.
>Is there any underlying reason for the design?

I've checked the bolt-js interface designer's design document again and found that the main intention of this design seemed to be just to transfer `metadata` to `/slack/oauth_redirect` URL (`handleCallback()` in code). This means that your question here is so reasonable. Going with InstallURLOptions here might be for flexibility to easily handle future enhancements but I cannot tell the exact reason behind that.

>That said, I understand we cannot immediately change the public API (StateStore and CallbackOptions) until next major version bump.

Yes, I'm glad to know that we're on the same page on this. In the long run, we may consider adding StateStore v2 that to make 3rd party developers' custom StateStore code much simpler. That being said, having two types of `StateStore` interface can be even more confusing (particularly for the developers that are new to Slack app development). Also, the improvement in this PR should enable developers to simply go with the default `StateStore` for the purpose of OAuth CSRF protection.

Considering all the above, my plans for the near future is to provide more flexibility to developers  by adding more callbacks in both `/slack/install` and `/slack/oauth_redirect`. The `metadata` in this SDK can be useful for some use cases but if you can set cookies, it should be more flexible and safer. Here is my proposal for it: https://github.com/slackapi/node-slack-sdk/issues/1438 If you have any comments on the design and plans, please feel free to write in there.

Applying the improvements in this PR plus having these callbacks in the next release should enable you to get rid of your own StateStore for customization (let me know if I'm missing something here).

Thanks again for your insights here!

seratch: >Also, the improvement in this PR should enable developers to simply go with the default StateStore for the purpose of OAuth CSRF protection.

Let me correct my comment a bit. I meant that, with the improvement in this PR, the default `ClearStateStore` may become safe in the acceptable level. However, the ideal implementation would be to use sever-side database to manage active state values and access the server-side data when `verifyStateParam()` is called. In Python and Java, we provide out-of-the-box solutions for this:

* https://github.com/slackapi/python-slack-sdk/tree/v3.15.1/slack_sdk/oauth/state_store
* https://github.com/slackapi/java-slack-sdk/tree/v1.20.0/bolt/src/main/java/com/slack/api/bolt/service/builtin

node-slack-sdk, specifically `@slack/oauth`, can provide something similar in future versions. I will create a new ticket issue for it later.

packages/oauth/src/install-provider-options.ts
filmaj: OK, thanks for clarifying, good to know. Not sure if it's appropriate to keep these links as breadcrumbs for the comment? Just for future maintainers who may not understand the difference between the versions or even which version this is referring to. Maybe we could have a comment like:

> * The default is "v2" (a.k.a. granular bot permissions), different from "v1" (a.k.a. "classic apps"). More details here: https://medium.com/slack-developer-blog/more-precision-less-restrictions-a3550006f9c3

packages/oauth/src/install-provider.ts
filmaj: OK, sounds good.

packages/oauth/src/install-provider-options.ts
seratch: @filmaj Sounds good. I've updated the comments on authVersion.

seratch:
@filmaj @misscoded @yujinakayama Thanks again for your great inputs here! I think that this PR already had a good amount of feedback. Let me merge it now.

packages/oauth/src/install-provider.ts
```diff
@@ -336,48 +449,65 @@ export class InstallProvider {
...
-        installOptions = await this.stateStore.verifyStateParam(new Date(), state);
+      if (this.stateVerification) {
+        try {
+          if (this.legacyStateVerification) {
+            // This mode is not enabled by default
+            // This option is for some of the existing developers that need time for migration
+            this.logger.warn('Enabling legacyStateVerification is not recommended as it does not properly work for OAuth CSRF protection. Please consider migrating from directly using InstallProvider#generateInstallUrl() to InstallProvider#handleInstallPath() for serving the install path.');
+          } else {
+            const stateInBrowserSession: string | undefined = extractCookieValue(req, this.stateCookieName);
+            if (!stateInBrowserSession || (stateInBrowserSession !== stateInQueryString)) {
```
odedpeer, line 483: In my use case the state is stored in a database. I created a custom State Store.

The client is a React app, and the backend is "AWS AppSync to Lambda" configuration, hence, not a nodejs web server.
I do not have nodejs `request` and `response` objects, nor access to the browser's cookies.

I am using `node-slack-sdk` to generate the install URL and to handle the callback, however, handleCallback is executed from a lambda function, without access to a browser.

`stateInBrowserSession` is always undefined in my case, and fails this check.

I am setting `legacyStateVerification` to true to avoid this.
I don't think the warning message is required in this case, and I ask you to keep `legacyStateVerification` since it is not only for `for backward-compatibility with v2.4 and olders` but also for backends that use custom state stores and do not have access to browser cookies

PR #1437, opened 2022-02-27
Update FileInstallationStore to behave more consistently

seratch:
###  Summary

This pull request improves the following points in the `@slack/oauth` package:

* Change all the write operations to use "sync" methods for atomicity
* Improve the exceptions coming from FileInstallationStore to include more details of the cause
* Add the corresponding tests for the above changes
* Simplify the rest in tests

### DIFF
--- a/packages/oauth/src/stores/file-store.spec.js
+++ b/packages/oauth/src/stores/file-store.spec.js
@@ -72,10 +72,15 @@ const storedOrgInstallation = {
 
 describe('FileInstallationStore', async () => {
 
-  let fsMakeDir, fsWriteFile, fsReadFileSync;
+  let fsMakeDir, fsMakeDirSync,
+    fsWriteFile, fsReadFileSync,
+    fsUnlink, fsReaddirSync;
 
   beforeEach(() => {
+    // Note that these sinon stubs affect the `os` package behaviors
+    // in the tests in this file
     fsMakeDir = sinon.stub(fs, 'mkdir').returns({});
+    fsMakeDirSync = sinon.stub(fs, 'mkdirSync').returns({});
     fsWriteFile = sinon.stub(fs, 'writeFile').returns({});
     fsReadFileSync = sinon.stub(fs, 'readFileSync').returns(Buffer.from(JSON.stringify(storedInstallation)));
     fsUnlink = sinon.stub(fs, 'unlink').returns({});
@@ -84,6 +89,7 @@ describe('FileInstallationStore', async () => {
 
   afterEach(() => {
     fsMakeDir.restore();
+    fsMakeDirSync.restore();
     fsWriteFile.restore();
     fsReadFileSync.restore();
     fsUnlink.restore();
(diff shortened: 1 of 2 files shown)


packages/oauth/src/stores/file-store.spec.js
```diff
@@ -72,10 +72,15 @@ const storedOrgInstallation = {
...
-  let fsMakeDir, fsWriteFile, fsReadFileSync;
+  let fsMakeDir, fsMakeDirSync,
+    fsWriteFile, fsReadFileSync,
+    fsUnlink, fsReaddirSync;
 
   beforeEach(() => {
+    // Note that these sinon stubs affect the `os` package behaviors
+    // in the tests in this file
     fsMakeDir = sinon.stub(fs, 'mkdir').returns({});
+    fsMakeDirSync = sinon.stub(fs, 'mkdirSync').returns({});
```
seratch, line 83: This stub had been missing

packages/oauth/src/stores/file-store.spec.js
```diff
@@ -92,24 +98,36 @@ describe('FileInstallationStore', async () => {
 
   it('should store the latest installation', async () => {
     const installationStore = new FileInstallationStore({ baseDir: os.tmpdir() });
-    const installer = new InstallProvider({ clientId, clientSecret, stateSecret, installationStore });
```
seratch, line 95: Except for the `authorize` tests, we don't need to instantiate `InstallProvider`.  In fact, this test uses only `installer.installationStore`.

packages/oauth/src/stores/file-store.ts
```diff
@@ -105,36 +105,28 @@ export default class FileInstallationStore implements InstallationStore {
...
   }
 }
 
 function writeToFile(filePath: string, data: string): void {
-  fs.writeFile(filePath, data, (err) => {
-    if (err !== null) {
-      throw new Error('There was an error writing to the InstallationStore');
-    }
-  });
+  fs.writeFileSync(filePath, data);
```
seratch, line 127: This needs to be "sync" to make the write operation atomic

packages/oauth/src/stores/file-store.ts
```diff
@@ -105,36 +105,28 @@ export default class FileInstallationStore implements InstallationStore {
...
 }
 
 function deleteFile(filePath: string): void {
-  fs.unlink(path.resolve(filePath), (err) => {
-    if (err !== null) {
-      throw new Error('Failed to delete installation from FileInstallationStore');
-    }
-  });
+  fs.unlinkSync(path.resolve(filePath));
```
seratch, line 131: This needs to be "sync" to make the write operation atomic

packages/oauth/src/stores/file-store.ts
```diff
@@ -105,36 +105,28 @@ export default class FileInstallationStore implements InstallationStore {
       filesToDelete = filesToDelete.concat(userFiles);
     }
 
-    try {
-      filesToDelete.map((filePath) => deleteFile(path.resolve(`${installationDir}/${filePath}`)));
-    } catch (err) {
-      throw new Error('Failed to delete installation from FileInstallationStore');
-    }
+    filesToDelete.forEach((filePath) => {
```
seratch, line 108: The error handling here should be done for each deletion

packages/oauth/src/stores/file-store.ts
```diff
@@ -83,7 +83,7 @@ export default class FileInstallationStore implements InstallationStore {
       }
       return installation;
     } catch (err) {
-      throw new Error('Failed to fetch data from FileInstallationStore');
+      throw new Error(`No installation data found (enterprise_id: ${query.enterpriseId}, team_id: ${query.teamId}, user_id: ${query.userId})`);
```
seratch, line 86: Updated to have more information

seratch (commented):
Revisited more and changed the title of this PR

misscoded (approved):

packages/oauth/src/stores/file-store.ts
filmaj: Doesn't this technically change the behaviour compared to the previous version? In the previous version, if `path.resolve` or `fs.unlink` would error, an exception was raised, whereas in this PR's version, no exception is thrown and an error is logged out instead.

filmaj (commented):
I have one question around changing the behaviour (logging out instead of throwing) that I'd like clarity on before I am comfortable moving forward. Thanks for adding the new tests though and improving the code overall!

packages/oauth/src/stores/file-store.ts
seratch: @filmaj Good point. I can revert the logic change here and add only log improvement this time.

seratch: This has been improved

filmaj (approved):
Nice, thanks! Not sure which approach is better (to throw or just to log) but if we do make a change there we should probably wait for a major version bump?

In any case: :shipit:

PR #1441, opened 2022-02-28
Fix #1439 Add built-in StateStore implementations using server-side database

### LINKED ISSUE #1439: Add built-in StateStore implementations using server-side database
Ad mentioned at https://github.com/slackapi/node-slack-sdk/pull/1436#discussion_r815564525, `@slack/oauth` library provides only `ClearStateStore` as the built-in implementations of `StateStore`. `ClearStateStore` does not rely on any server-side database. 

In some use case, more secure solutions would be necessary for complying with an organization's security standard. We can add more `StateStore` as built-in implementation (or as examples) in the future:

* `FileStateStore` - local file system based; this can be part of this library
* `MongooseStateStore` - Mongoose (MongoDB client) based; this should be part of bolt-js-extensions (we'll publish this in the future)
* `PrismaStateStore` - Prisma (RDB / MongoDB client) based; this should be part of bolt-js-extensions (we'll publish this in the future)
* `DynamoDBStateStore` - AWS DynamoDB based one; this should be part of bolt-js-extensions (we'll publish this in the future)
* `AmazonS3StateStore` - Amazon S3 based one; this should be part of bolt-js-extensions (we'll publish this in the future)
* Google Cloud and many more ...

#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [x] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

### Requirements

seratch:
This pull request resolves #1439 for the built-in support for `StateStore`. We may add more extensions for RDB and MongoDB in the future but we'll add those in a different repository and as different npm packages.

In this PR, the following changes are included:
* Add new `InvalidStateError` in errors.ts
* Add missing time validation in `ClearStateStore`, which is added in https://github.com/slackapi/node-slack-sdk/pull/1436
* Add `FileStateStore` as reference implementation of better `StateStore`
* Add testing utility for verifying `StateStore` behaviors

Once this PR is merged, I will rebase #1436.

###  Summary


packages/oauth/src/state-stores/clear-state-store.ts
```diff
@@ -1,23 +1,54 @@
...
-  public async generateStateParam(installOptions: InstallURLOptions, now: Date): Promise<string> {
-    return sign({ installOptions, now: now.toJSON() }, this.stateSecret);
+  public async generateStateParam(
+    installOptions: InstallURLOptions,
+    now: Date,
+  ): Promise<string> {
+    const source = {
+      installOptions,
+      now: now.toJSON(),
+      random: Math.floor(Math.random() * 1000000),
```
seratch, line 27: The reason why I added this is that, while doing tests with StateStore, I found that the generated JWT values with the same installOptions at the same timing can be exactly the same value. To prevent this happening, I've added additional random value here. This one is optional and no breaking changes to existing custom `StateStore` implementations.

packages/oauth/src/state-stores/file-state-store.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.stateExpirationSeconds = args.stateExpirationSeconds !== undefined ?
+      args.stateExpirationSeconds :
+      600;
+  }
+
+  public async generateStateParam(
+    installOptions: InstallURLOptions,
+    now: Date,
+  ): Promise<string> {
+    const state = randomUUID();
```
seratch, line 32: As JWT value is too long for file names, I chose UUID here. Also, the file content has simple JSON data structure. I think that there is no need to use JWT for server-side data.

seratch:
Oh, `crypto.randomUUID` was added in Node 14. Thus, the `FileStateStore` does not work with Node 12. I will change the state value generation logic later. If someone has a good suggestion for it, it'd be appreciated.

filmaj:
@seratch the de-facto RFC 4122 UUID generation module in nodejs is [uuid](https://www.npmjs.com/package/uuid). IT supports v1, v2 and v4 of UUID. Looks like `crypto.randomUUID` uses RFC 4122 v4.

seratch:
@filmaj Thanks. Yeah, I was also thinking about adding the dependency but I'm hesitating to add a new dependency only for this purpose. This may be fine, though.

packages/oauth/src/state-stores/file-state-store.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  private readFile(filename: string): StateObj | undefined {
+    const fullpath = path.resolve(`${this.baseDir}/${filename}`);
+    try {
+      const data = fs.readFileSync(fullpath);
+      if (data !== undefined) {
+        return JSON.parse(data.toString());
+      }
+      return undefined;
+    } catch (_) {
```
filmaj, line 97: I think just in case we should log something in this `catch`?

filmaj:
@seratch yeah, it's a tough decision. It is ~110kb of more code just to support node v12 😞

packages/oauth/src/state-stores/file-state-store.ts
```diff
@@ -0,0 +1,108 @@
...
+            (now.getTime() - generatedAt.getTime()) / 1000,
+          );
+          if (passedSeconds > this.stateExpirationSeconds) {
+            throw new InvalidStateError('The state value is already expired');
+          }
+          // return installOptions
+          return decoded.installOptions;
+        }
+      }
+    } finally {
```
filmaj, line 68: If I understand this correctly, the file corresponding to a specific state value will be deleted in the case that it exists but reading it returned undefined (which, as it stands, will only happen in case of a file reading, JSON parsing, or string casting error), or, if the state is expired.

Is this by design? To prevent any potential state tampering / hacking / investigation? Is my understanding of the intention here correct that, as soon as there is any issue in decoding or reading a state value, the state is invalidated and removed and thus the user would need to restart the installation dance?

packages/oauth/src/state-stores/file-state-store.ts
```diff
@@ -0,0 +1,108 @@
...
+    state: string,
+  ): Promise<InstallURLOptions> {
+    try {
+      if (this.findFile(state)) {
+        // decode the state using the secret
+        let decoded: StateObj | undefined;
+        try {
+          decoded = this.readFile(state);
+        } catch (e) {
+          const message = `Failed to load the data represented by the state parameter (error: ${e})`;
```
filmaj, line 52: Perhaps I am mistaken but I don't think this `catch` will ever run?

seratch:
There is no reason to stick with UUID here. Perhaps, generating long-enough random string plus checking if the same value does not exist in the file system should work fine.

filmaj:
@seratch agreed. Unless we assume that this implementation will be lifted and dropped into a production setting, if anything, we can leave a comment next to the state generation line to give a hint to developers to maybe use a more purposeful ID generation mechanism (like e.g. RFC 4122 v4).

filmaj (approved):
Looks good to me overall although the reference implementation left me with some questions.

I really like your spec-utils test generator, nicely done!

packages/oauth/src/state-stores/file-state-store.ts
seratch: Ah, true. I added `this.findFile(state)` later on. I can remove the `this.findFile(state)` from this code.

packages/oauth/src/state-stores/file-state-store.ts
seratch: You understanding here is correct. Regardless of the results of this method call, verification with the same state value should not be retried.

packages/oauth/src/state-stores/file-state-store.ts
seratch: Good point - will add some logs here

PR #1443, opened 2022-03-02
Rename sub directory for installation stores

seratch:
###  Summary

Now that we have `state-stores` directory for `StateStore`, renaming `stores` to `installation-stores` would be better in terms of consistency. 

As long as users import the interface / classes from `@slack/oauth`, this change does not affect any existing apps. Thus, I don't think that we can this a breaking change. What do you think?


filmaj (approved):
Agreed, this is fine. Breaking change would only apply to public APIs; importing specific package source files from sub directories is never recommended for exactly this reason.

seratch:
@filmaj Thanks for your quick response!

PR #1445, opened 2022-03-03
Updating beta releasing steps for maintainers

filmaj:
Now uses `npm publish --tag` and removes the need to invoke `npm dist-tag`.


srajiang (approved):
Sweet! Glad to be rid of the uncomfortable short period of time between `npm publish` and updating the dist-tag. I always imagine that person is installing the sdk right in that pocket of time.

srajiang:
Here you go! https://github.com/slackapi/bolt-js/pull/1348

PR #1478, opened 2022-05-04
Add "accessibility_label" field to Button interface

kpeters-cbsi:
###  Summary

The `Button` interface in `@slack/types` is missing the `accessibility_label` field as described in the [API reference](https://api.slack.com/reference/block-kit/block-elements#button). This PR adds that field.

NB: while I have read the discussion in #1227 , I think this PR is still appropriate as the missing field is part of the published API and is not present in the Typescript definitions.

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -241,6 +241,7 @@ export interface Button extends Action {
   url?: string;
   style?: 'danger' | 'primary';
   confirm?: Confirm;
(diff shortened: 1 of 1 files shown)


filmaj (approved):
Thanks and LGTM! Would like another maintainer's approval, though, before merging, as I'm not proficient in TS 😛

seratch (approved):

PR #1479, opened 2022-05-04
Tweaking logic for warning when using conversation APIs that are missing accessibility parameters

### LINKED ISSUE #1476: Fallback argument missing: confusing error text
Opening a new issue after reading slackapi/bolt-js#1249 which is closed.

@seratch, what was confusing on my end is that the error explicitly refers to a "\`fallback\` argument", so it leads to believe that the missing argument name is `fallback`, whereas it's actually `text`.

Maybe the warning message could be improved?

Suggestion:

```
The `text` argument is missing in the request payload for a chat.postMessage call - It's a best practice to always provide a `text` argument when posting a message which acts as a fallback. It is used in places where the content cannot be rendered such as: system push notifications, assistive technology such as screen readers, etc.
```

filmaj:
It is recommended to always provide a top-level `text` parameter for use in screen readers, etc. There is also a [legacy `fallback` field at the attachment level](https://api.slack.com/reference/messaging/attachments#legacy_fields).

This PR tweaks the logic slightly when displaying warnings to the user related to requests missing these fields. It also makes clear that `text` is the recommended field to always use.

This PR fixes #1476.

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -645,20 +645,24 @@ function warnIfFallbackIsMissing(method: string, logger: Logger, options?: WebAP
   const isTargetMethod = targetMethods.includes(method);
 
   const missingAttachmentFallbackDetected = (args: WebAPICallOptions) => Array.isArray(args.attachments) &&
-    args.attachments.some((attachment) => !attachment.fallback || attachment.fallback.trim() === 0);
+    args.attachments.some((attachment) => !attachment.fallback || attachment.fallback.trim() === '');
 
   const isEmptyText = (args: WebAPICallOptions) => args.text === undefined || args.text === null || args.text === '';
 
-  const buildWarningMessage = (missing: string) => `The \`${missing}\` argument is missing in the request payload for a ${method} call - ` +
-    `It's a best practice to always provide a \`${missing}\` argument when posting a message. ` +
-    `The \`${missing}\` is used in places where the content cannot be rendered such as: ` +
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -645,20 +645,24 @@ function warnIfFallbackIsMissing(method: string, logger: Logger, options?: WebAP
...
+    'It\'s a best practice to always provide a `text` argument when posting a message. ' +
+    'The `text` is used in places where the content cannot be rendered such as: ' +
     'system push notifications, assistive technology such as screen readers, etc.';
 
-  if (isTargetMethod && typeof options === 'object' && isEmptyText(options)) {
-    if (missingAttachmentFallbackDetected(options)) {
-      logger.warn(buildWarningMessage('fallback'));
-    } else {
-      logger.warn(buildWarningMessage('text'));
+  const buildMissingFallbackWarning = () => `Additionally, the attachment-level \`fallback\` argument is missing in the request payload for a ${method} call - ` +
```
filmaj, line 657: Could just as easily turn these two `build*Warning` methods into a simple string - if a reviewer has a preference, let me know, happy to change it.

seratch: Having these two warning messages look better than the current one to me 👍

seratch (approved):
LGTM 👍 

@filmaj If you have the bandwidth, can you apply the same (or similar) changes to Python and Java SDKs?

seratch:
@srajiang If you don't have any concerns on this PR, approve and merge this PR in your business hours tomorrow.

srajiang (approved):

filmaj:
@seratch yes, will work on this now!

PR #1496, opened 2022-06-13
Fix #1495 SocketModeClient still fails to automatically reconnect when apps.connections.open API returns an error code

### LINKED ISSUE #1495: SocketModeClient still fails to automatically reconnect when apps.connections.open API returns an error code
As reported at https://github.com/slackapi/bolt-js/issues/1428#issuecomment-1153495646, there is still a pattern that SocketModeClient's auto-reconnection does not work. See the repro steps below for details.

#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [x] `@slack/socket-mode`
- [ ] I don't know

#### The Slack SDK version

any version including 1.3.0

#### Node.js runtime version

any version

#### OS info

any

#### Steps to reproduce:

Copied from https://github.com/slackapi/bolt-js/issues/1428#issuecomment-1153495646

```
[INFO] A pong wasn't received from the server before the timeout of 5000ms!
[INFO] Reconnecting to Slack ...
[INFO] Going to establish a new connection to Slack ...

[ERROR] Failed to send a message as the client is not ready
[ERROR] An unhandled error occurred while Bolt processed (type: event_callback, error: Error: Failed to send a WebSocket message as the client is not ready)
[ERROR] Failed to send a message as the client is not ready
[ERROR] An unhandled error occurred while Bolt processed (type: event_callback, error: Error: Failed to send a WebSocket message as the client is not ready)

[ERROR] Faled to retrieve a new WSS URL for reconnection (error: Error: An API error occurred: internal_error)
[WARNING] Failed to start a Socket Mode connection (error: An API error occurred: internal_error)
```

And then endless loop of:
```
[ERROR] Failed to send a message as the client is not ready
[ERROR] An unhandled error occurred while Bolt processed (type: event_callback, error: Error: Failed to send a WebSocket message as the client is not ready)
```

Looking into similar occurrences, it is usually able to reconnect, so I think what caused the problem is:
```
Faled to retrieve a new WSS URL for reconnection
Failed to start a Socket Mode connection
```

### Expected result:

The SocketModeClient automatically reconnects when the server-side started working again.

### Actual result:

The reconnection does not happen.

seratch:
###  Summary

This pull request resolves #1495 . 

To test the behavior, you can modify `retrieveWSSURL` method as below plus run `cd examples && ./link.sh && npm start` to see how it works:

```typescript
  private reconnecting: boolean = false;

  private async retrieveWSSURL(): Promise<AppsConnectionsOpenResponse> {
    if (this.reconnecting && this.numOfConsecutiveReconnectionFailures % 2 === 0) {
      const e: WebAPIPlatformError = {
        name: 'XXX',
        message: 'An API error occurred: XXX',
        code: APICallErrorCode.PlatformError,
        data: {
          ok: false,
          error: 'internal_error',
        },
      };
      throw e;
    }
    this.reconnecting = true;
    try {
      this.logger.debug('Going to retrieve a new WSS URL ...');
      return await this.webClient.apps.connections.open();
    } catch (error) {
      this.logger.error(`Failed to retrieve a new WSS URL for reconnection (error: ${error})`);
      throw error;
    }
  }
```

The direct cause of the reconnection failure was that there is no logic to handle `connecting:reconnecting` in the current code. When you use a sub state machine in this state machine library, it has a state hierarchy and your code needs to have the comprehensive set of handlers for all the state sets.

To prevent this type of coding error happening again in the future (actually, this pattern errors existed much more before 1.3.0 release), I've added enums for the sub state machines - ConnectingState and ConnectedState. With these, passing a state to a sub machine is type-safe. Passing the top-level State enum to a sub state machine does not compile anymore.

### DIFF
--- a/packages/socket-mode/src/SocketModeClient.ts
+++ b/packages/socket-mode/src/SocketModeClient.ts
@@ -22,15 +22,23 @@ const packageJson = require('../package.json'); // eslint-disable-line import/no
 
 // These enum values are used only in the state machine
 enum State {
-  Ready = 'ready',
   Connecting = 'connecting',
   Connected = 'connected',
+  Reconnecting = 'reconnecting',
+  Disconnecting = 'disconnecting',
+  Disconnected = 'disconnected',
+  Failed = 'failed',
+}
+enum ConnectingState {
   Handshaking = 'handshaking',
   Authenticating = 'authenticating',
   Authenticated = 'authenticated',
   Reconnecting = 'reconnecting',
-  Disconnecting = 'disconnecting',
-  Disconnected = 'disconnected',
+  Failed = 'failed',
+}
+enum ConnectedState {
+  Preparing = 'preparing',
+  Ready = 'ready',
   Failed = 'failed',
 }
 
@@ -135,12 +143,12 @@ export class SocketModeClient extends EventEmitter {
     this.stateMachine.handle(Event.Start);
     // Return a promise that resolves with the connection information
     return new Promise((resolve, reject) => {
-      this.once(State.Authenticated, (result) => {
+      this.once(ConnectingState.Authenticated, (result) => {
         this.removeListener(State.Disconnected, reject);
         resolve(result);
       });
       this.once(State.Disconnected, (err) => {
-        this.removeListener(State.Authenticated, resolve);
+        this.removeListener(ConnectingState.Authenticated, resolve);
         reject(err);
       });
     });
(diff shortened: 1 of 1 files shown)


mwbrooks (approved):
Thanks for continuing to harden the Socket Mode package!

Is this a new, recent error or one that has existed for a while but we've only become aware of?
```
[ERROR] Failed to retrieve a new WSS URL for reconnection (error: Error: An API error occurred: internal_error)
```

filmaj (approved):
LGTM! I like the split between groups of states 👍 :shipit:

PR #1497, opened 2022-06-15
fix(install-provider): do not duplicate installation if tokens werent…

omermizr:
… refreshed

###  Summary

Today, if a token is expired we try to refresh it and store the installation - but we call `storeInstallation` even if no tokens were successfully refreshed. This PR aims to avoid that so we don't store the same installation over and over again.

* [ ] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [ ] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

### DIFF
--- a/packages/oauth/src/install-provider.ts
+++ b/packages/oauth/src/install-provider.ts
@@ -238,40 +238,45 @@ export class InstallProvider {
               'if it has a refresh token';
             throw new UnknownError(errorMessage);
           }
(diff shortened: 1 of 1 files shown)


seratch (approved):
LGTM 👍 Thanks for fixing this! We will release a new patch version shortly once we merge this.

Issue #1506, opened 2022-07-03
handleInstallPath throws an error when stateVerification is false


rockingskier:
I have my app enabled for org-wide installs.
Org apps require `stateVerification` is set to `false`, however, this breaks the `/slack/install` handler.

A `stateStore` is only created if you have `stateVerification=true`: https://github.com/slackapi/node-slack-sdk/blob/main/packages/oauth/src/install-provider.ts#L133

Down the line in `handleInstallPath` it bails out when no `stateStore` exists: https://github.com/slackapi/node-slack-sdk/blob/main/packages/oauth/src/install-provider.ts#L340-L342

I believe this was introduced in #1436.

It's very possible I'm missing something but it was working with `"@slack/bolt": "^3.10.0",`.

#### Packages:

- [ ] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [x] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

### Reproducible in:
Definitely 2.5.3
Probably from 2.5.0 (this was the first release after #1436)

```bash
"slack/bolt": "^3.11.3",
"slack/logger": "^3.0.0",
"slack/web-api": "^6.7.2",
```
#### Node.js runtime version

`v16.14.0`

#### OS info

```
ProductName:	macOS
ProductVersion:	11.6.7
```

#### Steps to reproduce:

1. Enable the app for org installation
2. Pass `stateVerification` as `false`
```
new ExpressReceiver({
    // config
    installerOptions: {
      stateVerification: false,
    },
    // more config
})
```
3. Run the app
4. Visit `/slack/install`

### Expected result:

A nice "Add to Slack" button

### Actual result:
```
Error: StateStore is not properly configured
    at new GenerateInstallUrlError (/MY_APP/node_modules/@slack/oauth/src/errors.ts:22:1)
    at InstallProvider.<anonymous> (/MY_APP/node_modules/@slack/oauth/src/install-provider.ts:341:13)
    at step (/MY_APP/node_modules/@slack/oauth/dist/install-provider.js:44:23)
    at Object.next (/MY_APP/node_modules/@slack/oauth/dist/install-provider.js:25:53)
    at /MY_APP/node_modules/@slack/oauth/dist/install-provider.js:19:71
    at new Promise (<anonymous>)
    at __awaiter (/MY_APP/node_modules/@slack/oauth/dist/install-provider.js:15:12)
[...]

PR #1507, opened 2022-07-03
Allow `handlerInstallPath` to work without `stateStore`

rockingskier:
###  Summary
#1506

This is a bit of a guess.  It worked when I edited the compiled code in my `node_modules` 🤮 but I wouldn't be surprised if there was a better way :)

I can't run the tests locally so I'm fully expecting them to fail.  This is more of a "here's what I found" PR.

### DIFF
--- a/packages/oauth/src/install-provider.ts
+++ b/packages/oauth/src/install-provider.ts
@@ -337,10 +337,6 @@ export class InstallProvider {
     const _printableOptions = JSON.stringify(_installOptions);
     this.logger.debug(`Running handleInstallPath() with ${_printableOptions}`);
 
-    if (this.stateStore === undefined) {
-      throw new GenerateInstallUrlError('StateStore is not properly configured');
-    }
-
     try {
       let shouldProceed = true;
       if (options?.beforeRedirection !== undefined) {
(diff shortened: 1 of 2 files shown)


packages/oauth/src/install-provider.ts
```diff
@@ -350,30 +346,33 @@ export class InstallProvider {
...
-      if (res.getHeader('Set-Cookie')) {
-        // If the cookies already exist
-        const existingCookies = res.getHeader('Set-Cookie') || [];
-        const allCookies: string[] = [];
-        if (Array.isArray(existingCookies)) {
-          allCookies.push(...existingCookies);
-        } else if (typeof existingCookies === 'string') {
-          allCookies.push(existingCookies);
+      let state;
+      if (this.stateStore) {
```
seratch, line 350: Can you use `this.stateVerification` flag instead and place L340 - L342 that you've deleted inside the if statement?
```suggestion
      if (this.stateVerification) {
```

packages/oauth/src/install-provider.ts
```diff
@@ -27,7 +27,7 @@ import defaultRenderHtmlForInstallPath from './default-render-html-for-install-p
 import { InstallPathOptions } from './install-path-options';
 
 /**
- * InstallProvider Class. Refer to InsallProviderOptions interface for the details of constructor arguments.
+ * InstallProvider Class. Refer to InstallProviderOptions interface for the details of constructor arguments.
```
seratch, line 30: Thanks for detecting errors in the comments!

seratch (changes requested):
Could you check my comments?

packages/oauth/src/install-provider.ts
rockingskier: No worries, my IDE does most of the work :)

packages/oauth/src/install-provider.ts
rockingskier: I did that initially but it makes TS sad due to `this.stateStore` being optional.
Screenshot 2022-07-04 at 07 30 31

I can do: 
* `if (this.stateStore)`
* `if(this.stateVerification && this.stateStore)
* `await this.stateStore?.generateStateParam` but this gets complicated as `state` then becomes optional

rockingskier: Sounds good 👍  Done.

seratch (changes requested):
Can you resolve the TS compilation error?

rockingskier:
Sorted my local env and now I'm not coding blind :)

Fixed, tests passing locally. Hopefully all good to go.

seratch (approved):
Looks great to me! Thanks for contributing to this SDK!

Issue #1191, opened 2021-03-22
Add `admin.analytics.getFile` API support


seratch:
### Description

https://api.slack.com/methods/admin.analytics.getFile is not yet supported in this SDK. 

For the implementation, we should be aware that this endpoint is a bit different from others. The response body can be gzipped binary data for successful response. If the response has an error code, the server-side returns text JSON data instead.

See also:
* https://github.com/slackapi/java-slack-sdk/pull/602
* https://github.com/slackapi/python-slack-sdk/pull/865

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

AdriannaBeck:
Hello, is there any estimate of when this feature may be available?

seratch:
Fixed by #1515

PR #1515, opened 2022-07-15
Added in functionality for admin.analytics.getFile method

hello-ashleyintech:
###  Summary

Adding in `admin.analytics.getFile` support as mentioned in: https://github.com/slackapi/node-slack-sdk/issues/1191

Link to docs for the method is here: https://api.slack.com/methods/admin.analytics.getFile. This method is a little more unique to other methods -
* In the event of a successful response, this method returns a binary string for a compressed JSON.gz file. Unlike other methods, there is no `ok` property to note its success. Due to the successful response being binary, it needs to be sent as an ArrayBuffer response to be able to be parsed properly - this was also added in this PR
* In the event of a failed response, this method has a similar JSON structure as errors for other methods: `{ ok: false, error: "error message" } `

Tested this functionality + tests locally on my own enterprise grid account.

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -11,6 +11,7 @@ import axios, { AxiosInstance, AxiosResponse } from 'axios';
 import FormData from 'form-data';
 import isElectron from 'is-electron';
 import zlib from 'zlib';
+import { TextDecoder } from 'util';
 
 import { Methods, CursorPaginationEnabled, cursorPaginationEnabledMethods } from './methods';
 import { getUserAgent } from './instrument';
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -12,6 +12,11 @@ import FormData from 'form-data';
 import isElectron from 'is-electron';
 import zlib from 'zlib';
 import { TextDecoder } from 'util';
+import {
+  MemberDetails,
+  PublicChannelDetails,
+  PublicChannelMetadataDetails,
+} from './response';
 
 import { Methods, CursorPaginationEnabled, cursorPaginationEnabledMethods } from './methods';
 import { getUserAgent } from './instrument';
@@ -234,7 +239,7 @@ export class WebClient extends Methods {
       team_id: this.teamId,
       ...options,
     }, headers);
-    const result = this.buildResult(response);
+    const result = await this.buildResult(response);
 
     // log warnings in response metadata
     if (result.response_metadata !== undefined && result.response_metadata.warnings !== undefined) {
@@ -263,6 +268,8 @@ export class WebClient extends Methods {
 
     // If result's content is gzip, "ok" property is not returned with successful response
     if (!result.ok && (response.headers['content-type'] !== 'application/gzip')) {
+      throw platformErrorFromResult(result as (WebAPICallResult & { error: string; }));
+    } else if ('ok' in result && result.ok === false) {
       throw platformErrorFromResult(result as (WebAPICallResult & { error: string; }));
     }
 
@@ -543,7 +550,7 @@ export class WebClient extends Methods {
    * @param response - an http response
    */
   // eslint-disable-next-line class-methods-use-this
-  private buildResult(response: AxiosResponse): WebAPICallResult {
+  private async buildResult(response: AxiosResponse): Promise<WebAPICallResult> {
     let { data } = response;
     const isGzipResponse = response.headers['content-type'] === 'application/gzip';
 
@@ -551,13 +558,25 @@ export class WebClient extends Methods {
     if (isGzipResponse) {
       // admin.analytics.getFile will return a Buffer that can be unzipped
       try {
-        const unzippedData = zlib.unzipSync(data).toString().split('\n');
-        const fileData: string[] = [];
-        unzippedData.forEach((dataset) => {
-          if (dataset && dataset.length > 0) {
-            fileData.push(JSON.parse(dataset));
-          }
-        });
+        const unzippedData = await new Promise((resolve, reject) => {
+          zlib.unzip(data, (err, buf) => {
+            if (err) {
+              return reject(err);
+            }
+            return resolve(buf.toString().split('\n'));
+          });
+        }).then((res) => res)
+          .catch((err) => {
+            throw err;
+          });
+        const fileData: Array<MemberDetails | PublicChannelDetails | PublicChannelMetadataDetails> = [];
+        if (Array.isArray(unzippedData)) {
+          unzippedData.forEach((dataset) => {
+            if (dataset && dataset.length > 0) {
+              fileData.push(JSON.parse(dataset));
+            }
+          });
+        }
         data = { file_data: fileData };
       } catch (err) {
         data = { ok: false, error: err };
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -13,9 +13,9 @@ import isElectron from 'is-electron';
 import zlib from 'zlib';
 import { TextDecoder } from 'util';
 import {
-  MemberDetails,
-  PublicChannelDetails,
-  PublicChannelMetadataDetails,
+  AdminAnalyticsMemberDetails,
+  AdminAnalyticsPublicChannelDetails,
+  AdminAnalyticsPublicChannelMetadataDetails,
 } from './response';
 
 import { Methods, CursorPaginationEnabled, cursorPaginationEnabledMethods } from './methods';
@@ -267,6 +267,8 @@ export class WebClient extends Methods {
     }
 
     // If result's content is gzip, "ok" property is not returned with successful response
+    // TODO: look into simplifying this code block to only check for the second condition
+    // if an { ok: false } body applies for all API errors
     if (!result.ok && (response.headers['content-type'] !== 'application/gzip')) {
       throw platformErrorFromResult(result as (WebAPICallResult & { error: string; }));
     } else if ('ok' in result && result.ok === false) {
@@ -569,7 +571,9 @@ export class WebClient extends Methods {
           .catch((err) => {
             throw err;
           });
-        const fileData: Array<MemberDetails | PublicChannelDetails | PublicChannelMetadataDetails> = [];
+        const fileData: Array<AdminAnalyticsMemberDetails |
+        AdminAnalyticsPublicChannelDetails |
+        AdminAnalyticsPublicChannelMetadataDetails> = [];
         if (Array.isArray(unzippedData)) {
           unzippedData.forEach((dataset) => {
             if (dataset && dataset.length > 0) {
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -10,6 +10,7 @@ import pRetry, { AbortError } from 'p-retry';
 import axios, { AxiosInstance, AxiosResponse } from 'axios';
 import FormData from 'form-data';
 import isElectron from 'is-electron';
+import zlib from 'zlib';
 
 import { Methods, CursorPaginationEnabled, cursorPaginationEnabledMethods } from './methods';
 import { getUserAgent } from './instrument';
@@ -259,7 +260,13 @@ export class WebClient extends Methods {
       });
     }
 
-    if (!result.ok) {
+    // If result's content is gzip, "ok" property is not returned with successful response
+    // In this case, check for the presence of "ok" property and if it's false
+    if (response.headers['content-type'] === 'application/gzip') {
+      if ('ok' in result && result.ok === false) {
+        throw platformErrorFromResult(result as (WebAPICallResult & { error: string; }));
+      }
+    } else if (!result.ok) {
       throw platformErrorFromResult(result as (WebAPICallResult & { error: string; }));
     }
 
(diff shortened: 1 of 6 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -536,6 +550,23 @@ export class WebClient extends Methods {
   private buildResult(response: AxiosResponse): WebAPICallResult {
     let { data } = response;
 
+    // Check for GZIP response - if so, it is a successful response from admin.analytics.getFile
+    if (response.headers['content-type'] === 'application/gzip') {
+      // admin.analytics.getFile will return a Buffer that can be unzipped
+      try {
+        const unzippedData = zlib.unzipSync(data).toString().split('\n');
+        const fileData: string[] = [];
```
hello-ashleyintech, line 558: Does this need to be a different type since we're pushing objects from `JSON.parse()` onto it? I tested accessing the output of this and I'm still able to access the objects normally (they aren't converted to strings and are successfully pushed into this array) but I'm wondering if there's a better convention for this?

I saw this convention as highlighted here (https://stackoverflow.com/questions/53964694/how-do-you-declare-an-array-of-objects-inside-typescript) for how to declare an array of objects, but it seems like it would be very involved and complex to declare this interface since we're expecting a large object for each line we're pushing into this array and also since I'm not sure if the output of each line of data from this API call is all expected to be the same structure of object. Would love to get thoughts on this!

prod-server-integration-tests/test/admin-web-api-analytics.js
```diff
@@ -0,0 +1,58 @@
...
+  ],
+});
+
+describe('admin.analytics.* Web API', function () {
+  // org-level admin user's token
+  const orgAdminClient = new WebClient(process.env.SLACK_SDK_TEST_GRID_ORG_ADMIN_USER_TOKEN, { logger, });
+
+  describe('admin.analytics.getFile', function () {
+    it('should get file data', async function () {
+      const body = { type: 'public_channel', date: '2022-07-06' };
```
hello-ashleyintech, line 19: these dates are hard-coded for the account i used for testing - should I add a note here about how these dates may need to be adjusted for the tests to be able to pass for others' accounts?

packages/web-api/src/WebClient.ts
seratch: You can place an interface for the each row data under [the `response` directory](https://github.com/slackapi/node-slack-sdk/tree/main/packages/web-api/src/response). Perhaps, the name can be `AdminAnalyticsGetFileResponseDetails` or so. Also, the auto-generated [`AdminAnalyticsGetFileResponse.ts`](https://github.com/slackapi/node-slack-sdk/blob/%40slack/web-api%406.7.2/packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts) does not have the file data property. Let's skip generating the interface from now on and modify it to have the details property. To prevent future modification by the code generator script, can you add if statement to check if the filename contains admin.analytics.getFile and skip the file generation for it here? https://github.com/slackapi/node-slack-sdk/blob/%40slack/web-api%406.7.2/scripts/code_generator.rb#L56

As for the interface details, if we don't have the bandwidth to work on the details right now, it's okay to go with any type as the first step. But if you want to complete the work this time, you can check the following resources:
* https://api.slack.com/methods/admin.analytics.getFile
* https://github.com/slackapi/java-slack-sdk/blob/v1.24.0/slack-api-client/src/main/java/com/slack/api/methods/response/admin/analytics/AdminAnalyticsGetFileResponse.java#L132-L476

The API document page for this endpoint is well-written. Also, you can refer to the Java SDK code to learn what properties are optional. This endpoint offers three response data patterns (1: public_channel, 2: member, 3: member+metadata_only). So, I think that there are three approaches to define the types:
1. Use unified interface covering all the patterns (most of the properties can be optional)
2. Define two types for the types (AdminAnalyticsGetFileResponseMemberDetails, AdminAnalyticsGetFileResponsePublicChannelDetails) and use a union type for the array
3. Define three types for covering all the possible patterns (~MemberDetails, ~PublicChannelMetadataOnlyDetails, ~PublicChannelDetails) and use a union type for the array

My suggestion here is either Option2 or Option3. Option2 should be great enough perhaps, but if we go with Option3, that'd be perfect for all use cases. Lastly, I know the suggested type names here are very long. I am fine to rename them if you would like to do so.

prod-server-integration-tests/test/admin-web-api-analytics.js
seratch: Yeah, it should be great for mitigating future maintainers' confusion!

seratch (commented):
Great work!

packages/web-api/src/WebClient.ts
```diff
@@ -259,7 +261,8 @@ export class WebClient extends Methods {
       });
     }
 
-    if (!result.ok) {
+    // If result's content is gzip, "ok" property is not returned with successful response
+    if (!result.ok && (response.headers['content-type'] !== 'application/gzip')) {
```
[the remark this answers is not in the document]
hello-ashleyintech: ahh yes, this is a good callout! 🙌  this is actually what i had originally, but I realized that when this API call fails, it returns a regular JSON response, not a GZIP (GZIP only returns on success), so the current logic should cover that use case!

packages/web-api/src/WebClient.ts
```diff
@@ -535,6 +545,28 @@ export class WebClient extends Methods {
   // eslint-disable-next-line class-methods-use-this
   private buildResult(response: AxiosResponse): WebAPICallResult {
     let { data } = response;
+    const isGzipResponse = response.headers['content-type'] === 'application/gzip';
+
+    // Check for GZIP response - if so, it is a successful response from admin.analytics.getFile
+    if (isGzipResponse) {
+      // admin.analytics.getFile will return a Buffer that can be unzipped
+      try {
+        const unzippedData = zlib.unzipSync(data).toString().split('\n');
```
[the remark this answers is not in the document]
hello-ashleyintech: this is also a great callout! i was originally using the async method `zlib.unzip()` but was running into issues with accessing the file data since it outputs that into a callback and the data couldn't be reached outside the callback 😢 it seemed complex to have all of the remaining logic that would use this file data within the callback. with that being said, if there's another workaround I could use here, i'm definitely open to it! i'll take a stab at converting this to an async function and see what happens 👀

prod-server-integration-tests/test/admin-web-api-analytics.js
hello-ashleyintech: my hesitation with auto-generating the dates is that it may still confuse maintainers since the dates for successful tests depend largely on the enterprise grid account they're using to test and how long it's been active (ex: I created an enterprise grid account 2022-07-05, tried to run this on that day, and got an error that the file wasn't yet available since the first bit of data is available the day after the enterprise grid account is created) 🤔 a good middle ground here might be to add some comments making sure they're using dates that work with the life cycle of the account they're using to test and also noting nuances like the 24 hour period for data being available - would love to get thoughts on this!

packages/web-api/src/WebClient.ts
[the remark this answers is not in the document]
filmaj: Ah great, thanks for confirming!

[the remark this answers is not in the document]
hello-ashleyintech: update: I did end up adding in some extra logic here similar to what you suggested to account for zlib errors / errors that were thrown when parsing a GZIP!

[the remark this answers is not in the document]
hello-ashleyintech: a follow up thought: if we can expect every error message in the node SDK to indicate an error by having the response have an `ok: false` property, we could honestly drop the `if / else if` structure here and just use `if ('ok' in result && result.ok === false)` as the conditional

prod-server-integration-tests/test/admin-web-api-analytics.js
hello-ashleyintech: added in the comments for the tests  - let me know your thoughts on them and if they're explicit enough!

packages/web-api/src/WebClient.ts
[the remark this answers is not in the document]
filmaj: That's a good point! I am not familiar enough with the API responses to be able to determine if that applies to all APIs. At a minimum, though, would be worthwhile to add this thought as a potential comment / TODO right next to this code. Think that's a worthwhile area of investigation / improvement that we could pursue later.

scripts/code_generator.rb
```diff
@@ -57,25 +57,27 @@ def append_to_index(root_class_name, index_file)
...
     prev_c = nil
     filename = json_path.split('/').last.gsub(/\.json$/, '')
-    filename.split('').each do |c|
-      if prev_c.nil? || prev_c == '.'
-        root_class_name << c.upcase
-      elsif c == '.'
-        # noop
-      else
-        root_class_name << c
+    if !filename.include? "admin.analytics.getFile"
```
hello-ashleyintech, line 60: @seratch let me know if this location for the `if` statement is what you had in mind to skip the `AdminAnalyticsGetFile` generation! I tested the script and it skipped it, but it also didn't add it into the `response/index.ts` file as it was rewriting it - is this expected behavior? also, since I'm importing the interfaces into `packages/web-api/src/response/index.ts` manually to access them in `WebClient.ts` (since `WebClient` cannot import from the `AdminAnalyticsGetFileResponse` file directly), is this ok since this file can be regenerated by running the script and will override those imports?

prod-server-integration-tests/test/admin-web-api-analytics.js
filmaj: Fair enough! That's fine 😄

packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts
```diff
@@ -10,6 +10,7 @@
 
 import { WebAPICallResult } from '../WebClient';
 export type AdminAnalyticsGetFileResponse = WebAPICallResult & {
+  file_data?:         Array<MemberDetails|PublicChannelDetails|PublicChannelMetadataDetails>[];
```
filmaj, line 13: Nice!

filmaj (approved):
Awesome work, looks great to me! Well done

packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts
```diff
@@ -20,3 +21,73 @@ export type AdminAnalyticsGetFileResponse = WebAPICallResult & {
 export interface ResponseMetadata {
   messages?: string[];
 }
+
+export interface MemberDetails {
+	enterprise_id:                 string;
```
seratch, line 26: Can you make the indentation consistent (meaning two space indents)?

packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts
```diff
@@ -20,3 +21,73 @@ export type AdminAnalyticsGetFileResponse = WebAPICallResult & {
 export interface ResponseMetadata {
   messages?: string[];
 }
+
+export interface MemberDetails {
```
seratch, line 25: What do you think naming like `AdminAnalyticsMemberDetails` or so? `MemberDetails` cannot be exposed to external as it's too general.

packages/web-api/src/response/index.ts
```diff
@@ -1,4 +1,4 @@
-export { AdminAnalyticsGetFileResponse } from './AdminAnalyticsGetFileResponse';
+export { AdminAnalyticsGetFileResponse, MemberDetails, PublicChannelDetails, PublicChannelMetadataDetails } from './AdminAnalyticsGetFileResponse';
```
seratch, line 1: TODO: This line needs to be manually added in the code generator script (the Ruby script)

scripts/code_generator.rb
seratch: @hello-ashleyintech Good point. The index.ts file needs to be updated correctly. As I mentioned above, you need to add the manually crafted line for AdminAnalyticsGetFileResponse in this script. You can add a new line to do that inside this if statement.

seratch (changes requested):
Thanks for the updates. Great progress

scripts/code_generator.rb
```diff
@@ -48,6 +48,12 @@ def append_to_index(root_class_name, index_file)
       index_f.puts("export { #{root_class_name} } from './#{root_class_name}';")
     end
   end
+
+  def append_multiple_classes_to_index(root_class_name, root_file_name, index_file)
+    File.open(index_file, 'a') do |index_f|
+      index_f.puts("export { #{root_class_name} } from './#{root_file_name}';")
+    end
+  end
```
seratch, line 56: How about renaming the first arg and expecting an array? The `classes` needs to be an array in the following suggestion:
```suggestion
  def append_multiple_classes_to_index(classes, root_file_name, index_file)
    File.open(index_file, 'a') do |index_f|
      index_f.puts("export { #{classes.join(',')} } from './#{root_file_name}';")
    end
  end
```

scripts/code_generator.rb
```diff
@@ -57,25 +63,30 @@ def append_to_index(root_class_name, index_file)
...
     prev_c = nil
     filename = json_path.split('/').last.gsub(/\.json$/, '')
-    filename.split('').each do |c|
-      if prev_c.nil? || prev_c == '.'
-        root_class_name << c.upcase
-      elsif c == '.'
-        # noop
-      else
-        root_class_name << c
+    if !filename.include? "admin.analytics.getFile"
```
seratch, line 66: Can you make the change similar by early returning? 

```ruby
    if filename.include? "admin.analytics.getFile"
      classes_to_export = [
        "AdminAnalyticsGetFileResponse",
        "AdminAnalyticsMemberDetails",
        "AdminAnalyticsPublicChannelDetails",
        "AdminAnalyticsPublicChannelMetadataDetails",
        "AdminAnalyticsGetFileResponse",
      ]
      ts_writer.append_multiple_classes_to_index(classes_to_export, index_file)
      return
    end
```

seratch (changes requested):
Great! Two last things..

seratch (approved):
LGTM! Once the CI builds pass, you can merge this PR (please don't forget squashing the commits into one when clicking the merge button)

seratch:
Okay, this PR is ready to merge. Thanks for adding this feature!

PR #1526, opened 2022-08-29
Add interactivity pointer

hello-ashleyintech:
###  Summary

Events like `block_action`, `views_*` events that are sent with interactivity payloads aka (function, function_data) will no longer contain `trigger_id` but instead a new field called `interactivity_pointer`. The motivation is to move away from the legacy term "trigger" as it doesn't actually refer to the 2.0 concept of trigger.

This doesn't affect those same events if they are not sent with interactivity payload data (aka regular 1.0 block actions and views events), so there shouldn't be an impact on existing devs / apps.


packages/web-api/src/methods.ts
```diff
@@ -2104,12 +2104,14 @@ export interface UsersProfileSetArguments extends WebAPICallOptions, TokenOverri
 }
 
 export interface ViewsOpenArguments extends WebAPICallOptions, TokenOverridable {
-  trigger_id: string;
```
hello-ashleyintech, line 2107: @srajiang - In order to accept either `trigger_id` or `interactivity_pointer` I made them both optional - is there a better way to handle this or should this work? Ideally it would require at least one of these

docs/_packages/web_api.md
```diff
@@ -246,7 +246,7 @@ pattern for people that use _functional programming_.
 ---
 
 ### Opening modals
-[Modals](https://api.slack.com/block-kit/surfaces/modals) can be created by calling the `views.open` method. The method requires you to pass a valid [view payload](https://api.slack.com/reference/block-kit/views) in addition to a `trigger_id`, which can be obtained when a user invokes your app using a slash command, clicking a button, or using [another interactive action](https://api.slack.com/reference/messaging/interactive-components).
```
hello-ashleyintech, line 249: Would love any feedback here on terminology for 1.0 vs 2.0 - went with "with CLI" and "not CLI" but let me know if there's a better way to phrase this! 😬

packages/web-api/src/methods.ts
srajiang: @hello-ashleyintech  I think it would still be good to require at least one. So setting both to optional might let devs fail by setting neither. 

What do you think about logical XOR, so that it must be either trigger_id OR interactivity but not both or neither? There seem to be a [few ways folks in SO have discussed tackling this](https://stackoverflow.com/questions/44425344/typescript-interface-with-xor-barstring-xor-cannumber) which might be a good resource for this.

docs/_packages/web_api.md
```diff
@@ -246,7 +246,7 @@ pattern for people that use _functional programming_.
...
+[Modals](https://api.slack.com/block-kit/surfaces/modals) can be created by calling the `views.open` method. The method requires you to pass a valid [view payload](https://api.slack.com/reference/block-kit/views). Additionally, for apps not built with the CLI, the  `trigger_id`, which can be obtained in the payload returned when a user invokes your app using a slash command, clicking a button, or using [another interactive action](https://api.slack.com/reference/messaging/interactive-components). For apps built with the CLI, an `interactivity_pointer` is required instead of a `trigger_id`, which can be obtained in the `interactivity` object within the payload returned when a user invokes your app using a slash command, clicking a button, or using [another interactive action](https://api.slack.com/reference/messaging/interactive-components).
```
srajiang, line 249: How about something along the lines of the below? 

### Opening modals
[Modals](https://api.slack.com/block-kit/surfaces/modals) can be created by calling the `views.open` method. The method requires you to pass a valid [view payload](https://api.slack.com/reference/block-kit/views).

#### Handling modals during Slack Function handling (beta)
 If you are trying to open or update a modal from a Slack Function interactivity handler {link to Interactivity page here}, pass the `interactivity_pointer` you received from your event payload in your `views.open` method. 
 
```js
// Code snippet here
```

srajiang: I also feel like we ought to have some version of this in the Bolt documentation as well!

srajiang (changes requested):
Yay thanks for tackling this! Left some comments for ya.

packages/web-api/src/methods.ts
hello-ashleyintech: @srajiang took a look at this approach and it seems like the XOR will work but only by declaring a new type alias to populate the union into, meaning that the output type of `ViewsOpenArguments` and `ViewsPushArguments` would be `type` instead of `interface`. will this pose any issues? it looks like the primary differences with types and interfaces are that types can't be changed after being created.

srajiang: @hello-ashleyintech Yeah, so I guess if we took this route, the dev wouldn't be able to say, import that type and extend it using the `extends` TS syntax. Or internally, we couldn't use the `extends` syntax to extend a new interface out from it. 

But they they could declare a new type from there and use that imported type in an intersection or something if they wanted to add more attributes..

I haven't yet come across a scenario that required only `interface` vs. `type` so afaik that change could be fine.

packages/web-api/src/methods.ts
```diff
@@ -2103,16 +2103,34 @@ export interface UsersProfileSetArguments extends WebAPICallOptions, TokenOverri
...
   view: View;
 }
 
+interface ViewsPushInteractivityRequired extends WebAPICallOptions, TokenOverridable {
+  interactivity_pointer: string;
+  trigger_id?: never;
+  view: View;
+}
+
+export type ViewsPushArguments = ViewsPushTriggerRequired | ViewsPushInteractivityRequired;
```
srajiang, line 2132: Union type working very nicely 🚀 

Can we add some comments here to leave context on why either trigger_id or interactivity_pointer is required to help maintainers in the future?

srajiang (approved):
Okay! Some documentation comment suggestions here to wrap up, but the changes to the types are looking good to me.

Let's branch `next-gen` off main and point this branch to merge to that. That way we'll be all set up for cutting the beta release of `web-api`.

Issue #1528, opened 2022-09-01
fallback always renders even if attachments are included


edwmurph:
#### Packages:

Select all that apply:

- [x] `@slack/web-api`

#### The Slack SDK version

`6.7.2`

#### Node.js runtime version

`16.16.0`

#### OS info

```
ProductName:	macOS
ProductVersion:	12.5.1
BuildVersion:	21G83
Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:35 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T8101
```

#### Steps to reproduce:

<details>
<summary>run this snippet to repro:</summary>

```
const { WebClient }               = require('@slack/web-api');
const { decrypt }                 = require('enigma');
const { channel, botTokenCipher } = require('amp/lib/config').slack.dev;

const blocks = [
  {
    'type': 'header',
    'text': {
      'type': 'plain_text',
      'text': 'header 1'
    }
  },
  {
    'type': 'context',
    'elements': [
      {
        'type': 'mrkdwn',
        'text': '*ctx1:* val1'
      },
      {
        'type': 'mrkdwn',
        'text': '*ctx2:* val2'
      }
    ]
  }
];

const main = async() => {
  const token = await decrypt( botTokenCipher );
  const slack = new WebClient( token );

  await slack.chat.postMessage({
    channel,
    text: 'fallback1',
    attachments: [
      {
        fallback: 'fallback2',
        color: '#ff0000',
        blocks
      }
    ]
  });
};

main()
.then( console.log, console.log )
.finally( () => process.exit( 0 ) );
```
</details>

it renders the fallback even though the attachments are included

<img width="316" alt="image" src="https://user-images.githubusercontent.com/20996513/188017230-0e7be584-16d3-49bb-8b07-d828142a25ed.png">

<details>
<summary>and then when i remove the top level `text` opt</summary>

```
  await slack.chat.postMessage({
    channel,
    attachments: [
      {
        fallback: 'fallback2',
        color: '#ff0000',
        blocks
      }
    ]
  });
```
</details>

<details>
<summary>it no longer renders the fallback message, but now i get a warning in my console:</summary>

```
[WARN]  web-api:WebClient:0 The top-level `text` argument is missing in the request payload for a chat.postMessage call - It's a best practice to always provide a `text` argument when posting a message. The `text` is used in places where the content cannot be rendered such as: system push notifications, assistive technology such as screen readers, etc.
```
[...]

PR #1529, opened 2022-09-02
fix(1528): better warnings about webclient fallbacks

edwmurph:
###  Summary

fixes https://github.com/slackapi/node-slack-sdk/issues/1528

this PR updates the warning logic for the conversation api to better match this logic:

> If the top-level text does not exist and any of its attachments does not have fallback property, we still should warn it (we can say "You should have either top-level text or fallback in all the attachments" if we want to make it more accurate). If all the attachments have the fallback property, no warnings should be printed.

### Requirements

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -229,6 +229,32 @@ describe('WebClient', function () {
       ];
       const warningTestPatterns = textWarningTestPatterns.concat(attachmentWarningTestPatterns);
 
+      attachmentWarningTestPatterns.reduce((acc, { method, args }) => {
+        const attachmentPatterns = [{ fallback: "fallback" }].map((attachmentOverrides) => {
+          const attachments = args.attachments.map( attachment => {
+            return Object.assign({}, attachment, attachmentOverrides);
+          })
+          return { method, args: Object.assign({}, args, { attachments }) };
+        });
+
+        return acc.concat(attachmentPatterns)
+      }, []).forEach(({ method, args }) => {
+        it(`should not send warning to logs when client executes ${method} without text but with attachment fallback argument`, function () {
+          const logger = {
+            debug: sinon.spy(),
+            info: sinon.spy(),
+            warn: sinon.spy(),
+            error: sinon.spy(),
+            setLevel: sinon.spy(),
+            setName: sinon.spy(),
+          };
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.spec.js
```diff
@@ -278,7 +278,7 @@ describe('WebClient', function () {
           .map((v) => ({ method, args: Object.assign({}, v, args) }))
         return acc.concat(textPatterns)
       }, []).forEach(({ method, args }) => {
-        it(`should send both text and fallback-argument-specific warning to logs when client executes ${method} without text argument(${args.text === "" ? "empty" : args.text}) nor without attachment-level fallback argument (${args.attachments[0].fallback === "  " ? "empty" : args.attachments[0].fallback})`, function () {
```
seratch, line 281: All the existing tests should behave the same. My intention here is that we should change the behavior only for the pattern with **no text + valid fallbacks for all attachments**, which is your situation. Can you add a new test pattern to cover your situation?

packages/web-api/src/WebClient.ts
```diff
@@ -714,9 +714,10 @@ function warnIfFallbackIsMissing(method: string, logger: Logger, options?: WebAP
     'Alternatively, you can provide an attachment-level `fallback` argument, though this is now considered a legacy field (see https://api.slack.com/reference/messaging/attachments#legacy_fields for more details).';
   if (isTargetMethod && typeof options === 'object') {
     if (isEmptyText(options)) {
```
seratch, line 716: My intention of the new requirement is something like this:
```javascript
if (missingAttachmentFallbackDetected(options)) {
  if (isEmptyText(options)) {
    logger.warn(buildMissingTextWarning());
  }
  logger.warn(buildMissingFallbackWarning());
}
```

seratch (changes requested):
Thanks for taking the time to send this PR! Can you check my comments?

packages/web-api/src/WebClient.ts
edwmurph: i couldn't use that snippet ^ exactly because i think it broke existing functionality but i just pushed a commit that i think matches your original intention. could you give this another look when you get a chance?

seratch (approved):
Sorry for my belated response here. Looks great to me 👍

PR #1532, opened 2022-09-07
Bump lerna dependency

srajiang:
###  Summary

~~Lerna is officially [not actively maintained](https://github.com/lerna/lerna/pull/3092), and as far as I can tell it isn't used at the individual package level - each package maintains it's own tests.~~

Lerna ownership was handed over to a corp sponsor! I've bumped the version. 

This change ~~removes~~ bumps this dependency. 

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -25,6 +25,7 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
+    # - TODO run npm install and test on individual packages
     # - run: npm install
     # - run: npm run setup
     # - run: npm run test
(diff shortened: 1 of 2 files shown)


misscoded (approved):
LGTM :shipit: The only note historically that prevented this from occurring sooner (that I could find) was [this](https://github.com/slackapi/node-slack-sdk/pull/1265#discussion_r649590219). 

So, should be good to go if that's a non-issue!

srajiang:
@misscoded! Good call out - on further investigation, looks like lerna active maintenance got handed over to an official corporate sponsor, and they do have a recent version out, whoops. I've changed up the PR to bump the dependency instead. If that sounds good to you, I will merge.

misscoded:
Sounds fine to me but why keep it around if we're not using it? Did it in fact cause issues with the CI/CD stuff?

srajiang:
Spot on - so based o[n the CI/CD summary](https://github.com/slackapi/node-slack-sdk/runs/8238899663?check_suite_focus=true), we are using it to build all the packages and run some tests (I am not sure if it runs all tests of each package though) for CI/CD only. Keeping that around seems like a good thing.

PR #1535, opened 2022-09-15
Update Deno Continuous Deployment check

hello-ashleyintech:
###  Summary

Update the Deno Continuous Deployment check to include dependencies for built-in Node modules, `zlib` and `util`, and also update the Deno check to omit beta releases (containing `nextGen`, `beta`, and `Beta`)

Tested by running these lines (https://github.com/slackapi/node-slack-sdk/blob/main/.github/workflows/deno-build.yml#L29-L34) manually with the changes and seeing if I no longer ran into an error [like this](https://github.com/slackapi/node-slack-sdk/actions/runs/3025115544/jobs/4867235176) with the `zlib` and `util` imports.

### DIFF
--- a/.github/workflows/deno-build.yml
+++ b/.github/workflows/deno-build.yml
@@ -2,8 +2,13 @@ name: deno.land Continuous Deployment
 
 on:
   push:
+    branches:
+      - main
     tags:
       - '@slack/web-api*'
+      - '!@slack/web-api*nextGen*'
+      - '!@slack/web-api*Beta*'
+      - '!@slack/web-api*beta*'
 
 defaults:
   run:
@@ -30,6 +35,7 @@ jobs:
       run: |
         perl -pi.bak -e 's/"tsd": {/"browser":{"os":"os-browserify","path":"path-browserify","querystring":".\/deno-shims\/qs-shim.js"},"tsd": {/g' package.json
         perl -pi.bak -e 's/"axios": /"path-browserify":"1.0.1","os-browserify":"0.3.0","axios": /g' package.json
+        perl -pi.bak -e 's/"p-retry":/"zlib":".\/deno-shims\/zlib-shim.js","util":".\/deno-shims\/util-shim.js","p-retry": /g' package.json
     - run: npm install
     - run: npm run build:deno
     - run: mv mod.js ../../.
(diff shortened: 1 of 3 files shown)


.github/workflows/deno-build.yml
```diff
@@ -2,8 +2,13 @@ name: deno.land Continuous Deployment
 
 on:
   push:
+    branches:
+      - main
     tags:
       - '@slack/web-api*'
+      - '!@slack/web-api*nextGen*'
```
hello-ashleyintech, line 9: Is this the best syntax? I didn't see anything specific in the docs about using asterisks within a string to detect something that contains a certain word but may have words or characters before or after it, but that's what i'm trying to do here with `nextGen`, `Beta`, and `beta`.

.github/workflows/deno-build.yml
```diff
@@ -2,8 +2,13 @@ name: deno.land Continuous Deployment
 
 on:
   push:
+    branches:
+      - main
```
hello-ashleyintech, line 6: also..is there a scenario where we'd ever be pushing a release from a branch other than main?

filmaj: I don't think so, but I don't think we want this `branches` section, as this means this workflow will run on any push to `main`. I believe that each entry under `push:` is logically OR'ed together - so with this current workflow file, if _any_ of the following happen, then the deno build would be kicked off:

- a push to main, regardless of tag
- a tag that begins with `@slack/web-api`
  - ... but if a tag matches any of the negation specs (the ones that start with `!`), then the previous match tag would be negated (at least based on the [filtering logic cheatsheet docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet), if I understand them correctly).

It's worth mentioning that a tag is actually its own entity independent of any branch - it's like a reference to a git commit SHA.

.github/workflows/deno-build.yml
filmaj: Yeah I think that's fine.

It is maybe worth testing the matching syntax out on your own fork (and e.g. just commenting out the run steps or changing them to do something harmless like `echo` some text), just to make sure that pushing different tags and to different branches triggers the workflow as you expect.

filmaj (changes requested):
I think this is fine, though I'm not sure we want the `branches:` specifier.

hello-ashleyintech:
Tested this with my fork (only up to the `npm run build:deno` step and pushing up different tags , behavior is working as expected! Marking this as "Ready for Review".

filmaj (approved):

PR #1542, opened 2022-10-04
Reviewing Types

filmaj:
⚠️ DO NOT MERGE

###  Summary

This is my first pass at reviewing the types in this repo in preparation for wanting to consume them in the deno-slack-sdk.

I used the following api.slack.come pages as part of this review:

- https://api.slack.com/reference/block-kit/blocks
- https://api.slack.com/reference/block-kit/block-elements
- https://api.slack.com/reference/block-kit/composition-objects

I tried to organize a majority of the types in this module into roughly three categories based on the above api.slack.com pages.

Added some missing properties, updated what is optional and what is not based on the above api.slack.com pages. I left some TODOs in, as well, based on what issues I saw that I did not feel like tackling immediately.

### TODO

I do want to split this file up into smaller, more focussed/thematic files - but I will do that in a follow-up PR. I want to also put in some time to add nice JSDoc comments to all interfaces and types, I think that would improve the developer experience as well. Initially, however, I just want the properties to be reflected accurately so that I can test this repo in bolt-js, web-api and in deno-slack-sdk apps, and see how they work together.

### Testing

Speaking of testing: I tested this with both the `@slack/web-api` and `@slack/bolt` repos. Both repos' unit tests pass without issue. That said, I am wondering if this is a false negative... likely needs more testing in a real app.

### Releasing

Not sure if this should be released / merged into `main`, or if this should maybe be released as a RC/alpha/beta style release on npm in the short term, until more complete manual testing can be complete. Thoughts?

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -63,15 +63,9 @@ export interface WorkflowStepView {
 
 export type View = HomeView | ModalView | WorkflowStepView;
 
-/*
- * Block Elements
+/**
+ * Composition Objects: https://api.slack.com/reference/block-kit/composition-objects
  */
-
-export interface ImageElement {
-  type: 'image';
-  image_url: string;
-  alt_text: string;
-}
 
 export interface PlainTextElement {
   type: 'plain_text';
@@ -87,31 +81,51 @@ export interface MrkdwnElement {
 
 export interface MrkdwnOption {
   text: MrkdwnElement;
-  value?: string;
+  value: string;
   url?: string;
   description?: PlainTextElement;
 }
 
 export interface PlainTextOption {
   text: PlainTextElement;
-  value?: string;
+  value: string;
   url?: string;
   description?: PlainTextElement;
 }
 
+export interface Confirm {
+  title: PlainTextElement;
+  text: PlainTextElement | MrkdwnElement;
+  confirm: PlainTextElement;
+  deny: PlainTextElement;
+  style?: 'primary' | 'danger'; // TODO: factor out into own type
+}
+
 export type Option = MrkdwnOption | PlainTextOption;
 
-export interface Confirm {
-  title?: PlainTextElement;
-  text: PlainTextElement | MrkdwnElement;
-  confirm?: PlainTextElement;
-  deny?: PlainTextElement;
-  style?: 'primary' | 'danger';
+export interface DispatchActionConfig {
+  trigger_actions_on?: ('on_enter_pressed' | 'on_character_entered')[];
 }
 
 /*
- * Action Types
+ * Block Elements: https://api.slack.com/reference/block-kit/block-elements
+ * Also known as interactive elements
+ * Exported below as Action
  */
+
+export type KnownAction =
+  Select | MultiSelect | Button | Overflow | Datepicker | Timepicker | RadioButtons | Checkboxes | PlainTextInput;
+
+export interface Action {
+  type: string;
+  /**
+   * @description: A string acting as a unique identifier for a block.
+   */
+  block_id?: string; // TODO: in event payloads, this property will always be present. When defining a block, however,
+  // it is optional. If not defined when the block is created, Slack will auto-assign a block_id for you - thus why it
+  // is always present in payloads. So: how can we capture that? Differentiate between the event payload vs. the block
+  // element?
+}
 
 // Selects and Multiselects are available in different surface areas so I've separated them here
 export type Select = UsersSelect | StaticSelect | ConversationsSelect | ChannelsSelect | ExternalSelect;
@@ -119,23 +133,20 @@ export type Select = UsersSelect | StaticSelect | ConversationsSelect | Channels
 export type MultiSelect =
   MultiUsersSelect | MultiStaticSelect | MultiConversationsSelect | MultiChannelsSelect | MultiExternalSelect;
 
-export interface Action {
-  type: string;
-  action_id?: string;
-}
-
 export interface UsersSelect extends Action {
   type: 'users_select';
+  action_id: string;
   initial_user?: string;
-  placeholder?: PlainTextElement;
+  placeholder: PlainTextElement;
   confirm?: Confirm;
   focus_on_load?: boolean;
 }
 
 export interface MultiUsersSelect extends Action {
   type: 'multi_users_select';
+  action_id: string;
   initial_users?: string[];
-  placeholder?: PlainTextElement;
+  placeholder: PlainTextElement;
   max_selected_items?: number;
   confirm?: Confirm;
   focus_on_load?: boolean;
@@ -143,7 +154,8 @@ export interface MultiUsersSelect extends Action {
 
 export interface StaticSelect extends Action {
   type: 'static_select';
-  placeholder?: PlainTextElement;
+  placeholder: PlainTextElement;
+  action_id: string;
   initial_option?: PlainTextOption;
   options?: PlainTextOption[];
   option_groups?: {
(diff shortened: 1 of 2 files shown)


packages/types/src/index.ts
```diff
@@ -87,63 +81,81 @@ export interface MrkdwnElement {
...
-  type: string;
-  action_id?: string;
-}
-
 export interface UsersSelect extends Action {
   type: 'users_select';
+  action_id: string;
   initial_user?: string;
-  placeholder?: PlainTextElement;
+  placeholder: PlainTextElement;
```
seratch, line 140: I am sure that this is optional, so this change should be reverted. [The document](https://api.slack.com/reference/block-kit/block-elements#users_select) is wrong! See [a Block Kit Builder example](https://app.slack.com/block-kit-builder/#%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22Test%20block%20with%20users%20select%22%7D,%22accessory%22:%7B%22type%22:%22users_select%22,%22action_id%22:%22users_select-action%22%7D%7D%5D%7D)

packages/types/src/index.ts
```diff
@@ -87,63 +81,81 @@ export interface MrkdwnElement {
...
   confirm?: Confirm;
   focus_on_load?: boolean;
 }
 
 export interface MultiUsersSelect extends Action {
   type: 'multi_users_select';
+  action_id: string;
   initial_users?: string[];
-  placeholder?: PlainTextElement;
+  placeholder: PlainTextElement;
```
seratch, line 149: same as above

packages/types/src/index.ts
```diff
@@ -87,63 +81,81 @@ export interface MrkdwnElement {
...
   max_selected_items?: number;
   confirm?: Confirm;
   focus_on_load?: boolean;
 }
 
 export interface StaticSelect extends Action {
   type: 'static_select';
-  placeholder?: PlainTextElement;
+  placeholder: PlainTextElement;
+  action_id: string;
```
seratch, line 158: nit: when you build a block element, setting action_id is optional (auto-generated one will be used in the case). thus, strictly speaking, this can be optional

packages/types/src/index.ts
```diff
@@ -156,9 +168,10 @@ export interface StaticSelect extends Action {
 
 export interface MultiStaticSelect extends Action {
   type: 'multi_static_select';
-  placeholder?: PlainTextElement;
+  placeholder: PlainTextElement;
+  action_id: string;
   initial_options?: PlainTextOption[];
-  options?: PlainTextOption[];
+  options?: PlainTextOption[]; // TODO: options and option_groups are mutually exclusive
```
seratch, line 174: good point but i am unsure how we can represent this type of constraints in a simple way

seratch (changes requested):
Thanks for improving this!

seratch:
@filmaj For better testing, checking the compatibility with actual production payload data may be helpful. I have been doing so for python/java projects, which is quite beneficial for detecting errors in real life.

packages/types/src/index.ts
filmaj: Are you sure `action_id` is optional? The docs say it is required. I know `block_id` is optional and one will be auto-generated for you in that case.
I will test this today in real apps and report back!

filmaj:
@seratch sounds good, I was planning on doing that yesterday but ran out of time. I will do so today.

packages/types/src/index.ts
filmaj: Wow I can't believe it, action_id _is_ optional 😆

packages/types/src/index.ts
filmaj: You are correct! I was very surprised but confirmed that for all these [Block Elements](https://api.slack.com/reference/block-kit/block-elements), both placeholder and action_id are optional. I am working on a PR to update the docs. Thanks for pointing this out, I have updated these to set placeholder and action_id to optional.

packages/types/src/index.ts
```diff
@@ -1,6 +1,7 @@
 /*
  * Reusable shapes for argument values
  */
+// TODO: dialogs are deprecated https://api.slack.com/dialogs
```
srajiang, line 4:
```suggestion
// TODO: dialogs are deprecated https://api.slack.com/dialogs
// Guide on upgrading dialogs to modals: https://api.slack.com/block-kit/dialogs-to-modals
```

packages/types/src/index.ts
```diff
@@ -226,6 +241,7 @@ export interface ExternalSelect extends Action {
 
 export interface MultiExternalSelect extends Action {
   type: 'multi_external_select';
+  // event to return option data. so de-facto this is necessary
```
srajiang, line 245: I'm not certain I understand what this comment means! Is this a note to self?

filmaj: LOL it was a much longer comment that I had before that I then decided to remove _which I clearly didn't do a great job of_ 😆

packages/types/src/index.ts
```diff
@@ -87,43 +82,60 @@ export interface MrkdwnElement {
...
+  Select | MultiSelect | Button | Overflow | Datepicker | Timepicker | RadioButtons | Checkboxes | PlainTextInput;
 
 export interface Action {
   type: string;
+  /**
+   * @description: A string acting as a unique identifier for a block.
+   */
+  block_id?: string; // TODO: in event payloads, this property will always be present. When defining a block, however,
+  // it is optional. If not defined when the block is created, Slack will auto-assign a block_id for you - thus why it
+  // is always present in payloads. So: how can we capture that? Differentiate between the event payload vs. the block
```
srajiang, line 129: We probably have to, since I don't see a world where this particular behavior changes - and the behavior seems reasonable. Were you thinking that this Block Kit type could somehow be used to model event payload responses?

filmaj: Yes exactly! At least, I was consider if that would be possible! If you look at the [`block_actions` payload event's fields](https://api.slack.com/reference/interaction-payloads/block-actions#fields), you'll see that there is an array of `actions` provided in the payload. These actions are the exact block kit elements that were interacted with - so I was considering if it would be possible to re-use the types in this repo to try to model the items of that `actions` array in the `block_actions` payload.

srajiang (commented):
Thanks for the review add and so glad you're working on this (mostly because of all the adjacent improvements to our documentation this work becomes a forcing function too...)! I don't have much to add wrt to correctness of these types but: 

wrt: Testing 
👍 to more manual testing needed, and wondering whether we can spin off a couple of concrete tasks around incorporating into sample apps.

wrt: release strategy
An rc release of this off a branch while we're testing sounds good to me! Any strong benefit you see to incorporating into `main` at this early stage?

filmaj:
> An rc release of this off a branch while we're testing sounds good to me! Any strong benefit you see to incorporating into main at this early stage?

RC release seems safe to me at this stage, yeah. I do plan on adding a ton of niceties to this repo (JSdocs, links to api.slack.com, etc.), so releasing to main at some point would be a nice little present for our devs 😊

packages/types/src/index.ts
seratch: As for the item in actions in a block_actions payload, we already have complete types in bolt-js. If we would like to have it, we can consider moving the ones in bolt-js into this repo. Reusing `Action` and so on here may cause unnecessary complexity.

seratch (approved):
Looks good to me so far

packages/types/src/index.ts
filmaj: @seratch Yes I would like to eventually move as many types for things shared across repos as possible into this repo. For the block_actions payload, for example, I know right now we only need it in bolt-js. However, in deno-slack-sdk, it is also needed. So all of this work is in prep for trying to consolidate not just around our existing node projects, but across our newer deno ones too.

filmaj:
So maybe instead of merging this PR into the `main` branch, I keep this work on the branch it currently exists on (`centralized-types`) and create beta / release-candidate tags off of this branch for the time being?

filmaj:
I've started to release this branch under the `2.8.0-centralized.X` suite of git tags and the `centralized` dist-tag on npm.

PR #1549, opened 2022-10-28
Add types for DateTimepicker, EmailInput, NumberInput and URLInput

filmaj:
###  Summary

Added new types for new Block Elements

### Requirements (place an `x` in each `[ ]`)

* [ ] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [ ] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).

PR #1550, opened 2022-10-28
Factor out `confirm`, `placeholder`, `focus_on_load` and `dispatch_action_config` properties

filmaj:
The idea behind this PR is to have a set of interfaces that other Block Kit elements can extend from easily.

This PR also starts adding some of the documentation hints from api.slack.com as JSDoc comments.

It is a small refactor, but it is especially helpful when documenting properties, as then they can be documented in one place but used in many places.

Let me know what you think! Like/dislike/don't care?

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -178,18 +178,18 @@ export interface Dispatchable {
   dispatch_action_config?: DispatchActionConfig;
 }
 
-export interface UsersSelect extends Action, Confirmable, Focusable, Placeholderable {
+export interface UsersSelect extends Action, Confirmable, Focusable, Placeholdable {
   type: 'users_select';
   initial_user?: string;
 }
 
-export interface MultiUsersSelect extends Action, Confirmable, Focusable, Placeholderable {
+export interface MultiUsersSelect extends Action, Confirmable, Focusable, Placeholdable {
   type: 'multi_users_select';
   initial_users?: string[];
   max_selected_items?: number;
 }
 
-export interface StaticSelect extends Action, Confirmable, Focusable, Placeholderable {
+export interface StaticSelect extends Action, Confirmable, Focusable, Placeholdable {
   type: 'static_select';
   initial_option?: PlainTextOption;
   options?: PlainTextOption[];
@@ -199,7 +199,7 @@ export interface StaticSelect extends Action, Confirmable, Focusable, Placeholde
   }[];
 }
 
-export interface MultiStaticSelect extends Action, Confirmable, Focusable, Placeholderable {
+export interface MultiStaticSelect extends Action, Confirmable, Focusable, Placeholdable {
   type: 'multi_static_select';
   initial_options?: PlainTextOption[];
   options?: PlainTextOption[];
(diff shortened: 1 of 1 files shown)


packages/types/src/index.ts
```diff
@@ -121,181 +137,170 @@ export type MultiSelect =
...
-export interface MultiUsersSelect extends Action {
+export interface Placeholderable {
+  /**
+   * @description A {@see PlainTextElement} object that defines the placeholder text shown on the element. Maximum
+   * length for the `text` field in this object is 150 characters.
+   */
+  placeholder?: PlainTextElement;
+}
+
+export interface Dispatchable {
```
filmaj, line 173: Right now only one element uses this interface, but if/when we merge https://github.com/slackapi/node-slack-sdk/pull/1549, then the new input elements like URL, email and number inputs will also use this interface.

packages/types/src/index.ts
```diff
@@ -121,181 +137,170 @@ export type MultiSelect =
...
+  /**
+   * @description Indicates whether the element will be set to auto focus within the
+   * {@link https://api.slack.com/reference/surfaces/views `view` object}. Only one element can be set to `true`.
+   * Defaults to false.
+   */
   focus_on_load?: boolean;
 }
 
-export interface MultiUsersSelect extends Action {
+export interface Placeholderable {
```
srajiang, line 165:
```suggestion
export interface Placeholdable {
```
Too many syllables for my little brain

packages/types/src/index.ts
```diff
@@ -121,181 +137,170 @@ export type MultiSelect =
...
+
+export interface Dispatchable {
+  /**
+   * @description A {@see DispatchActionConfig} object that determines when during text input the element returns a
+   * {@link https://api.slack.com/reference/interaction-payloads/block-actions `block_actions` payload}.
+   */
+  dispatch_action_config?: DispatchActionConfig;
+}
+
+export interface UsersSelect extends Action, Confirmable, Focusable, Placeholderable {
```
srajiang, line 181: :chefskiss:

WilliamBergamin (approved):
Great work 💯 this reminds me of a decorator pattern

srajiang (changes requested):
I like this direction, and thank you for additional jsdoc to help document things. Left a minor comment about naming.

packages/types/src/index.ts
```diff
@@ -121,181 +137,170 @@ export type MultiSelect =
...
+   * with.
+   */
   confirm?: Confirm;
+}
+
+export interface Focusable {
+  /**
+   * @description Indicates whether the element will be set to auto focus within the
+   * {@link https://api.slack.com/reference/surfaces/views `view` object}. Only one element can be set to `true`.
+   * Defaults to false.
```
mwbrooks, line 160:
```suggestion
   * Defaults to `false`.
```

mwbrooks (approved):
<3 Love it! Only found a minor improvement suggestion.

e-zim (approved):
These documentation hints are wonderful, particularly when `@see` comes into play! I'm a huge fan 👏

PR #1551, opened 2022-10-28
ChatUnfurlArguments require either `channel` and `ts` or `unfurl_id` and `source` properties to be specified in pairs

### LINKED ISSUE #1530: Incomplete types for ChatUnfurlArguments
According to documentation [`chat.unfurl`](https://api.slack.com/methods/chat.unfurl) accepts either `unfurl_id` and `source` or `channel` and `ts`. But `ChatUnfurlArguments` type misses `unfurl_id` and `source` params ([source](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/methods.ts#L1457-L1465)).

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] `@slack/types`
- [ ] I don't know

filmaj:
This fixes #1530.

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -1481,8 +1481,10 @@ export interface ChatScheduledMessagesListArguments extends WebAPICallOptions, T
 }
 cursorPaginationEnabledMethods.add('chat.scheduledMessages.list');
 export interface ChatUnfurlArguments extends WebAPICallOptions, TokenOverridable {
-  channel: string;
-  ts: string;
+  channel?: string; // if specified, `ts` must be specified too. otherwise, `unfurl_id` and `source` must be specified.
+  ts?: string;
+  unfurl_id?: string; // if specified, `source` must be specified too. otherwise, `channel` and `ts` must be specified.
+  source?: 'composer' | 'conversations_history';
   unfurls: LinkUnfurls;
   user_auth_message?: string;
   user_auth_required?: boolean;
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -1480,17 +1480,60 @@ export interface ChatScheduledMessagesListArguments extends WebAPICallOptions, T
   team_id?: string; // required if org token is used
 }
 cursorPaginationEnabledMethods.add('chat.scheduledMessages.list');
-export interface ChatUnfurlArguments extends WebAPICallOptions, TokenOverridable {
-  channel?: string; // if specified, `ts` must be specified too. otherwise, `unfurl_id` and `source` must be specified.
-  ts?: string;
-  unfurl_id?: string; // if specified, `source` must be specified too. otherwise, `channel` and `ts` must be specified.
-  source?: 'composer' | 'conversations_history';
-  unfurls: LinkUnfurls;
-  user_auth_message?: string;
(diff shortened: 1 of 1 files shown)


stevengill (approved):
lgtm

stevengill (approved):
lgtm

srajiang (approved):
LGTM!

packages/web-api/src/methods.ts
```diff
@@ -1481,8 +1481,10 @@ export interface ChatScheduledMessagesListArguments extends WebAPICallOptions, T
 }
 cursorPaginationEnabledMethods.add('chat.scheduledMessages.list');
 export interface ChatUnfurlArguments extends WebAPICallOptions, TokenOverridable {
```
filmaj, line 1483: @stevengill @srajiang and rest of the team: what about using a union type in order to represent the situation in this API of "either provide `source` and `unfurl_id`, or `channel` and `ts`"? It's a bit more complicated but it should make it clear to developers that one needs to provide either pair of arguments, but can't mix and match between the four properties (and also disambiguates a bit since we remove the superfluous optionality of all four props).

I.e. something like:

```
interface ChannelAndTS {
  channel: string;
  ts: string;
}

interface SourceAndUnfurlID {
  source: 'composer' | 'conversations_history';
  unfurl_id: string;
}

type UnfurlArguments = (ChannelAndTS | SourceAndUnfurlID) & {
  unfurls: string[];
  whatever?: string;
};

const _channelArgs: UnfurlArguments = {
  channel: 'C12345',
  ts: '1234.56',
  unfurls: ['yes', 'no'],
};

const _sourceArgs: UnfurlArguments = {
  source: 'composer',
  unfurl_id: '12345',
  unfurls: ['maybe', 'so'],
};
```

srajiang: Yeah, you can definitely do that - we did something like that for next-gen views.open method with `interactivity_pointer` vs. 'trigger_id'.

seratch: This is definitely better for TS users

filmaj:
Did another pass at this, this time trying to encode the either-or argument passing for these methods. Lemme know what y'all think!

packages/web-api/src/methods.ts
```diff
@@ -1480,15 +1480,60 @@ export interface ChatScheduledMessagesListArguments extends WebAPICallOptions, T
...
+}
+
+interface SourceAndUnfurlIDArguments {
+  /**
+   * @description The source of the link to unfurl. The source may either be `composer`, when the link is inside the
+   * message composer, or `conversations_history`, when the link has been posted to a conversation.
+   */
+  source: 'composer' | 'conversations_history';
+  /**
+   * @description The ID of the link to unfurl. Both `unfurl_id` and `source` must be provided together, or `channel`
```
srajiang, line 1503:
```suggestion
   * @description The ID of the link to unfurl.
```
Think this part of the jsdoc is still necessary?

srajiang (approved):
This looks fine to me! :)

seratch (approved):

PR #1650, opened 2023-08-24
Logger updates: dependencies and higher minimum node version

filmaj:
This is mainly intended as a discussion point.

I would like to do some updates in this repository. Recently the build for this repo was broken and this prompted me to look into it more. There are many issues with this repo, but I'd like to slowly chip away at improving things.

This PR is a template for what I would like to roll out to all the other non-deprecated packages in this repo. In short:

1. Updating all the dependencies.
2. Addressing any linter warnings.
3. Bumping the minimum node version to v18 (v12 and 14 are already EOL, and 16 will be EOL in ~2 weeks).

I wanted to see what the team thinks of these changes? What are your opinions? Do you agree or disagree with any of these changes, in particular the raising of the minimum node version? And do you think this should warrant a major semver version increase?

Longer term, after these initial updates I would also like to tackle:

- Updating the docs revisioned in this repository and possibly generating the reference documentation on every commit in CI, instead of manually.
- Factoring out developer dependencies / tools, such as eslint, that are shared or common across all sub-packages, into the top-level `package.json`. Possibly this could leverage npm workspaces, though my initial foray into this area was unsuccessful, but I think mostly due to trying to support too many incompatible node versions at the same time.

### DIFF
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -1,0 +1,33 @@
+# Slack Logger
+
+The `@slack/logger` package is intended to be used as a simple logging interface that supports verbosity levels.
+
+## Requirements
+
+This package supports Node v18 and higher. It's highly recommended to use [the latest LTS version of
+node](https://github.com/nodejs/Release#release-schedule), and the documentation is written using syntax and features
+from that version.
+
+## Installation
+
+```shell
+$ npm install @slack/logger
+```
+
+## Usage
+
+This package exports a `ConsoleLogger` class, a generic `Logger` interface and a `LogLevel` enum.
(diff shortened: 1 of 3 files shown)


seratch:
When we drop older versions than 18, README (and some other document sections too?) needs to be updated accordingly: https://github.com/slackapi/node-slack-sdk#requirements

seratch (approved):
I don't see any issues with these changes. Left comment on the Node version requirement on other docs and README

filmaj:
I've added a README to this sub-package and made the node v18 requirement explicit. I will merge and release this and send similar pull requests for the other packages. In the next PR, I will update the root README.

PR #1655, opened 2023-08-30
types: Adding deprecation notices to Dialogs and Steps From Apps types

filmaj:
Also added information about Message Attachments: while not deprecated per se, there be dragons.

This is in preparation for larger changes I am planning in this package that will warrant a major version upgrade. Might as well deprecate the types we no longer officially support so that we can remove them in a future major version.


srajiang (approved):
LGTM!

PR #1656, opened 2023-08-31
Types: split single index.ts file into files based on function/category

filmaj:
###  Summary

This is purely a text-movement PR; no changes (other than a few comments at the top of certain files) from the base.

I'm mainly looking for feedback on the organization of the files. If you have feedback on directory structure or file names, let me know!


packages/types/src/block-kit/blocks.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export interface Block {
+  type: string;
+  block_id?: string;
+}
+
+export type KnownBlock = ImageBlock | ContextBlock | ActionsBlock | DividerBlock |
+SectionBlock | InputBlock | FileBlock | HeaderBlock | VideoBlock;
+
+export interface ActionsBlock extends Block {
```
srajiang, line 19: Seems worth it to adding additional annotation to more easily find the documentation link while we're at it? 

Fine to punt to the later TODO moment referenced in the comments above as well. 

```suggestion
// Reference: https://api.slack.com/reference/block-kit/blocks#actions
export interface ActionsBlock extends Block {
```

packages/types/src/block-kit/elements.ts
```diff
@@ -0,0 +1,224 @@
+// This file contains objects documented here: https://api.slack.com/reference/block-kit/block-elements
```
srajiang, line 1: Should this just be called block-elements.ts as well based on the documentation? Would line up better alphabetically with blocks.ts and makes it clear they belong to blocks

packages/types/src/dialog.ts
```diff
@@ -0,0 +1,38 @@
+/**
```
srajiang, line 1: Nice that we'll just be able to poof this file when major version bump comes along 🧹 !

packages/types/src/views.ts
```diff
@@ -0,0 +1,38 @@
+import { Block, KnownBlock } from './block-kit/blocks';
+import { PlainTextElement } from './block-kit/composition-objects';
+
+export interface HomeView {
```
srajiang, line 4:
```suggestion
// Reference: https://api.slack.com/surfaces/app-home#composing
export interface HomeView {
```

packages/types/src/views.ts
```diff
@@ -0,0 +1,38 @@
...
+  type: 'home';
+  blocks: (KnownBlock | Block)[];
+  private_metadata?: string;
+  callback_id?: string;
+  external_id?: string;
+}
+
+export interface ModalView {
```
srajiang, line 12:
```suggestion
// Reference: https://api.slack.com/surfaces/modals#composing_views
export interface ModalView {
```

srajiang (commented):
I like the structure! Left a few small suggestions most notably renaming elements.ts as block-elements.ts.

packages/types/src/block-kit/blocks.ts
filmaj: Yes exactly, I will be doing exactly these kind of changes in the next pull request 😄 I didn't want to add more additional lines in this specific PR as the diff would be too confusing given the movement of code across files. I think that in the next PR, though, these kinds of documentation additions you suggest would be clearer and easier to review.

packages/types/src/block-kit/elements.ts
filmaj: Yes excellent point! Will do

srajiang (approved):
Great, LGTM!

Issue #1037, opened 2020-06-11
Default Content-Type rule is being deleted from Axios config globally


notable-marcelo:
### Description

[This piece of code](https://github.com/slackapi/node-slack-sdk/blob/766851cb0db3822cda471dfbb8aaf59581846267/packages/web-api/src/WebClient.ts#L129) seems to be removing the Axios default Content-Type config for POST requests at a global level.

So, a system using `node-slack-sdk` could get very odd side effect if it also relies on Axios.

Note how the `post` default config is removed globally.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Example, an axios default config on a random application would look something like:
```
> axios.defaults
{
  adapter: [Function: httpAdapter],
  transformRequest: [ [Function: transformRequest] ],
  transformResponse: [ [Function: transformResponse] ],
  timeout: 0,
  xsrfCookieName: 'XSRF-TOKEN',
  xsrfHeaderName: 'X-XSRF-TOKEN',
  maxContentLength: -1,
  validateStatus: [Function: validateStatus],
  headers: {
    common: { Accept: 'application/json, text/plain, */*' },
    delete: {},
    get: {},
    head: {},
    post: { 'Content-Type': 'application/x-www-form-urlencoded' },
    put: { 'Content-Type': 'application/x-www-form-urlencoded' },
    patch: { 'Content-Type': 'application/x-www-form-urlencoded' }
  }
}
```
However, after sending a slack message using `node-slack-sdk`, it would be:
```
> axios.defaults
{
  adapter: [Function: httpAdapter],
  transformRequest: [ [Function: transformRequest] ],
  transformResponse: [ [Function: transformResponse] ],
  timeout: 0,
  xsrfCookieName: 'XSRF-TOKEN',
  xsrfHeaderName: 'X-XSRF-TOKEN',
  maxContentLength: -1,
  validateStatus: [Function: validateStatus],
[...]

PR #1667, opened 2023-10-03
@slack/web-api: prep for next major release. bump min node to v18

filmaj:
# Summary

- cleaning up the readme.
- remove no longer used codecov and deno build npm run scripts.
- update dependencies as much as possible.
- fix linter errors.
- remove a test that is no longer applicable with new axios.

I'd like to work on slowly getting the next 7.0 major release out, and this would be a good opportunity to eliminate EOL'ed node support while we are at it. I will be reviewing the existing issues for this package to see if we can tackle any other breaking changes while we are at it.


packages/web-api/src/WebClient.spec.js
```diff
@@ -34,20 +34,6 @@ describe('WebClient', function () {
       assert.instanceOf(client, WebClient);
       assert.notExists(client.axios.defaults.headers.Authorization);
     });
-    it('should not modify global defaults in axios', function () {
```
filmaj, line 37: I didn't feel like this test offers much value and since this PR moves from axios 0.x to 1.x, also not sure how much we should be testing the output of _other_ packages.

packages/web-api/src/WebClient.spec.js
```diff
@@ -149,7 +135,6 @@ describe('WebClient', function () {
             assert.equal(error.code, ErrorCode.RequestError);
             assert.equal(error.original.config.timeout, timeoutOverride);
             assert.equal(error.original.isAxiosError, true);
-            assert.equal(error.original.request.aborted, true);
```
filmaj, line 152: In axios 1.0, this assertion no longer passes. Not sure how important that is. Let me know if you have opinions on that.

packages/web-api/src/WebClient.spec.js
seratch: @filmaj As you can see, the context behind this unit test is https://github.com/slackapi/node-slack-sdk/issues/1037 This package had an issue polluting the global singleton instance of `axios` package. This test tries to prevent the same issue from occurring again. If axois 1.x does not allow such destructive operations to the global singleton, we can safely remove this test. Thus, I don't think we're testing 3rd party library here. We verify our package works without any major side effects.

If investigation on how to properly maintain this test with axios 1.x could require our time and efforts, it's fine to delete the test anyways. Whenever we have something similar in the future, we can add tests again.

packages/web-api/src/WebClient.spec.js
seratch: I don't see any issues with the deletion

packages/web-api/src/errors.ts
```diff
@@ -94,7 +94,13 @@ export function httpErrorFromResponse(response: AxiosResponse): WebAPIHTTPError
   ) as Partial<WebAPIHTTPError>;
   error.statusCode = response.status;
   error.statusMessage = response.statusText;
-  error.headers = response.headers;
+  const nonNullHeaders: Record<string, string> = {};
```
seratch, line 97: good catch

filmaj: TypeScript made me do it 😄

packages/web-api/src/WebClient.spec.js
filmaj: It seems [in the latest axios, no specific value is set for the HTTP POST default headers](https://github.com/axios/axios/blob/a48a63ad823fc20e5a6a705f05f09842ca49f48c/lib/defaults/index.js#L155-L157). This is [also the case for 'common' HTTP all-verb headers](https://github.com/axios/axios/blob/a48a63ad823fc20e5a6a705f05f09842ca49f48c/lib/defaults/index.js#L150).

[In v0.27.2 of axios, this was not the case: POST would set a default `Content-Type`](https://github.com/axios/axios/blob/v0.27.2/lib/defaults/index.js#L142-L144). and [the default was `application/x-www-form-urlencoded`](https://github.com/axios/axios/blob/v0.27.2/lib/defaults/index.js#L9-L11).

Therefore, I think this test can be safely removed.

packages/web-api/README.md
```diff
@@ -394,6 +383,6 @@ If you get stuck, we're here to help. The following are the best ways to get ass
 
   * [Issue Tracker](http://github.com/slackapi/node-slack-sdk/issues) for questions, feature requests, bug reports and
     general discussion related to these packages. Try searching before you create a new issue.
-  * [Email us](mailto:developers@slack.com) in Slack developer support: `developers@slack.com`
-  * [Bot Developers Hangout](https://community.botkit.ai/): a Slack community for developers
-    building all types of bots. You can find the maintainers and users of these packages in **#sdk-node-slack-sdk**.
+  * [Email us](mailto:feedback@slack.com): `feedback@slack.com`
```
WilliamBergamin, line 386: Should they email feedback@slack.com or support@slack.com as indicted in [other repos](https://github.com/slackapi/bolt-js#getting-help)

filmaj: My understanding is `feedback@slack.com`

packages/web-api/src/WebClient.spec.js
seratch: Thanks for checking this!

seratch (approved):

PR #1668, opened 2023-10-04
Update files.upload.v2 internals due to server-side improvements

seratch:
###  Summary

This pull request updates the internals of files.upload.v2 method to eliminate `files.info` API calls, which are no longer necessary because the server-side now returns the same metadata as part of `files.completeUploadExternal` API responses.

See also:
* https://github.com/slackapi/python-slack-sdk/pull/1408
* https://github.com/slackapi/java-slack-sdk/releases/tag/v1.34.0

### DIFF
--- a/packages/web-api/src/methods.ts
+++ b/packages/web-api/src/methods.ts
@@ -1884,7 +1884,7 @@ interface FileUpload {
 
 export interface FilesUploadV2Arguments extends FileUploadV2, WebAPICallOptions, TokenOverridable {
   file_uploads?: Omit<FileUploadV2, 'channel_id' | 'channels' | 'initial_comment' | 'thread_ts'>[];
-  request_file_info?: boolean;
+  request_file_info?: boolean; // since v7, this flag is no longer used
 }
 
 export type FileUploadV2 = FileUpload & {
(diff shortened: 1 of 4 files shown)


packages/web-api/src/methods.ts
```diff
@@ -1884,7 +1884,7 @@ interface FileUpload {
 
 export interface FilesUploadV2Arguments extends FileUploadV2, WebAPICallOptions, TokenOverridable {
   file_uploads?: Omit<FileUploadV2, 'channel_id' | 'channels' | 'initial_comment' | 'thread_ts'>[];
-  request_file_info?: boolean;
+  request_file_info?: boolean; // since v7, this flag is no longer used
```
filmaj, line 1887: Maybe we can add a JSdoc description to this parameter with the `@deprecated` tag, to make clear not to use this?

filmaj (approved):
LGTM, I left one suggestion on deprecating the v2 upload argument interface to make clear the files info parameter is no longer used or needed.

packages/web-api/src/methods.ts
filmaj: Alternatively, since we plan on releasing this in a major version, we can drop this argument altogether.

seratch: Thanks! Removing it may prevent smooth upgrade for many people, so I've added a deprecation comment for it.

Issue #1323, opened 2021-08-27
WebClient arguments and responses are not type safe in TypeScript


75asa:
### Description

When using `chat.postMessage()`, I was able to put a non-existent property in `ChatPostMessageArguments` and the process was executed without a compile error!
(Specifically, I typo `icon_url` as `icon_ur`...)
I got a response from @seratch , and it seems that this is largely due to the scope of WebAPICallOptions.
In the future, I'd like to see compilations not go through here, too.

![Actual use of ChatPostMessageArguments](https://user-images.githubusercontent.com/35181442/131130874-053858d6-cb32-4c1e-bfca-af7e5b5b88b7.png)
![methods.d.ts: ChatPostMessageArguments](https://user-images.githubusercontent.com/35181442/131131728-bac4ec04-16c0-49f8-a41e-a2683fabdef6.png)
![WebClient.d.ts: WebAPICallOptions](https://user-images.githubusercontent.com/35181442/131131642-32d07983-6679-41a5-aa9a-d1196055c178.png)

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [x] bug
- [x] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Packages: 

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/interactive-messages`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

#### Reproducible in:

package version: 6.2.4

node version: v16.4.0

OS version(s): MacOS 11.5.2

#### Steps to reproduce:

1.  import `ChatPostMessageArguments` from @slack/web-api
2. define a variable of type ChatPostMessageArguments
3. define a property that does not exist in ChatPostMessageArguments. like a below code

```typescript
[...]

PR #1670, opened 2023-10-04
@slack/web-api making request and response shapes type-safe, and chat.postMessage arguments prototype

filmaj:
⚠️ DO NOT MERGE ⚠️ 

## Purely Intended For Discussion

This draft PR prototypes addressing #1323: the web API arguments and responses not being type-safe due to allowing arbitrary properties in both request and response payloads.

This passes tests locally, and I also integrated with bolt-js, `@slack/oauth` and `@slack/socket-mode` locally (via `npm link`) and the local unit tests in those packages also passed.

**TL;DR**: should we do more fine-grained testing against production of this change to ensure this doesn't break anything?

I think manual testing might be needed before merging this change. There are many different HTTP API methods with a variety of arguments and responses. I think it may be worth testing every method to see if at least basic interactions with each API are possible. 😬 Not sure, what do you think?

On the flip side, if I am able to put the time into manual testing each method, I think there would be many benefits:

1. A good ol' audit and opportunity to add better documentation and make sure things look relevant.
2. Could carefully examine cases where certain APIs accept either/or style of arguments. For example, the [`chat.postMessage` API requires _one of_ `attachments`, `blocks` or `text`](https://api.slack.com/methods/chat.postMessage#args). Currently we model these as optional properties, instead, we should model them as unions. I included a prototype of this in this PR - jump to [this comment of mine](https://github.com/slackapi/node-slack-sdk/pull/1670/files#r1346453396) to see what that looks like in `src/method.ts`.
    a. Could also do this for API response shapes.
    b. This could also be left as an exercise for the future. For example, maybe such an audit, and modifying the request and response payloads for Slack's HTTP API, could be left as something to work on for an eventual web-api 8.0 release.
    c. I am also considering moving these kinds of shapes to the `@slack/types` repo. Let me know if you have thoughts on this (I posted about this in our internal Slack team channel as well). Pulling the types out of this repo would ease re-use in other projects, such as the deno SDK projects.

Depending on the discussion, I could see different potential outcomes / decisions and I am looking for feedback from the team on these:

- We accept just the removal of arbitrary properties from API request/responses, wanting to move forward with addressing _only_ #1323 in the upcoming 7.0 release.
- Regarding more finely typing the request/response shapes, not worth it. Or maybe it is, but not right now. Or maybe it is and we want to do this for the impending 7.0 release. You tell me what you think!
- You like the idea of adding request/response shapes to `@slack/types` and having `@slack/web-api` pull them from there. Or maybe you don't and you think we should keep them here.


packages/web-api/src/WebClient.ts
```diff
@@ -64,10 +64,6 @@ export enum WebClientEvent {
   RATE_LIMITED = 'rate_limited',
 }
 
-export interface WebAPICallOptions {
```
filmaj, line 67: Removing this generic interface, which is the basis for _all_ API request arguments.

packages/web-api/src/WebClient.ts
```diff
@@ -82,7 +78,6 @@ export interface WebAPICallResult {
     // `chat.postMessage` returns an array of error messages (e.g., "messages": ["[ERROR] invalid_keys"])
     messages?: string[];
   };
-  [key: string]: unknown;
```
filmaj, line 85: The other key change to address #1323: request responses allow for arbitrary keys, making them type-unsafe.

packages/web-api/src/WebClient.ts
```diff
@@ -220,11 +216,10 @@ export class WebClient extends Methods {
 
   /**
    * Generic method for calling a Web API method
-   *
    * @param method - the Web API method to call {@link https://api.slack.com/methods}
    * @param options - options
    */
-  public async apiCall(method: string, options: WebAPICallOptions = {}): Promise<WebAPICallResult> {
```
filmaj, line 227: Any areas where we still want things to extend from a shapeless foundation, we can instead use `Record<string, unknown>` or `Record<string, any>` to communicate "an object with any shape but with root keys being strings".

packages/web-api/src/WebClient.ts
```diff
@@ -395,24 +389,25 @@ export class WebClient extends Methods {
...
    * **#3**: Complete uploads {@link https://api.slack.com/methods/files.completeUploadExternal files.completeUploadExternal}
-   * 
+   *
    * **#4**: Unless `request_file_info` set to false, call {@link https://api.slack.com/methods/files.info files.info} for
    * each file uploaded and returns that data. Requires that your app have `files:read` scope.
    * @param options
    */
-  public async filesUploadV2(options: FilesUploadV2Arguments): Promise<WebAPICallResult> {
+  public async filesUploadV2(options: FilesUploadV2Arguments): Promise<
+  WebAPICallResult & { files: FilesCompleteUploadExternalResponse[] | WebAPICallResult[] }
```
filmaj, line 409: Small change to make this part typesafe as well.

packages/web-api/src/WebClient.ts
```diff
@@ -613,12 +608,11 @@ export class WebClient extends Methods {
    * a string, used when posting with a content-type of url-encoded. Or, it can be a readable stream, used
    * when the options contain a binary (a stream or a buffer) and the upload should be done with content-type
    * multipart/form-data.
-   *
    * @param options - arguments for the Web API method
    * @param headers - a mutable object representing the HTTP headers for the outgoing request
    */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  private serializeApiCallOptions(options: WebAPICallOptions, headers?: any): string | Readable {
+  private serializeApiCallOptions(options: Record<string, unknown>, headers?: Record<string, string>): string |
```
filmaj, line 614: Another small change to not have to use `any`.

packages/web-api/src/methods.ts
```diff
@@ -1593,25 +1641,174 @@ export interface ChatPostEphemeralArguments extends WebAPICallOptions, TokenOver
   icon_url?: string; // if specified, as_user must be false
   username?: string; // if specified, as_user must be false
 }
-export interface ChatPostMessageArguments extends WebAPICallOptions, TokenOverridable {
+
```
filmaj, line 1644: This is the area of the code where I prototype modeling the `chat.postMessage` arguments differently, using unions and intersections instead of optional properties.

packages/web-api/src/methods.ts
```diff
@@ -1593,25 +1641,174 @@ export interface ChatPostEphemeralArguments extends WebAPICallOptions, TokenOver
...
+type ChatPostMessageArguments =
+  // Must provide one of the following to determine message contents
+  ChatPostMessageArgumentsContents &
+  // Controls the username/identity and icon display for the message
+  ChatPostMessageArgumentsIdentity &
+  // Controls whether a message is a thread reply or not
+  ChatPostMessageArgumentsType;
+
+const textOnly: ChatPostMessageArguments = {
```
filmaj, line 1791: These are just tiny test cases, to show off how this woud look like.

packages/web-api/src/methods.ts
```diff
@@ -1593,25 +1641,174 @@ export interface ChatPostEphemeralArguments extends WebAPICallOptions, TokenOver
...
+ * have a `thread_ts`.
+ */
+interface ChatPostMessageArgumentsConversationMessage {
+  thread_ts?: never;
+  reply_broadcast?: never;
+}
+
+type ChatPostMessageArgumentsType = ChatPostMessageArgumentsThreadReply | ChatPostMessageArgumentsConversationMessage;
+
+type ChatPostMessageArguments =
```
filmaj, line 1783: This intersection type shows off the core of this approach: assemble the entire shape of a payload using intersections (`&`). Each constituent type within is a union (`|`). This allows us to be able to finely control when certain properties should be available, or not, or optional, in different circumstances.

packages/web-api/src/methods.ts
```diff
@@ -1593,25 +1641,174 @@ export interface ChatPostEphemeralArguments extends WebAPICallOptions, TokenOver
...
+/**
+ * @description `chat.postMessage` arguments for specifying a 'normal' message - not a thread reply - which will never
+ * have a `thread_ts`.
+ */
+interface ChatPostMessageArgumentsConversationMessage {
+  thread_ts?: never;
+  reply_broadcast?: never;
+}
+
+type ChatPostMessageArgumentsType = ChatPostMessageArgumentsThreadReply | ChatPostMessageArgumentsConversationMessage;
```
filmaj, line 1781: This union (`|`) type models a message that is either a conversation message, or a reply-in-thread. Either a message does not specify neither of `thread_ts` and `reply_broadcast` (to model a conversation message, in-channel or DM), or a message has a `thread_ts` and an optional `reply_broadcast` (to model a reply-in-thread).

packages/web-api/src/WebClient.ts
seratch: This is great! but to proceed with this, we may need to review all the argument types to verify if they have latest full list. Have you already done?

packages/web-api/src/WebClient.ts
seratch: If the currently generated response types miss some of response properties, this change may affect TypeScript users. That being said, we can improve it by providing complete typing as much as possible.

packages/web-api/src/WebClient.ts
seratch: yeah, i don't see any issues with this change

packages/web-api/src/WebClient.ts
seratch: good catch

packages/web-api/src/methods.ts
```diff
@@ -1593,25 +1641,174 @@ export interface ChatPostEphemeralArguments extends WebAPICallOptions, TokenOver
...
+}
+
+/**
+ * @description `chat.postMessage` arguments where the `blocks` property captures the message contents.
+ */
+interface ChatPostMessageArgumentsBlocksBased extends ChatPostMessageArgumentsBase {
+  /**
+   * @description The fallback text used in notifications.
+   */
+  text?: string;
```
seratch, line 1712: Is this intentionally optional? In my undertstanding, having `text` is still recommended even when having `blocks`.

packages/web-api/src/methods.ts
```diff
@@ -1593,25 +1641,174 @@ export interface ChatPostEphemeralArguments extends WebAPICallOptions, TokenOver
...
+   * Can only be used by classic Slack apps.
+   * @see {@link https://api.slack.com/methods/chat.postMessage#legacy_authorship Legacy Concerns and Authorship when using `chat.postMessage`}.
+   */
+  as_user?: boolean; // Legacy
+}
+
+/**
+ * @description `chat.postMessage` arguments for specifying the message icon using a URL.
+ */
+interface ChatPostMessageArgumentsIconURL extends ChatPostMessageArgumentsIdentityBase {
```
seratch, line 1731: nice

packages/web-api/src/methods.ts
```diff
@@ -1593,25 +1641,174 @@ export interface ChatPostEphemeralArguments extends WebAPICallOptions, TokenOver
...
+const bs: ChatPostMessageArguments = {
+  channel: 'C1234',
+  blocks: [{
+    type: '',
+  }],
+  thread_ts: '123',
+  reply_broadcast: true,
+  icon_url: '1234',
+};
+console.log(textOnly, attachments, bs);
```
seratch, line 1809: TODO: remove this (perhps the tests in this file too)

seratch (commented):
Great PR! Overall, the changes here look great to me

packages/web-api/src/WebClient.ts
filmaj: I have not, and wanted to get feedback on this point; I mentioned in my original post that this change is significant enough that maybe it requires testing all the endpoints. It sounds to me like you agree that testing is needed. So, I will proceed with that.

packages/web-api/src/methods.ts
filmaj: Yes exactly, this is just a draft to have a discussion.

filmaj:
Thank you for reviewing @seratch 🙇 

I think what I will do is extract the changes that address just #1323 into a separate PR, and start testing each API endpoint to make sure things work, and the types are functional for developers (i.e. no missing properties in request or response payloads, no static type errors prevent issuing API calls, and so on).

Separately / in parallel / bit by bit over time, I plan on moving API and event payloads as types into `@slack/types`. During that process, I will plan on redoing the payloads in a similar style as I did for `chat.postMessage` arguments in this PR.

filmaj:
Closing this PR in favour of https://github.com/slackapi/node-slack-sdk/pull/1673

PR #1672, opened 2023-10-05
`npm link` up the packages in this repo during CI

filmaj:
Allowing for type-level integration testing. Just a small awareness boost, to understand if changes in one of the packages in this monorepo will affect another package in this monorepo that depends on it.

Also fix linter warning re: JSdoc

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -26,4 +26,11 @@ jobs:
       working-directory: ${{ matrix.package }}
       run: |
         npm install || npm list
+        # depending on which package we are testing, also npm link up other dependent packages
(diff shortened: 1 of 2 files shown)


zimeg (approved):
LGTM and a big fan of the guards around typing! :clap: There's [some confusing output from the action](https://github.com/slackapi/node-slack-sdk/actions/runs/6432844795/job/17468638793#step:5:91) but this [seems expected](https://github.com/slackapi/node-slack-sdk/blob/5c52029d713b343f03f59600d21dcb1c2d6fcf9b/packages/web-api/src/WebClient.spec.js#L159) so no worries at all. The formatting change was just a bit confusing to me...

PR #1673, opened 2023-10-05
Web API Type Safety

### LINKED ISSUE #1323: WebClient arguments and responses are not type safe in TypeScript

filmaj:
This work builds off #1670. Web API arguments are worked on to make type safe, though responses were not closely validated as that would require a lot more work. I am also leaving notes on API arguments and responses as I test them out, with breadcrumbs on future improvements / breaking changes.

Once merged, this PR intends to fix #1323.

**USEFUL WHEN ASSEMBLING CHANGELOG**: For the list of breaking changes introduced, best to look at [the `src/methods.ts` file in this commit](https://github.com/slackapi/node-slack-sdk/blob/9dea3ab221482030d8f043cee3c54c633a8b27b6/packages/web-api/src/methods.ts) and search for "TODO".

# Open Questions

- [x] How deeply do we want to type the [Manifest schema](https://api.slack.com/reference/manifests#fields)? The answer here determines changes to still apply to the `apps.manifest.*` APIs in this PR.
  - Answer: somewhat reasonably deeply. Will do my best!

# Deprecations being removed

While here, and since this PR is in prep for a new major release, also took the opportunity to remove methods that were marked as deprecated:

- [x] `channels.*`
- [x] `groups.*`
- [x] `im.*`
- [x] `mpim.*`

#  New Deprecations

We will deprecate the following method groups, given they are mentioned as such in their reference docs.

- [x] `oauth.access` (superseded by the new `oauth.v2.*` methods)
- [x] `stars.*` (docs specifically mention end-users cant interact with them in the client anymore, presumably will be replaced with the Later feature at some point)
- [x] `workflows.*` (legacy steps-from-apps related workflow methods)

# Updates/Testing

Documenting, DRYing up and type-safety-ifying arguments for [all the methods against the documentation](https://api.slack.com/methods). Also writing type tests for the arguments as I go.

- `admin.*`:
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `admin.analytics.getFile` (need to get unblocked on specific enterprise org setup, should be able to get access to that Monday Nov 27):
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `api.test`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `apps.manifest.*` (see open questions above)
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `auth.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `bookmarks.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `calls.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `chat.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `conversations.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `dialog.open`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `dnd.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `emoji.list`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `files.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `migration.exchange`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `oauth.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `openid.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `pins.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `reactions.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `reminders.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `rtm.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `search.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `stars.*`
  - [x] refactored request shapes argument w/ docs
- `team.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `tooling.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `usergroups.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `users.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `views.*`
  - [x] refactored request shapes argument w/ docs
  - [x] type tests
- `workflows.*`
  - [x] refactored request shapes argument w/ docs


packages/web-api/src/WebClient.ts
```diff
@@ -64,10 +64,6 @@ export enum WebClientEvent {
   RATE_LIMITED = 'rate_limited',
 }
 
-export interface WebAPICallOptions {
```
filmaj, line 67: Removing this interface makes the arguments to each API method now 'type safe' (more or less) in that developers can't provide whatever arguments they want anymore (at least not without overriding/ignoring TypeScript)

packages/web-api/src/WebClient.ts
```diff
@@ -82,7 +78,6 @@ export interface WebAPICallResult {
     // `chat.postMessage` returns an array of error messages (e.g., "messages": ["[ERROR] invalid_keys"])
     messages?: string[];
   };
-  [key: string]: unknown;
```
filmaj, line 85: Similarly, this change now prevents responses from being, basically, any shape.

packages/web-api/src/WebClient.ts
```diff
@@ -394,22 +389,22 @@ export class WebClient extends Methods {
...
    * URLs returned from step 1 (e.g. https://files.slack.com/upload/v1/...\")
-   * 
+   *
    * **#3**: Complete uploads {@link https://api.slack.com/methods/files.completeUploadExternal files.completeUploadExternal}
-   * 
    * @param options
    */
-  public async filesUploadV2(options: FilesUploadV2Arguments): Promise<WebAPICallResult> {
+  public async filesUploadV2(options: FilesUploadV2Arguments): Promise<
+  WebAPICallResult & { files: FilesCompleteUploadExternalResponse[] }
```
filmaj, line 406: Small tweak to make this API a bit more type safe with its response.

packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts
```diff
@@ -1,16 +1,8 @@
-/* eslint-disable */
-/////////////////////////////////////////////////////////////////////////////////////////
-//                                                                                     //
-// !!! DO NOT EDIT THIS FILE !!!                                                       //
-//                                                                                     //
-// This file is auto-generated by scripts/generate-web-api-types.sh in the repository. //
```
filmaj, line 6: This file is, in fact, not auto-generated.

packages/web-api/src/methods.ts
```diff
@@ -1985,263 +2048,160 @@ interface FileUploadComplete {
...
   redirect_uri?: string;
-  single_channel?: string;
+  single_channel?: string; // TODO: docs state this is a boolean, only valid for legacy apps
 }
-export interface OAuthV2AccessArguments extends WebAPICallOptions {
-  client_id: string;
-  client_secret: string;
+// https://api.slack.com/methods/oauth.v2.access
+export interface OAuthV2AccessArguments {
+  client_id: string; // TODO: docs state this is optional
```
seratch, line 2127: The document might be wrong; I don't know what the possible scenario to call this API without the pair could be

packages/web-api/src/methods.ts
```diff
@@ -2399,88 +2393,108 @@ export interface UsergroupsUsersUpdateArguments extends WebAPICallOptions, Token
...
   trigger_id: string;
-  view: View;
 }
-
-export interface ViewsPublishArguments extends WebAPICallOptions, TokenOverridable {
+interface ViewInteractivityPointer {
+  interactivity_pointer: string;
+}
+// https://api.slack.com/methods/views.open
+export type ViewsOpenArguments = BaseViewsArguments & TokenOverridable & (ViewTriggerId | ViewInteractivityPointer);
```
seratch, line 2461: nice

seratch (commented):
Great work! I left a few minor comments

seratch:
>Long-term, I think we should keep the API argument and response shapes in the @slack/types package. Thoughts? 👍 / 👎 ?

This is fine, but doing tests towards web-api package changes could be harder with the approach (meaning, linking the local revision of types package is necessary to do tests). What do you think?

packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts
```diff
@@ -24,6 +16,7 @@ export interface ResponseMetadata {
 
 export interface AdminAnalyticsMemberDetails {
   enterprise_id:                 string;
+  team_id:                       string;
```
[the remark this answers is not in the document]
filmaj: Awesome, I think I will spend the required time to review all the java-slack-sdk responses, make sure the node sdk has parity with it and document them as well as I go.

packages/web-api/src/methods.ts
filmaj: I think I will spend more time and just make similar types of changes for all API arguments...

filmaj:
> This is fine, but doing tests towards web-api package changes could be harder with the approach (meaning, linking the local revision of types package is necessary to do tests). What do you think?

Hmm, how do you see it being harder? Isn't that also the case for e.g. the `Blocks` / `KnownBlocks` interfaces for Block Kit?

seratch:
@filmaj I might misunderstand what you meant by the sentence. I thought that you're considering to move more types such as argument interfaces from web-api to types. If yes, we will need to maintain two packages at a time when adding a new API support. But perhaps that's not the case. Never mind!

packages/web-api/src/methods.ts
```diff
@@ -979,361 +923,464 @@ export interface TraditionalPagingEnabled {
...
-export interface AdminAppsUninstallArguments extends WebAPICallOptions {
+// TODO: breaking changes - potential type improvements:
+// - exactly one of `team_id` or `enterprise_id` is required - but not both
+// https://api.slack.com/methods/admin.apps.uninstall
+export interface AdminAppsUninstallArguments {
   app_id: string;
-  enterprise_id?: string;
-  team_ids?: string[];
+  enterprise_id?: string; // TODO: breaking change, enforce at least one array item?
+  team_ids?: string[]; // TODO: breaking change, enforce at least one array item?
```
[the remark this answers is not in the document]
seratch: Ah I see. Going with this looks great to me

filmaj:
That was indeed my hope 😅 for one reason: making them reusable in other projects, like Deno SDK project. Or personal projects directly speaking with certain slack APIs.

It was just an idea, not necessary right now, but I will continue thinking about it.. Maybe there are other ways of achieving reuse of these shapes in other projects/runtime.

packages/web-api/src/response/AdminAnalyticsGetFileResponse.ts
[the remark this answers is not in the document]
filmaj: I will spend some time setting up the full java-slack-sdk integration tests as I want to see if it would be possible to somehow generate documentation for these fields with a semi-automated method as well.

@seratch the documentation in the generated Java fields, like [here](https://github.com/slackapi/java-slack-sdk/blob/main/slack-api-client/src/main/java/com/slack/api/methods/response/admin/analytics/AdminAnalyticsGetFileResponse.java#L144-L148), are they hand-written?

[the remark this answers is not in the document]
seratch: @filmaj Yes, they are. The benefits of the tests in the SDK is to detect unknown properties. When enabling [this detection mode](https://github.com/slackapi/java-slack-sdk/blob/v1.34.0/slack-api-client/src/test/java/config/SlackTestConfig.java#L81C16-L81C42), JSON parser throws an exeption with a newly added property information. The unit test results in a failure, and then we, developers, can modfiy the code until it passes the test without any exceptions.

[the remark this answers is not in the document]
filmaj: I see, that makes sense! Maintainers of this project should follow the same approach. I will do my best to document the fields in a nice way as I work through this.

packages/web-api/src/types/request/bots.ts
```diff
@@ -0,0 +1,7 @@
+import type { OptionalTeamAssignable, TokenOverridable } from './common';
+
+// https://api.slack.com/methods/bots.info
+export interface BotsInfoArguments extends TokenOverridable, OptionalTeamAssignable {
+  /** @description Bot user ID to retrieve information about. */
```
seratch, line 5: Bot user ID can be misleading because many of our incoming payloads can contain bot_id (B...) and bot_user_id (U/W...)
```suggestion
  /** @description Bot ID, which starts with 'B', to retrieve information about. */
```

packages/web-api/src/types/request/chat.ts
```diff
@@ -0,0 +1,224 @@
+import type { KnownBlock, Block, MessageAttachment, LinkUnfurls, MessageMetadata } from '@slack/types';
```
seratch, line 1: I am still not 100% confident that all the patterns here are correct ... but well done!

packages/web-api/src/types/request/manifest.ts
```diff
@@ -0,0 +1,547 @@
...
+
+interface ManifestFeatures {
+  /**
+   * @description A subgroup of settings that describe {@link https://api.slack.com/surfaces/app-home App Home} configuration.
+   * @see {@link https://api.slack.com/surfaces/app-home App Home}.
+   */
+  app_home?: ManifestAppHome;
+  /**
+   * @description A subgroup of settings that describe {@link https://api.slack.com/legacy/enabling-bot-users bot user} configuration.
+   * @see {@link https://api.slack.com/legacy/enabling-bot-users Legacy bots}.
```
seratch, line 55: I don't like linking this "legacy" document here but indeed there is no other document for this

packages/web-api/src/types/response/AdminAnalyticsGetFileResponse.ts
```diff
@@ -24,6 +16,7 @@ export interface ResponseMetadata {
 
 export interface AdminAnalyticsMemberDetails {
   enterprise_id:                 string;
+  team_id:                       string;
```
seratch, line 19: good catch

packages/web-api/test/types/methods/admin.analytics.test-d.ts
```diff
@@ -0,0 +1,45 @@
+import { expectAssignable, expectError } from 'tsd';
```
seratch, line 1: Thanks for adding these tests!

scripts/code_generator.rb
```diff
@@ -5,7 +5,7 @@
```
seratch, line 5: Perfect!

seratch (approved):
I can't guarantee everything works as with before but I am confident that this PR is carefully crafted. Great work!

packages/web-api/src/types/request/chat.ts
filmaj: Yes I think this is a risky area.. all possible combinations of Block elements, etc, may be hard to verify... but I think as long as we are OK to tell developers that find a problem, "please use `// @ts-expect-error` as a workaround until we release a patch", it should be OK.

packages/web-api/src/types/request/manifest.ts
filmaj: I took this from the Manifest fields documentation here: https://api.slack.com/reference/manifests#features

The link originally links to api.slack.com/bot-user which redirects to the above link.

packages/web-api/src/WebClient.ts
```diff
@@ -306,28 +304,24 @@ export class WebClient extends Methods {
...
     reduce?: PageReducer<A>,
   ): Promise<A>;
   public paginate<R extends PageReducer, A extends PageAccumulator<R>>(
     method: string,
-    options?: WebAPICallOptions,
+    options?: Record<string, unknown>,
     shouldStop?: PaginatePredicate,
     reduce?: PageReducer<A>,
   ): (Promise<A> | AsyncIterable<WebAPICallResult>) {
-    if (!cursorPaginationEnabledMethods.has(method)) {
```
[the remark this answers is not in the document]
seratch: Thanks for clarifying. Makes sense!

PR #1690, opened 2023-11-07
Add support for apps.manifest.* endpoints

### LINKED ISSUE #1340: Add Support for apps.manifest.* Endpoints
### Description

Add support for the following `apps.manifest.*` methods:

`apps.manifest.create`
`apps.manifest.validate`
`apps.manifest.update`
`apps.manifest.delete`
`apps.manifest.export`

- [ ] bug
- [ ] question
- [ ] documentation related
- [ ] testing related
- [ ] discussion

misscoded:
###  Summary

Fixes #1340


packages/web-api/src/methods.ts
```diff
@@ -1456,6 +1475,28 @@ export interface AppsEventAuthorizationsListArguments
...
+  app_id: string;
+}
+
+export interface AppsManifestUpdateArguments extends WebAPICallOptions, TokenOverridable {
+  app_id: string;
+  manifest: string;
+}
+
+export interface AppsManifestValidateArguments extends WebAPICallOptions, TokenOverridable {
+  manifest: string;
```
seratch, line 1497: It's also possible to have app_id if you want to do so:
```suggestion
  app_id?: string;
  manifest: string;
```
reference: https://api.slack.com/methods/apps.manifest.validate

seratch (changes requested):
Thanks for the addition. Mentioned one missing optional parameter

prod-server-integration-tests/scripts/apps_manifest_test.js
```diff
@@ -0,0 +1,77 @@
...
+const options = {};
+
+const BOT_TOKEN = process.env.SLACK_BOT_TOKEN;
+const TOOLING_TOKEN = process.env.SLACK_TOOLING_TOKEN;
+
+if (process.env.SLACK_SDK_TEST_DEV_API_URL !== undefined) {
+  options.slackApiUrl = process.env.SLACK_SDK_TEST_DEV_API_URL;
+}
+
+const client = new WebClient(BOT_TOKEN, options);
```
seratch, line 12: We should verify if a WebClient instance can work without a bot token in this scenario.

misscoded: I think the token used for `BOT_TOKEN` is irrelevant for the testing of the endpoints since they rely on the tooling token override, correct? The endpoints will not work with a non-tooling token, if that's what you mean.

seratch: Yes, you're right. The bot token is not necessary at all. Removing it makes the code more realistic and we should demonstrate how to use the API client usually.

misscoded: > Removing it makes the code more realistic

Won't most folks using the manifest API have a `client` that's instantiated using a bot token (or user token)? Happy to remove it, but want to clarify we're on the same page.

seratch: When it comes to a bolt app, yes it is. However, in this case, a developer using the client for the app manifest scenario may not have any bot token, right? (say, starting a new app development). Indeed, a developer may initialize the client with a bot token (as part of a larger code base or due to whatever reasons), but this assuption does not cover all use cases. The e2e test code should make sure if it works for most common use cases. Regarding the APIs we're working on now, I believe the primary one is without a bot token, and then secoundary could be with a bot token. Does this make sense to you?

seratch (approved):

prod-server-integration-tests/scripts/apps_manifest_test.js
seratch: Just approved. Once this integration test is improved, go ahead with merging this PR 👍

prod-server-integration-tests/scripts/apps_manifest_test.js
```diff
@@ -0,0 +1,77 @@
+const { WebClient } = require('@slack/web-api');
+
+const options = {};
+
+const BOT_TOKEN = process.env.SLACK_BOT_TOKEN;
```
seratch, line 5: You can remove this line

filmaj:
FYI I have cherry-picked this merged commit to the `web-api-6.x` maintenance branch and will cut a 6.10 release of web-api from there.

PR #1702, opened 2023-12-05
Add support for functions.* (complete) methods

misscoded:
###  Summary

Adds support for the following `functions.*` methods:

`functions.completeError`
`functions.completeSuccess`


packages/web-api/src/methods.ts
```diff
@@ -2077,6 +2087,19 @@ export interface FilesRemoteShareArguments extends WebAPICallOptions, TokenOverr
...
+ * `functions.*`
+ */
+export interface FunctionsCompleteErrorArguments extends WebAPICallOptions, TokenOverridable {
+  function_execution_id: string;
+  error: string;
+}
+
+export interface FunctionsCompleteSuccessArguments extends WebAPICallOptions, TokenOverridable {
+  function_execution_id: string;
+  outputs?: Record<string, unknown>;
```
seratch, line 2100: I believe this paramter is required (our work-in-progress document says so too)

seratch (changes requested):

packages/web-api/src/methods.ts
misscoded: You're probably right. I'll retest to be doubly certain and remove optionality upon confirmation.

misscoded: Although, to be fair, `outputs` may not be passed on so it's odd to require them!

seratch: if i remember correctly, an empty `outputs` object might be required even when the function does not have any required output_parameters in its definition.

filmaj: Worth testing as you can also set `complete:false` on a function in the case it has interactivity sub-steps involved (such as the function creates a modal, which requires user input and interaction).

filmaj: Indeed, on the deno side, `outputs` is a required property (see [this code](https://github.com/slackapi/deno-slack-api/blob/main/src/typed-method-types/functions.ts#L3-L5)).

filmaj (changes requested):
Thanks for tackling this! Yep, Kaz is right, `outputs` is required when marking a function as complete.

filmaj (approved):
:shipit:

seratch (approved):
Thanks for working on this; now this looks good to me :+1:

PR #1714, opened 2024-01-08
Introduce a `@slack/cli-hooks` package that implements Slack CLI hooks

zimeg:
###  Summary

This PR introduces a standalone package that implements hooks for the Slack CLI, meant for use in Bolt for JavaScript apps.

Early versions of this package are published to NPM here: https://www.npmjs.com/package/@slack/cli-hooks

### Reviewers

Update a new Bolt app with the latest version of the CLI and the following commands:

```sh
$ slack create hooks --template slack-samples/bolt-js-custom-function-template

# Confirm a manifest.json file is found
$ slack manifest

# Check that the app starts without problems
$ slack run

# Verify dependencies with updates are shown
# Note: Installing updates is not yet supported
$ slack update
```

### Notes

A few quick questions rose during development:

- ~Should the `start` hook default file be `app.js` or `index.js`?~ 👍 
- ~Is the `install-update` hook alright to implement as a follow up?~ 👍

~Some tasks remain as TODOs as well: https://github.com/slackapi/node-slack-sdk/pull/1714#issuecomment-1888277249~ 👍

### Requirements


packages/hooks/README.md
```diff
@@ -0,0 +1,105 @@
+# Slack CLI Hooks
+
+The `@slack/hooks` package contains scripts that implement the contract between
```
misscoded, line 3: Should we name this directory/project `cli-hooks` to make it clear what it's used for, or is there a potential alternative use case that we're omitting that for?

packages/hooks/README.md
```diff
@@ -0,0 +1,105 @@
+# Slack CLI Hooks
```
misscoded, line 1: We went back and forth a bit on this, but I prefer to have relative parity between this and what we establish as the README design proposed on the Python hooks package (absent of a good reason not to, of course).

packages/hooks/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@slack/hooks",
+  "version": "0.0.6",
+  "description": "Node implementations of the contract between the Slack CLI and Bolt for JavaScript",
```
misscoded, line 4:
```suggestion
  "description": "Node implementation of the contract between the Slack CLI and Bolt for JavaScript",
```

packages/hooks/src/check-update.js
```diff
@@ -0,0 +1,330 @@
+#!/usr/bin/env node
+import fs from 'fs';
+import util from 'util';
+import childProcess from 'child_process';
+
+const exec = util.promisify(childProcess.exec);
+
+const SLACK_BOLT_SDK = '@slack/bolt';
+const SLACK_CLI_HOOKS = '@slack/hooks';
+const SLACK_DENO_SDK = '@slack/deno-slack-sdk';
```
misscoded, line 10: Can't we remove this, as well as the subsequent references to it?

packages/hooks/src/check-update.js
```diff
@@ -0,0 +1,330 @@
...
+  /** @type {string[]} */
+  const dependencyErrors = [];
+  const releases = Object.entries(versionMap).map(([dependency, version]) => {
+    if (version.error) {
+      dependencyErrors.push(dependency);
+    }
+    return version;
+  });
+  return {
+    name: 'the Slack SDK',
```
misscoded, line 303: Is this correct?

packages/hooks/src/check-update.js
```diff
@@ -0,0 +1,330 @@
...
+    message: '',
+    releases,
+    url: 'https://api.slack.com/automation/changelog',
```
misscoded, line 306: Again, is this what we would want to point them to?

packages/hooks/src/get-hooks.js
```diff
@@ -0,0 +1,18 @@
...
+// eslint-disable-next-line no-console
+console.log(JSON.stringify({
+  hooks: {
+    'get-manifest': 'npx -q --no-install -p @slack/hooks slack-cli-get-manifest',
+    'check-update': 'npx -q --no-install -p @slack/hooks slack-cli-check-update',
+    start: 'npx -q --no-install -p @slack/hooks slack-cli-start',
+  },
+  config: {
+    watch: {
+      'filter-regex': '^manifest\\.(ts|js|json)$',
```
misscoded, line 11: Have we confirmed that `.ts` and `.js` work as expected? Per [`get-manifest.js`](https://github.com/slackapi/node-slack-sdk/pull/1714/files#diff-be817d63e5b2302e9b335351016f0a0c2de6c5488527691df45e45175df73b6bR21), I believe we only support `manifest.json` atm.

packages/hooks/src/start.js
```diff
@@ -0,0 +1,49 @@
...
+  app.stdout.setEncoding('utf-8');
+  app.stdout.on('data', (data) => {
+    process.stdout.write(data);
+  });
+  app.stderr.on('data', (data) => {
+    process.stderr.write(data);
+  });
+
+  app.on('close', (code) => {
+    console.log(`bolt-app local run exited with code ${code}`); // eslint-disable-line no-console
```
misscoded, line 35:
```suggestion
    console.log(`Local run exited with code ${code}`); // eslint-disable-line no-console
```

packages/hooks/src/start.js
```diff
@@ -0,0 +1,49 @@
...
+const { main: pkgJSONMain } = JSON.parse(fs.readFileSync('./package.json', 'utf-8'));
+
+/**
+ * Start hook implementation that verifies and runs an app in Socket Mode.
+ * @param {string} cwd - The current working directory of the project.
+ * @param {string | undefined} customPath - An optional path for the app.
+ */
+(function _(cwd, customPath) {
+  // TODO - Format so that its less miss-able in output
+  console.log('Preparing local run in developer mode (Socket Mode)'); // eslint-disable-line no-console
```
misscoded, line 15: Something about the phrase "local run" is niggling at me. That makes a lot of sense in the hosted vs. local paradigm, but less so in Bolt world. Thoughts on this? Just me?

zimeg: I'm kinda leaning towards removing the first two logs in this hook entirely 😳 Or possibly included these in some message-boundary and displayed only in CLI debug logs?

I agree that it's a strange output. Especially in the Bolt context. Ignoring `DEBUG` logs from the Bolt app itself, we're left with this output which feels noisy to me...

```txt
$ slack run -w largespeck
Updating local app install for "largespek"
✨ somename of largespek
Preparing local run in developer mode (Socket Mode)
/Users/somename/programming/samples/bolt-js-starter-template/app.js
[INFO]  socket-mode:SocketModeClient:0 Going to establish a new connection to Slack ...
⚡️ Bolt app is running! ⚡️
[INFO]  socket-mode:SocketModeClient:0 Now connected to Slack
```

It's a really good question though, since we're advising `slack run` is only used for development and something else for deployed apps.

Maybe this line changes to something like "Preparing development app in Socket Mode"?

packages/hooks/src/start.js
zimeg: I'm in favor of keeping this log though! It only seems to appear if the app crashes which is pretty useful. And this is a nice change! Thanks 😄

packages/hooks/src/get-hooks.js
zimeg: Totally right that `.ts` and `.js` isn't supported right now. I'll remove these!

zimeg (commented):
@misscoded Really appreciate the review! Going to follow up with a few updates to the README and an update to `get-manifest` but will hold off on changes to naming and removal of Deno for right now. Definitely open to making these changes soon though!

packages/hooks/README.md
```diff
@@ -0,0 +1,106 @@
...
+section.
+
+## Requirements
+
+This package supports Node v18 and higher. It's highly recommended to use [the
+latest LTS version of Node][node].
+
+An updated version of the Slack CLI is also encouraged while using this package.
+
+## Installation
```
zimeg, line 25: Including this to match other Node packages, but open to removing it. I'm thinking any errors encountered are expected in early versions, but not thinking very much

WilliamBergamin (changes requested):
This looks great 💯 I left a few suggestions and NITs

> Should the start hook default file be app.js or index.js?

Our templates and samples use `app.js` defaulting to it makes sense

> Is the install-update hook alright to implement as a follow up?

I say go for it but I think there is also value in implementing the message boundary protocol

packages/hooks/src/check-update.js
misscoded: My bad. I'm sure I mistakenly left it in. Let's pull it (and verify all works as expected, ofc).

packages/hooks/src/check-update.js
misscoded: I've always disliked what we chose for the Deno one since it feels so general. No harm though in staying the course: let's keep it for now!

packages/hooks/src/check-update.js
misscoded: Ah, thanks for the visual! That's helpful. I guess my question is will the `hooks` package update be present in the Changelog with everything else? If so, it makes sense. If not, we probably need to tweak the CLI logic at some point.

packages/hooks/src/start.js
misscoded: I vote remove them then. Keep it clean and tidy!

filmaj:
Gonna start my review now, but re: your questions in your original comment:

> Should the start hook default file be app.js or index.js?

[slack.dev getting started instructions](https://slack.dev/bolt-js/tutorial/getting-started) use `app.js`, so that has my vote.

> Is the install-update hook alright to implement as a follow up?

100%

filmaj:
@zimeg one thing I noticed: I have the sample up and running, but I can't add the function bundled in the app to WFB when I deploy it to my test workspace. I think because the output of `slack manifest` doesn't include any `functions`, so there's no 'custom step' in WFB to add.

zimeg:
@filmaj Sorry about that! This sample might perform a bit better: https://github.com/slack-samples/bolt-js-automation-template

filmaj:
@zimeg got farther with the automation template. I can create a workflow in WFB that includes the sample function, but when I go to run the workflow, I get the dreaded `event_dispatch_failed` error in my logs from local-run:

```
[INFO]  socket-mode:SocketModeClient:0 Now connected to Slack
[DEBUG]  socket-mode:SocketModeClient:0 Transitioning to state: connected:preparing
[DEBUG]  socket-mode:SocketModeClient:0 Started running a new heart beat job
[DEBUG]  socket-mode:SocketModeClient:0 Transitioning to state: connected:ready
2024-01-11 15:13:27 [info] [Fn06DBNWMACW] (Trace=Tr06D3S1KR2B) Function 'Sample function' (app function) started
2024-01-11 15:13:27 [error] [Fn06DBNWMACW] (Trace=Tr06D3S1KR2B) Function 'Sample function' (app function) failed
	event_dispatch_failed
```

packages/cli-hooks/src/check-update.js
```diff
@@ -0,0 +1,330 @@
...
+import childProcess from 'child_process';
+
+const exec = util.promisify(childProcess.exec);
+
+const SLACK_BOLT_SDK = '@slack/bolt';
+const SLACK_CLI_HOOKS = '@slack/cli-hooks';
+const SLACK_DENO_SDK = '@slack/deno-slack-sdk';
+
+/**
+ * @typedef {object} UpdateInfo
```
filmaj, line 13: yeahhhh nice! jsdoc typedefs, take that TypeScript!

packages/cli-hooks/src/get-hooks.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
```
filmaj, line 1: What about framing these hooks as standard javascript modules/functions, but detecting whether they are invoked from the CLI using `require.main`? This would allow us to keep the source to be runnable in node natively, and also allows for running the tests in node.

Consider this file `test.js`:

```
module.exports = function () {
  return {some: 'thing'};
}

if (require.main === module) {
  console.log('main is its own module!');
}
```

Now if I run this file from the command line:

```
➜ node test.js
main is its own module!
```

But if I `require` it within node:

```
➜ node
Welcome to Node.js v18.15.0.
Type ".help" for more information.
> require('./test.js')
[Function (anonymous)]
>
```

I think we can use this pattern to:
- not have to hashbang the env to node at the top of each of the hook
- we can have hooks be wrapped in a function that return objects, making it easier to test. Each test file can invoke the hook function. Hook functions can optionally take arguments, too (many of our hooks accept CLI flags) so testing those can be a pain if you have to frame each test in terms of a process invocation (not to mention that should significantly speed the tests up, since we're not shelling out to additional processes).

So I'd envision the get-hooks implementation to look something like:

```
module.exports = function () {
  return { hooks: {}, whatever: {} };
}

if (require.main === module) {
  console.log(JSON.stringify(module.exports()));
}
```

So we leave the basic, condition-less processing of JSON into strings and outputting to stdout to the `if (require.main === module)` conditional, while any actual hook logic is kept to pure javascript in a simple function.

filmaj (commented):
First round of comments in. The big one is how to structure the hooks in terms of testability. I'd like to have a discussion about this approach and see what you think about my suggestion.

I can do a follow-up review once the big testability question is settled.

packages/cli-hooks/src/get-hooks.js
zimeg: This seems so very much cleaner than the magic of IIEF in scripts! Really appreciate the suggestion and will start hacking on it soon 🙌

packages/hooks/src/get-hooks.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
zimeg: Great catch! Looking into this addition for this PR. 🙇

packages/hooks/src/check-update.js
zimeg: +1 to future changes in the CLI since I don't think these `hooks` will be included in the changelog / follow the same release cycle. Maybe we do align these releases though?

For now I think this is alright to leave but will add it to the backlog!

zimeg:
@misscoded @WilliamBergamin @filmaj I really appreciate all of the reviews on this! So much good feedback. Currently in the process of implementing some changes but don't think I'll get around to everything until early next week.

Currently planning the following, but please let me know if I'm missing anything or should hold off on some things for now:

- [ ] Use JavaScript modules instead of IIFE
- [ ] Remove references to the Deno packages
- [ ] Implement the message boundary protocol
- [ ] Compare versions with the [node-semver](https://www.npmjs.com/package/semver) package
- [ ] Refresh the README with a few changes

I'll send a ping once those checks are checked or if there are questions beforehand, but thank y'all once more for the kind and quick reviews 🙇

packages/hooks/src/get-hooks.js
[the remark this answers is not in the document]
zimeg: Currently in the process of implementing this and am wondering if the `get-hooks` should have both the `message-boundaries` and `default` [protocols specified](https://github.com/slackapi/python-slack-hooks/blob/efed2b0f85e6bf8d2e6e860ee94e35ab519d3c3b/slack_cli_hooks/hooks/get_hooks.py#L22) or just the `message-boundaries` [protocol](https://github.com/slackapi/deno-slack-hooks/blob/ffafaf4f5e598b0cfaecb2a4b862f94b8603f30d/src/mod.ts#L23)?

In either case, both will be supported by `cli-hooks`, just curious about the `get-hooks` implementation!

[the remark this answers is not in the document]
filmaj: Supporting both puts no lower bound on the minimum CLI version. Supporting only message boundaries puts the required minimum CLI version to.. well, whatever the first version to support it was, I forget.
Your call! CLI new version adoption is very good so I don't think it is a risky or limiting move to only support message boundaries.

packages/hooks/src/start.js
zimeg: Update: These are logs are removed. Clean and tidy just feels better. Also, the error message for a missing `app.js` (or whatever's specified in the `package.json` is clear IMO)

<details>
<summary>This is the error message</summary>

```sh
$ pwd
/private/tmp/hooks
$ slack run
? Choose a local environment devrelsandbox T038J6TH5PF A06FQE9PVK4
Updating local app install for "DevRel Sandbox"
✨ ethan.zimbelman of DevRel Sandbox
node:internal/modules/cjs/loader:1051
  throw err;
  ^

Error: Cannot find module '/private/tmp/hooks/apppppppppppp.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1048:15)
    at Module._load (node:internal/modules/cjs/loader:901:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:83:12)
    at node:internal/main/run_main_module:23:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v20.8.0
Local run exited with code 1
```

packages/hooks/src/check-update.js
```diff
@@ -0,0 +1,330 @@
...
+  return targetPatch > currPatch;
+}
+
+/**
+ * Checks if updating a dependency to the latest version causes a breaking change.
+ * @param {string | undefined} current - Package version available in the project.
+ * @param {string | undefined} latest - Most up-to-date dependency version available.
+ * @returns {boolean} If the update will result in a breaking change.
+ */
+export function hasBreakingChange(current, latest) {
```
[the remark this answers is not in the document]
zimeg: Really solid call using `semver` it's handling cases of release candidates exactly as specified. So much hassle saved!

packages/hooks/src/get-hooks.js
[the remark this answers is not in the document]
zimeg: And now there is no minimum to the supported CLI version! Nor upper bound, if I can be so confident. At least for the foreseeable times. Both the default and message-boundaries protocols should be working now :)

packages/cli-hooks/src/get-hooks.js
zimeg: Attempts to use `require.main` raised errors because these scripts are ES modules instead of common JS (I think). I've found that the following has the same effect:

```js
if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
  console.log(JSON.stringify(getHooks())); // eslint-disable-line no-console
}
```

and makes the tests so very much faster! But always open to iterations!

zimeg:
@misscoded @WilliamBergamin @filmaj A few patches later and I believe this is ready for re-review! I've updated the steps to review to hopefully be easier (everything should be working right away) but you will need a dev build of the CLI to handle automagic tokens.

Please let me know if there's anything more that needs updating! Or just changing! 🙏

filmaj:
I had some issues testing this out from scratch, I think because the template repo you linked to is internal:

```
10:40:14 in ~/src
➜ slak create remote-func-js --template slack-samples/bolt-js-custom-function-template
⚙️  Creating a new Slack app in ~/src/remote-func-js

Username for 'https://github.com': filmaj
Password for 'https://filmaj@github.com':
Check /Users/fmaj/.slack/logs/slack-debug-20240126.log for full error logs

🚫 Error: While cloning repository, the following error occurred:

git clone --depth=1 https://github.com/slack-samples/bolt-js-custom-function-template.git /Users/fmaj/src/remote-func-js

Cloning into '/Users/fmaj/src/remote-func-js'...
remote: Support for password authentication was removed on August 13, 2021.
```

Even entering my username and password doesn't work as GitHub seems to block those attempts.

Another sharp edge: I then tried just manually cloning the repo and running `slack manifest` within - but forgetting to run `npm i` within. The error I get is:

```
➜ slak manifest
Check /Users/fmaj/.slack/logs/slack-debug-20240126.log for full error logs

🚫 There was an error while reading the Slack Configuration file (`slack.json`) or running the `get-hooks` hook (sdk_config_load_error)

Error Details:

1: Command for '' returned an error: exit status 1
npm ERR! canceled

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/fmaj/.npm/_logs/2024-01-26T15_44_44_782Z-debug-0.log (sdk_hook_invocation_failed)
Suggestion: Run `slack doctor` to check that your system dependencies are up-to-date.


Suggestion:

Run `slack doctor` to check that your system dependencies are up-to-date.
```

Is there any way to return a nicer / clearer error in this case?

filmaj:
But - in general works GREAT! AWESOME!

Screenshot 2024-01-26 at 10 55 33 AM

Screenshot 2024-01-26 at 10 55 15 AM

packages/cli-hooks/src/get-hooks.js
filmaj: As long as it works in both node 18 and 20 I'm cool with it

packages/cli-hooks/src/check-update.js
```diff
@@ -1,13 +1,25 @@
 #!/usr/bin/env node
```
filmaj, line 1: I think we can drop these now, yeah?

packages/cli-hooks/src/get-hooks.js
```diff
@@ -0,0 +1,62 @@
...
+ * Printed as an object containing featured provided by the SDK.
+ */
+
+if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
+  console.log(JSON.stringify(getHooks())); // eslint-disable-line no-console
+}
+
+/**
+ * Standardized communication format between the SDK and CLI regarding hooks.
+ * @typedef SDKInterface
```
filmaj, line 17: Great job documenting these

packages/cli-hooks/src/get-hooks.js
```diff
@@ -0,0 +1,62 @@
...
+        'filter-regex': '^manifest\\.json$',
+        paths: [
+          '.',
+        ],
+      },
+      'protocol-version': ['message-boundaries', 'default'],
+      'sdk-managed-connection-enabled': true,
+    },
+    runtime: 'node',
+  };
```
filmaj, line 61: The cleanliness of this implementation gives me the warm n fuzzies

packages/cli-hooks/src/get-manifest.js
```diff
@@ -38,9 +43,12 @@ function readManifestJSONFile(searchDir, filename) {
...
     }
-  } catch (error) {
-    console.error(error); // eslint-disable-line no-console
-    throw new Error('Failed to parse the manifest file for this project');
+  } catch (err) {
+    let message = 'Failed to parse the manifest file for this project';
+    if (err instanceof Error) {
+      message += `\n${err.name}: ${err.message}`;
+    }
+    throw new Error(message);
```
filmaj, line 51: Nicely done wrapping this error.

Not directly related to this PR or your work, and I think this issue has existed before, but it seems like in the case a hook returns a non-zero code to the CLI, the CLI seems to lose spacing of the output, for some reason?

Here is a screenshot of directly invoking the cli-hooks script from a project with a malformed manifest (left) vs. doing so w/ the CLI (right):

Screenshot 2024-01-26 at 11 14 17 AM

Maybe just something to file over on the CLI as a future fix.

packages/cli-hooks/package.json
```diff
@@ -35,7 +36,7 @@
   "scripts": {
     "prebuild": "shx rm -rf ./coverage",
     "build": "shx chmod +x src/*.js",
-    "prelint": "tsc --noemit --project ./jsconfig.json",
+    "prelint": "tsc --noemit --module es2022 --project ./jsconfig.json",
     "lint": "eslint --ext .js src",
     "pretest": "npm run lint -- --fix",
     "test": "c8 mocha src/*.spec.js"
```
filmaj, line 42: Can we tweak this to output the coverage?

packages/cli-hooks/src/protocols.js
```diff
@@ -0,0 +1,119 @@
+import minimist from 'minimist';
+
+/**
+ * An interface encapsulating a specific set of communication rules that the
+ * SDK and the CLI both implement.
+ * @typedef {object} Protocol
+ * @property {string} name -
+ *   Label representing the name of the protocol as expected by the Slack CLI.
+ * @property {typeof console.log} log -
```
filmaj, line 9: Wow does using `typeof` actually work for setting a type in JSDoc?

packages/cli-hooks/src/get-hooks.js
```diff
@@ -0,0 +1,62 @@
...
+      start: 'npx -q --no-install -p @slack/cli-hooks slack-cli-start',
+    },
+    config: {
+      watch: {
+        'filter-regex': '^manifest\\.json$',
+        paths: [
+          '.',
+        ],
+      },
+      'protocol-version': ['message-boundaries', 'default'],
```
filmaj, line 57: Should we pull these protocol names from the protocol source file in this package? The reason I bring this up is that `protocol.js` defines a `SUPPORTED_NAMED_PROTOCOLS` array that _only_ includes the message boundary protocol, but by returning this line in get-hooks, we're telling the CLI that either of the two named protocols is fine to use (with a preference for message-boundaries). We should decide on one approach.

packages/cli-hooks/src/protocols.spec.js
```diff
@@ -0,0 +1,81 @@
...
+      consoleLogStub = sinon.stub(console, 'log');
+    });
+
+    afterEach(() => {
+      sinon.restore();
+    });
+
+    it('errors if no boundary is specified', () => {
+      assert.throws(() => {
+        MessageBoundaryProtocol([]);
```
filmaj, line 46: I'd add an assertion here on the error text, in case we introduce a different kind of bug in the future raising a different kind of error.

packages/cli-hooks/src/protocols.spec.js
```diff
@@ -0,0 +1,81 @@
...
+    it('uses console log methods without a manifest', () => {
+      const protocol = BaseProtocol(['--flag']);
+      assert.equal(protocol.log, console.log);
+      assert.equal(protocol.error, console.log);
+      assert.equal(protocol.warn, console.log);
+      assert.equal(protocol.respond, console.log);
+    });
+  });
+
+  describe('message boundary protocol', () => {
```
filmaj, line 32: Excellent work!

packages/cli-hooks/src/start.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Confirms environment variables are prepared by the CLI.
+ */
+function validateEnvironment() {
+  if (!process.env.SLACK_CLI_XOXB) {
+    throw new Error('Missing local run bot token. Please see Slack CLI maintainers to troubleshoot.');
+  }
+  if (!process.env.SLACK_CLI_XAPP) {
+    throw new Error('Missing local run app token. Please see Slack CLI maintainers to troubleshoot');
```
filmaj, line 65: These are probably not the instructions we should provide to devs hitting these errors... not sure what else to suggest though 🤔

filmaj (changes requested):
Awesome work, very close to being ready I think!

packages/cli-hooks/src/protocols.js
```diff
@@ -0,0 +1,119 @@
...
+const SUPPORTED_NAMED_PROTOCOLS = [MSG_BOUNDARY_PROTOCOL];
+
+/**
+ * The baseline CLI-SDK protocol. All responses in this protocol go to stdout.
+ * The CLI combines both stdout and stderr to interpret the hook response.
+ * This simplistic protocol has inherent limitation: no logging diagnostics!
+ * @param {string[]} args - Command-line arguments passed to this process.
+ * @returns {Protocol} Specified communication rules for the SDK to follow.
+ */
+export function BaseProtocol(args) {
```
WilliamBergamin, line 39: NIT: should we name this `DefaultProtocol`?

packages/cli-hooks/src/check-update.js
```diff
@@ -0,0 +1,314 @@
...
+ * Checks if the latest version is more recent than the current version.
+ * @param {string | undefined} current - Package version available in the project.
+ * @param {string | undefined} latest - Most up-to-date dependency version available.
+ * @returns {boolean} If the update will result in a breaking change.
+ */
+export function hasAvailableUpdates(current, latest) {
+  if (!current || !latest || !clean(current) || !clean(latest)) {
+    return false;
+  }
+  return gt(latest, current);
```
WilliamBergamin, line 232: Awesome 💯

packages/cli-hooks/src/protocols.js
```diff
@@ -0,0 +1,119 @@
...
+  [SUPPORTED_NAMED_PROTOCOLS[0]]: MessageBoundaryProtocol,
+};
+
+/**
+ * Determines the protocol interface to use when communicating with the CLI.
+ * Based on the arguments provided by the CLI to the SDK hook process.
+ * @param {string[]} args - Command-line flags and arguments passed to the hook.
+ * @returns {Protocol} The interface to follow when messaging to the CLI.
+ */
+export function getProtocolInterface(args) {
```
WilliamBergamin, line 92: Nit: could we name this `getProtocol` since this function returns an implementation of a protocol

WilliamBergamin (commented):
This is awesome 💯 just left some nits take them or leave them

packages/cli-hooks/src/check-update.js
zimeg: I hoped so but think it might be necessary to run this script as a standalone script 🤔 This is what happens when I try removing it for `get-hooks` in a sample app:

```sh
$ hermes manifest
Check /Users/ethan.zimbelman/.slack/logs/slack-debug-20240126.log for full error logs

🚫 There was an error while reading the Slack Configuration file (`slack.json`) or running the `get-hooks` hook (sdk_config_load_error)

Error Details:

1: Command for '' returned an error: exit status 2
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 1: import: command not found
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 2: import: command not found
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 4: /Applications: is a directory
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 5: LICENSE: command not found
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 6: LICENSE: command not found
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 7: node_modules/: is a directory
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 9: syntax error near unexpected token `process.argv[1]'
/private/tmp/sample_app/node_modules/.bin/slack-cli-get-hooks: line 9: `if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {' (sdk_hook_invocation_failed)
Suggestion: Run `slack doctor` to check that your system dependencies are up-to-date.
```

Confusingly, both of the following commands work fine without `#!/usr/bin/env node`:

```sh
$ node ./node_modules/.bin/slack-cli-get-hooks
$ sh -c "node ./node_modules/.bin/slack-cli-get-hooks"
```

This kinda makes me think something strange might be happening with how the CLI is calling these.

I also just noticed this command returns the same `import: command not found`+ errors:

```sh
$ npx slack-cli-get-hooks
```

I think that's a solid hint, but I'm not sure if we want to call the hooks another way? I'll poke into this a bit more soon!

packages/cli-hooks/src/get-manifest.js
zimeg: Thanks! And that formatted error would be so much cleaner, good catch! I'll look into this soon but will add it to our backlog for now 🗄️

packages/cli-hooks/src/protocols.js
zimeg: Yes! JSDoc has surprised me in such a great way! 🤩

Though I'm now not so sure the `.log` part is strictly honored. It seems like any of these console loggers meets the logging interface so passes type checks (which is solid enough IMO). This is what the `log` type is showing for me 😄 

```
(property) log: {
    (...data: any[]): void;
    (message?: any, ...optionalParams: any[]): void;
    (message?: any, ...optionalParams: any[]): void;
}
─────────────────────────────────────────────────────────────────────────
Logging utility to surface diagnostic info from the SDK or username code.
```

packages/cli-hooks/src/check-update.js
zimeg: Ah but of course, this is the command we use in the `slack.json`:

```json
{
  "hooks": {
    "get-hooks": "npx -q --no-install -p @slack/cli-hooks slack-cli-get-hooks"
  }
}
```

I'm leaning towards keeping this command in the as is in the `slack.json` since `npx` takes care of the symlink magic, but substituting it for calling `node` might remove the need for `#!` 🤔

packages/cli-hooks/src/get-hooks.js
zimeg: Totally in favor of using `SUPPORTED_NAMED_PROTOCOLS` - good call. Making the update now to use that exported array with only `message-boundaries` for now. Open to also including `default` in the export, but I lean towards including just the `message-boundaries` in this hook.

packages/cli-hooks/src/check-update.js
filmaj: All good, I'm surprised by the behaviour too. Oh well. Not a big deal.

packages/cli-hooks/src/protocols.js
filmaj: That looks correct, those are overloads for the `log` method in particular.

packages/cli-hooks/src/start.js
zimeg: This might be a bit verbose, but I think this might be more useful. I still remain open to any suggestions!

```sh
$ node ~/programming/tools/node-slack-sdk/packages/cli-hooks/src/start.js
file:///Users/ethan.zimbelman/programming/tools/node-slack-sdk/packages/cli-hooks/src/start.js:66
    throw new Error(missingBotTokenError);
          ^

Error: Missing the bot token needed to start the app with Socket Mode.
Hints: Setting the SLACK_CLI_XOXB environment variable is required.
Check: Confirm that you are using the latest version of the Slack CLI.
    at validateEnvironment (file:///Users/ethan.zimbelman/programming/tools/node-slack-sdk/packages/cli-hooks/src/start.js:66:11)
    at start (file:///Users/ethan.zimbelman/programming/tools/node-slack-sdk/packages/cli-hooks/src/start.js:22:3)
    at file:///Users/ethan.zimbelman/programming/tools/node-slack-sdk/packages/cli-hooks/src/start.js:14:3
    at ModuleJob.run (node:internal/modules/esm/module_job:217:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)
```

packages/cli-hooks/src/protocols.js
zimeg: Certainly a more correct and clear name! Will make the update now :raised_hands:

packages/cli-hooks/src/check-update.js
zimeg: Thanks! It's so much cleaner. Appreciate the suggestion 🙌

packages/cli-hooks/src/start.js
filmaj: Yeah, awesome, much better. Thanks!

packages/cli-hooks/src/get-hooks.js
zimeg: Just checked and this is supported with both `v18.19.0` and `v20.8.0` 🚀

zimeg:
@filmaj @WilliamBergamin appreciate another round of reviews and pointers on polish! I think I've addressed all of the comments and suggestions (or added more polish-related CLI-things to our backlog) so I think this is ready for another round of review.

I've tested this on Node v18.19.0 and v20.8.0 for macOS and Node v18.19.0 and v20.11.0 for Windows. Everything is working as hoped! There is a problem forming the actual web socket connection on Windows, but I think this is a problem with the virtual environment I'm using and not the hooks (the app is in the "authenticating socket mode connection" state!) 😬

filmaj (approved):
:shipit:

zimeg:
@filmaj :excited: 🚀  let's goooooo!! Do we want to start this at `v1.0.0` or stay in the patch range for now? Or maybe that's extreme and we want something more minor?

filmaj:
Your call big guy!

zimeg:
@filmaj The time feels right for a 1.0.0 😎

zimeg:
@filmaj @WilliamBergamin @misscoded Really appreciate all of the reviews on this! Thanks a ton! 🚢 💨

PR #1718, opened 2024-01-09
Bump axios to 1.6.5 (another security vuln), publish `@slack/web-api` v6.11.2

filmaj:


rafael-fecha:
@filmaj we need also to this for @slack/oauth. there is still using old @slack/web-api which uses old axios
https://github.com/slackapi/node-slack-sdk/blob/main/packages/oauth/package.json#L43

PR #1725, opened 2024-01-16
Drop end year on LICENSE dates, add licenses to each sub-package

filmaj:
@zimeg astutely noted that we don't ship the LICENSE when we package for an npm release. This PR addresses that by simply copying the LICENSE into each sub-package dir. Also drops the end-year on the copyright (only start date is needed, end date is implied).


zimeg (approved):
📜 🖊️ LGTM!

PR #1727, opened 2024-01-18
Prep for new major version of oauth library

filmaj:
- Bump minimum required node version to 18 (warrants a major version bump)
- Update unit test run script to print out project coverage
- Make sure `npm test` runs the build to ensure there are no typescript compilation issues
- Bump various dependencies to latest versions (including web-api to latest major version, which requires one small update to the `auth.test` API call)
- Drop a few unused dev dependencies (`nop`, `superagent`, and `uncaughtException`)


hello-ashleyintech (approved):
🚀

PR #1730, opened 2024-01-18
lint: enforce max line length / max items per line / one item per line for imports using import-newlines eslint plugin

filmaj:
A majority of the changes in here are style changes introduced by enforcing import line styles using `eslint-plugin-import-newlines`.

For reviewers: just need to check changes to `.eslintrc.js`. You can see the impact this change has on the `import` style in the rest of the changes.


zimeg (approved):
L
G
T
M
😄

PR #1732, opened 2024-01-19
socket-mode: prep for major release

filmaj:
- bump minimum node version to 18
- remove unused p-* dependencies
- add a `.nycrc.json` file as other packages in this repo use, for holding code coverage configuration
- code changes to move to new major version of `ws` and `web-api` dependencies
  - I went through [this breaking changes list for `ws`](https://github.com/websockets/ws/releases/tag/8.0.0) and made the updates in here based on that. In summary, raw string buffers are passed to handler methods now (previously, `ws` would decode the string on your behalf automatically). Presumably to support binary payloads.
  - Related: do we need to worry about binary payloads being provided over websocket messages?

### DIFF
--- a/packages/socket-mode/package.json
+++ b/packages/socket-mode/package.json
@@ -41,7 +41,8 @@
     "build:clean": "shx rm -rf ./dist ./coverage ./.nyc_output",
     "lint": "eslint --ext .ts src",
     "mocha": "mocha --config .mocharc.json src/*.spec.js",
-    "test": "npm run lint && npm run build && nyc --reporter=text-summary npm run mocha",
+    "test:integration": "mocha --config .mocharc.json --timeout 10000 test/integration.spec.js",
+    "test": "npm run lint && npm run build && nyc --reporter=text-summary npm run mocha && npm run test:integration",
     "watch": "npx nodemon --watch 'src' --ext 'ts' --exec npm run build"
   },
   "dependencies": {
--- a/packages/socket-mode/package.json
+++ b/packages/socket-mode/package.json
@@ -24,8 +24,8 @@
     "dist/**/*"
   ],
   "engines": {
-    "node": ">=12.13.0",
-    "npm": ">=6.12.0"
+    "node": ">= 18",
+    "npm": ">= 8.6.0"
   },
   "repository": "slackapi/node-slack-sdk",
   "homepage": "https://slack.dev/node-slack-sdk/socket-mode",
@@ -40,40 +40,39 @@
     "build": "npm run build:clean && tsc",
     "build:clean": "shx rm -rf ./dist ./coverage ./.nyc_output",
     "lint": "eslint --ext .ts src",
-    "test": "npm run lint && npm run build && nyc mocha --config .mocharc.json src/*.spec.js",
+    "mocha": "mocha --config .mocharc.json src/*.spec.js",
+    "test": "npm run lint && npm run build && nyc --reporter=text-summary npm run mocha",
     "watch": "npx nodemon --watch 'src' --ext 'ts' --exec npm run build"
   },
   "dependencies": {
-    "@slack/logger": "^3.0.0",
-    "@slack/web-api": "^6.11.2",
-    "@types/node": ">=12.0.0",
-    "@types/p-queue": "^2.3.2",
-    "@types/ws": "^7.4.7",
-    "eventemitter3": "^3.1.0",
+    "@slack/logger": "^4",
+    "@slack/web-api": "^7.0.1",
+    "@types/node": ">=18",
+    "@types/ws": "^8",
+    "eventemitter3": "^5",
     "finity": "^0.5.4",
-    "p-cancelable": "^1.1.0",
-    "p-queue": "^2.4.2",
-    "ws": "^7.5.3"
+    "ws": "^8"
   },
   "devDependencies": {
-    "@types/chai": "^4.3.5",
-    "@types/mocha": "^10.0.1",
-    "@typescript-eslint/eslint-plugin": "^6.4.1",
-    "@typescript-eslint/parser": "^6.4.0",
-    "chai": "^4.3.8",
-    "eslint": "^8.47.0",
-    "eslint-config-airbnb-base": "^15.0.0",
-    "eslint-config-airbnb-typescript": "^17.1.0",
-    "eslint-plugin-import": "^2.28.1",
+    "@types/chai": "^4",
+    "@types/mocha": "^10",
+    "@types/sinon": "^17",
+    "@typescript-eslint/eslint-plugin": "^6",
+    "@typescript-eslint/parser": "^6",
+    "chai": "^4",
+    "eslint": "^8",
+    "eslint-config-airbnb-base": "^15",
+    "eslint-config-airbnb-typescript": "^17",
+    "eslint-plugin-import": "^2",
     "eslint-plugin-import-newlines": "^1.3.4",
-    "eslint-plugin-jsdoc": "^46.5.0",
-    "eslint-plugin-node": "^11.1.0",
-    "mocha": "^10.2.0",
-    "nyc": "^15.1.0",
+    "eslint-plugin-jsdoc": "^48",
+    "eslint-plugin-node": "^11",
+    "mocha": "^10",
+    "nyc": "^15",
     "shx": "^0.3.2",
-    "sinon": "^15.2.0",
+    "sinon": "^17",
     "source-map-support": "^0.5.21",
-    "ts-node": "^10.8.1",
-    "typescript": "^4.1.0"
+    "ts-node": "^10",
+    "typescript": "5.3.3"
   }
 }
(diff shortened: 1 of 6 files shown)


seratch:
@filmaj 
>Related: do we need to worry about binary payloads being provided over websocket messages?

A Socket Mode client never receives binary payloads. You can safely assume that WS message payloads are always JSON format text data.

packages/socket-mode/src/SocketModeClient.ts
```diff
@@ -694,8 +694,12 @@ export class SocketModeClient extends EventEmitter {
    * `onmessage` handler for the client's WebSocket.
    * This will parse the payload and dispatch the relevant events for each incoming message.
    */
-  protected async onWebSocketMessage({ data }: { data: string }): Promise<void> {
-    this.logger.debug(`Received a message on the WebSocket: ${data}`);
+  protected async onWebSocketMessage(data: WebSocket.RawData, isBinary: boolean): Promise<void> {
```
seratch, line 697: Is it safe enough to assume that this data cannot be undefined | null? If we're unsure about it, we may want to check the value existence in the beginning. of this internal method.

seratch (approved):
Looks good to me. Left a minor comment about data value existence. If we don't need to worry about it, please ignore it.

packages/socket-mode/package.json
```diff
@@ -40,40 +40,39 @@
...
-    "@types/ws": "^7.4.7",
-    "eventemitter3": "^3.1.0",
+    "@slack/logger": "^4",
+    "@slack/web-api": "^7.0.1",
+    "@types/node": ">=18",
+    "@types/ws": "^8",
+    "eventemitter3": "^5",
     "finity": "^0.5.4",
-    "p-cancelable": "^1.1.0",
-    "p-queue": "^2.4.2",
```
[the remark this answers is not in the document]
filmaj: Yes! I've been reviewing the dependencies as I've been working on new major releases and these were unused 🤷

packages/socket-mode/src/SocketModeClient.ts
filmaj: Maybe it is worth adding some tests to answer this question? The coverage numbers for this package are poor (30%). To be honest I'm nervous making this web socket dependency major version upgrade without a basic integration test with a websocket peer.

I did test these changes manually with a real app and a few events were handled fine, but having more tests would give us more confidence to make changes in this project.

seratch: Adding tests is great! but just making the code robust enough to deal with any patterns (meaning RawData | undefined | null) would sound enough for me.

filmaj: @seratch I added a couple of basic integration tests, if you could please have a look.
I think those uncovered a bug, or at least a documentation discrepancy... I had to introduce an additional sleep or wait for the `connected` event to be emitted to be able to `start()` and then `disconnect()` the socket mode client gracefully (see the test code for details). If I omit this and just call:

```
await client.start();
await client.disconnect();
```

.. I get the following error and output:

```
➜ npm run test:integration

> @slack/socket-mode@1.3.3 test:integration
> mocha --config .mocharc.json --timeout 10000 test/integration.spec.js



  Integration tests with a WebSocket server
    ✔ connects to a server
    1) connects to a server


  1 passing (35ms)
  1 failing

  1) Integration tests with a WebSocket server
       connects to a server:
     Uncaught Error: WebSocket was closed before the connection was established
```

I also think the README is slightly off with respect to the events that are emitted. I tried listening to the `ready` event, which we [document here](https://github.com/slackapi/node-slack-sdk/blob/main/packages/socket-mode/README.md#lifecycle-events) instead of the `connected` event, but it never gets raised. I believe this is because [the only events that are emitted are the ones on the "main" state machine in this project](https://github.com/slackapi/node-slack-sdk/blob/main/packages/socket-mode/src/SocketModeClient.ts#L249-L259), whereas [the `ready` event is part of the "connected" sub-machine](https://github.com/slackapi/node-slack-sdk/blob/main/packages/socket-mode/src/SocketModeClient.ts#L240) (and state transitions in this sub-machine are not emitted).

Let me know if you have thoughts on that: we could either remove the `ready` event from the docs, or we can emit events from the sub-machine(s) as well.

Since this PR is part of upcoming major release, we can introduce backwards incompatible changes at this time.

filmaj: @seratch one more thing: I added a `describe.skip` block of a couple of failure tests: sending a `null` WS message and sending an empty string WS message.

The null message gets treated as binary, is logged out as an ERROR, and is essentially ignored in the implementation in this PR.

The empty string message raises a JSON parsing exception (as is expected), which is logged as an ERROR, but is not catastrophic to the client (the client continues to function).

If you have thoughts on these, let me know. If you don't think they are useful, we can remove them.

seratch: @filmaj Thanks for the tests! 

>The ready state topic

I remember that when I improved this module's internals a lot to make it stable, I had to move the ready state to the "connected" sub state machine. Thus, I agree that the documents are now outdated. Could you remove the "ready" state from the docs? I am confident that no one has been relying on it (because it never worked) but we can quickly mention it in the migration guide as necessary.

> the null / empty string tests

Thanks for verifying this! The only slight fear I still have is that the possibility to receive invalid text message from Slack servers due to its incidents/outage/misbehaviors (since the WS server is not a simple hosted OSS code and it has its own custom logic such as disconnection warnings etc., there is a certain risk to cause unexpected data delivery). Writing lots of error logs of JSON parsing error could be a false monitoring issue on the customer side (they may be PDed due to that). Thus, I think ignoring these invalid partterns without raising errors coluld be beneficial for SDK users. But if it takes your time a lot, we can do this later on. It's not a breaking change but an internal improvements. We can do it at any time!

packages/socket-mode/test/integration.spec.js
```diff
@@ -0,0 +1,101 @@
+const { assert } = require('chai');
+const { SocketModeClient, LogLevel } = require('../.');
+const { WebSocketServer} = require('ws');
+const { createServer } = require('http');
+
+const HTTP_PORT = 8080;
```
seratch, line 6: We may want to avoid 8080 port for testing on our local machines because Zscaler and other similar products could use this port...

seratch (approved):
Thank you! Left a few minor comments.

packages/socket-mode/src/SocketModeClient.ts
filmaj: > Writing lots of error logs of JSON parsing error could be a false monitoring issue on the customer side (they may be PDed due to that). Thus, I think ignoring these invalid partterns without raising errors coluld be beneficial for SDK users.

Thanks for your input! In this case I will set any logs for these results as DEBUG level, and see if I can modify the tests to verify this behaviour.

I will also add a few basic tests to verify the various event names/states we document are being emitted, and remove the `ready` one.

packages/socket-mode/src/SocketModeClient.ts
```diff
@@ -134,21 +134,21 @@ export class SocketModeClient extends EventEmitter {
...
+   * or to disconnect the client via the `disconnect` method.
    */
   public start(): Promise<AppsConnectionsOpenResponse> {
     this.logger.debug('Starting a Socket Mode client ...');
     // Delegate behavior to state machine
     this.stateMachine.handle(Event.Start);
     // Return a promise that resolves with the connection information
     return new Promise((resolve, reject) => {
-      this.once(ConnectingState.Authenticated, (result) => {
+      this.once(State.Connected, (result) => {
```
filmaj, line 146: @seratch This is a change I just introduced that fixes the issue I mentioned. Now consumers of this package can run:

```
await client.start();
await client.disconnect();
```

.. without issue. Previously, since `start()` completed while the client was still connecting - that is, the websocket connection was not established yet - the above code would cause a "socket already closed" exception to be raised.

However, I understand this may be a significant behaviour change, though it feels like a positive one to me. What do you think?

seratch: @filmaj This looks a nice improvement. Indeed, the behavior could be changed but it's favorable to almost 100% users. The only downside I can think of would be a slightly longer bootstrapping time (perhaps, 1-2 seconds?) and it may make deployment time longer accordingly. However, it's not 10+ seconds or minutes, so it should be totally fine for anyone. 

If you want to be as careful as possible about the behavior change, one possible solution would be to provide an opt-out option such as asyncStart: true (default value can be false, which means await for Connected state; true means waiting for Authenticated) for existing users. I don't think we should provide such, though.

seratch (approved):
Thank you; Looks good to me

packages/socket-mode/src/SocketModeClient.ts
filmaj: I think I will release as-is. People can always use the previous version if this behaviour is a show-stopper.

filmaj:
Thank you so much for your patience and consistent reviews Kaz 🙇

PR #1737, opened 2024-01-26
web-api: prevent apps.event.authorizations.list API from ever sending token in the body

### LINKED ISSUE #1498: token parameter in apps.event.authorizations.list method call results in invalid_auth error
If you pass an app-level token to `WebClient`'s `apps.event.authorizations.list` method call, it results in invalid_auth error code. A workaround is to pass the token to `WebClient` constructor instead.

This fails with invalid_auth error:
```javascript
const result = await web.apps.event.authorizations.list({
  token: 'xapp-xxx',
  event_context: 'xxx'
});
```

This following works:
```javascript
const result = await web.apps.event.authorizations.list({
  event_context: 'xxx'
});
```

This behavior should be improved.

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

#### The Slack SDK version

any

#### Node.js runtime version

any

#### OS info

any

filmaj:
Sending `token` in the body for the `apps.event.authorizations.list` API will cause the API to return an error ([see the last sentence in this doc](https://api.slack.com/methods/apps.event.authorizations.list#how)).

This fixes #1498.

Additionally, this PR compiles all headers when logging them out in debug mode - even the ones axios sets by default. This critically also includes the token if the token is passed in to WebClient constructor. I made this change and included it here as otherwise determining the full list of headers sent by web-api was challenging. Optionally, this logging feature can be removed if people have opinions about it. Previously, the debug output would look like this:

```
[DEBUG]  web-api:WebClient:0 apiCall('apps.event.authorizations.list') start
[DEBUG]  web-api:WebClient:0 http request url: https://slack.com/api/apps.event.authorizations.list
[DEBUG]  web-api:WebClient:0 http request body: {"event_context":"4-eyJldCI6Im1lc3NhZ2UiLCJ0aWQiOiJUMDI5VjY0NjhSTCIsImFpZCI6IkEwNjAxUkVRTDkzIiwiY2lkIjoiQzAyOVlUNUtFTUIifQ"}
[DEBUG]  web-api:WebClient:0 http request headers: {}
```

Note the request headers are stated to be empty (when passing a token into the `WebClient` constructor). Impossible, as the token needs to be provided SOMEHOW!

With the logging changes introduced in this PR, the same call as above now yields:

```
[DEBUG]  web-api:WebClient:0 apiCall('apps.event.authorizations.list') start
[DEBUG]  web-api:WebClient:0 http request url: https://slack.com/api/apps.event.authorizations.list
[DEBUG]  web-api:WebClient:0 http request body: {"event_context":"4-eyJldCI6Im1lc3NhZ2UiLCJ0aWQiOiJUMDI5VjY0NjhSTCIsImFpZCI6IkEwNjAxUkVRTDkzIiwiY2lkIjoiQzAyOVlUNUtFTUIifQ"}
[DEBUG]  web-api:WebClient:0 http request headers: {"User-Agent":"@slack:web-api/7.0.1 node/18.15.0 darwin/23.2.0","Authorization":"[[REDACTED]]","Accept":"application/json, text/plain, */*"}
```

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -9,7 +9,7 @@ import { TextDecoder } from 'util';
 import isStream from 'is-stream';
 import pQueue from 'p-queue';
 import pRetry, { AbortError } from 'p-retry';
-import axios, { AxiosInstance, AxiosResponse } from 'axios';
+import axios, { AxiosHeaderValue, AxiosInstance, AxiosResponse } from 'axios';
 // eslint-disable-next-line @typescript-eslint/naming-convention
 import FormData from 'form-data';
 import isElectron from 'is-electron';
@@ -48,7 +48,8 @@ import {
 /*
  * Helpers
  */
-
+// Props on axios default headers object to ignore when retrieving full list of actual headers sent in any HTTP requests
+const axiosHeaderPropsToIgnore = ['delete', 'common', 'get', 'put', 'head', 'post', 'link', 'patch', 'purge', 'unlink', 'options'];
 const defaultFilename = 'Untitled';
 const defaultPageSize = 200;
 const noopPageReducer: PageReducer = () => undefined;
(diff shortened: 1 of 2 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -547,6 +548,28 @@ export class WebClient extends Methods {
         if (url.endsWith('admin.analytics.getFile')) {
           config.responseType = 'arraybuffer';
         }
+        // apps.event.authorizations.list will reject HTTP requests that send token in the body
```
seratch, line 551: If this change does not break anything, we may want to apply this to all API methods. But if we make the change as small as possible, going with if condition this time sounds good to me too.

seratch (approved):
Thanks for fixing this!

packages/web-api/src/WebClient.ts
filmaj: I will add a comment here to note your point, but I would rather initially keep the scope of this change just to this one method; I am afraid that expanding to all methods may have unintended consequences.

PR #1757, opened 2024-03-15
socket-mode: add more debug logging to low level websocket event handlers

filmaj:
also differentiate between primary and secondary websocket events in logs.

I have a suspicion that sometimes, both primary and secondary websockets could be briefly active and overlapping, however both websockets emit events into the SINGLE state machine configuration - I wonder if that could lead to unexpected state transitions (two websockets each with their own independent events being emitted into a single state machine).

NOTE: this change probably should cherry-picked into a new `socket-mode-1.3.x` maintenance branch, and a 1.3.4 patch release could be cut from that.


zimeg (approved):
So much state! This'll be super helpful for debugging :raised_hands:

PR #1772, opened 2024-04-08
web-api(fix): Update p-retry package

cotsupa:
###  Summary

When using web-api, @types/retry may conflict and build may not be possible, so I updated the p-retry package to v6.
```bash
node_modules/@slack/web-api/dist/retry-policies.d.ts:5:39 - error TS2312: An interface can only extend an object type or intersection of object types with statically known members.

5 export interface RetryOptions extends OperationOptions {
                                        ~~~~~~~~~~~~~~~~


Found 1 error in node_modules/@slack/web-api/dist/retry-policies.d.ts:5
```

### DIFF
--- a/packages/web-api/package.json
+++ b/packages/web-api/package.json
@@ -55,7 +55,7 @@
     "is-electron": "2.2.2",
     "is-stream": "^2",
     "p-queue": "^6",
-    "p-retry": "^4",
+    "p-retry": "^6",
     "retry": "^0.13.1"
   },
   "devDependencies": {
--- a/packages/web-api/package.json
+++ b/packages/web-api/package.json
@@ -49,6 +49,7 @@
     "@slack/logger": "^4.0.0",
     "@slack/types": "^2.9.0",
     "@types/node": ">=18.0.0",
+    "@types/retry": "0.12.0",
     "axios": "^1.6.5",
     "eventemitter3": "^5.0.1",
     "form-data": "^4.0.0",
--- a/packages/web-api/package.json
+++ b/packages/web-api/package.json
@@ -49,7 +49,7 @@
     "@slack/logger": "^4.0.0",
     "@slack/types": "^2.9.0",
     "@types/node": ">=18.0.0",
-    "@types/retry": "0.12.0",
+    "@types/retry": "^0.12.0",
     "axios": "^1.6.5",
     "eventemitter3": "^5.0.1",
     "form-data": "^4.0.0",
--- a/packages/web-api/package.json
+++ b/packages/web-api/package.json
@@ -55,7 +55,7 @@
     "is-electron": "2.2.2",
     "is-stream": "^2",
     "p-queue": "^6",
-    "p-retry": "^6",
+    "p-retry": "^4",
     "retry": "^0.13.1"
   },
   "devDependencies": {
--- a/packages/web-api/package.json
+++ b/packages/web-api/package.json
@@ -49,7 +49,7 @@
     "@slack/logger": "^4.0.0",
     "@slack/types": "^2.9.0",
     "@types/node": ">=18.0.0",
-    "@types/retry": "^0.12.0",
+    "@types/retry": "0.12.0",
     "axios": "^1.6.5",
     "eventemitter3": "^5.0.1",
     "form-data": "^4.0.0",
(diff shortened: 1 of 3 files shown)


filmaj:
Thanks for the PR! I as well have tried to update this package in the future, but in v6 (and many of `sindresorhus`' packages), he has moved to an ESM-only approach. You can see in the GitHub Actions run that this fails the test. Not sure if there is a way to be able to do this - perhaps a build configuration tweak would resolve this? Not sure, but if you can figure out a fix for this I would be very appreciative!

cotsupa:
@filmaj 
Thank you for your comment!
When I looked into p-retry, it turned out to be a pure ESM, and it was recommended to use ESM. It's difficult to convert the web-api to ESM, isn't it?
https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

filmaj:
I'm actually not fully understanding how the dependencies work in this case. It seems like we are using this `@types` repo implicitly, even though it is not in the `package.json` of this project? https://www.npmjs.com/package/@types/retry

Feel free to try whatever changes you think would work and pass the tests; I am curious to see how one could resolve the issues. I recall spending an hour or two trying to fix this stuff a few months ago and failed to figure it out.

cotsupa:
This issue may be resolved without upgrading p-retry. I would like to ask you to merge this commit. I think it would be better to tackle the upgrade of p-retry separately from this issue.

filmaj:
@cotsupa can you explain where the `OperationOptions` type is coming from? I would like to understand the import graph better. Is my suspicion that somehow our IDEs are pulling in the `@types/retry` repo in implicitly correct?

cotsupa:
@filmaj 
It reproduces when the local-package dependencies are as follows:
```bash
$ npm ls @types/retry
@local-package@1.0.0 /Path/local-package
├─┬ @hoge-package@1.0.0
│ └─┬ @types/async-retry@1.4.8
│   └── @types/retry@0.12.5
└─┬ @slack/web-api@7.0.2
  └─┬ p-retry@4.6.2
    └── @types/retry@0.12.0
```

In @types/retry version 0.12.5, OperationOptions is defined as follows:
```js
export type OperationOptions = WrapOptions | number[];
```

Accepting this commit or writing "@types/retry": "0.12.0" in your @slack/web-api dependencies may resolve this issue. I think it's better to accept this commit.

filmaj:
Yes I would feel better about explicitly depending upon the relevant types package instead of assuming it would come in via a transient dependency. Another concern I have is that the latest `@types/retry` package is v0.12 while the package with the ACTUAL retry policies is v0.13. We would have to assume the types provided by the `@types/retry` still match the implied types used by the `retry` package.

From the library perspective it comes in via the `p-retry` package for the currently-used version (4.x):

```
11:22:54 in node-slack-sdk/packages/web-api
➜ npm ls @types/retry
@slack/web-api@7.0.2 /Users/fmaj/src/node-slack-sdk/packages/web-api
└─┬ p-retry@4.6.2
  └── @types/retry@0.12.0
```

cotsupa:
Thank you for your acceptance. I have committed the fix to add it to the dependencies, so please check it.
Indeed, there are concerns, but since we are dependent on p-retry, I think it's fine as it is.

cotsupa:
@filmaj Please confirm this commit.

packages/web-api/package.json
```diff
@@ -49,6 +49,7 @@
     "@slack/logger": "^4.0.0",
     "@slack/types": "^2.9.0",
     "@types/node": ">=18.0.0",
+    "@types/retry": "0.12.0",
```
filmaj, line 52:
```suggestion
    "@types/retry": "^0.12.0",
```

filmaj (changes requested):
I left one small tweak suggestion, but LGTM otherwise.

filmaj:
Lol I see why you locked to a specific version of the types now 🤦 if you can revert that last commit I swear we will get this merge. My apologies

cotsupa:
@filmaj Thanks for the comment. Please check again!

filmaj (approved):

filmaj:
Released on npm as 7.0.3! Thanks so much. Full release is here: https://github.com/slackapi/node-slack-sdk/releases/tag/%40slack%2Fweb-api%407.0.3

PR #1782, opened 2024-04-25
Docs on verifyStateParam for server-side verification 

sidpremkumar:
###  Summary

Pulling my hair out trying to figure out why `verifyStateParam` wasn't being called on the server. Did not find any issues or docs around this so updating the readme so others don't go through what I want through 😭

### DIFF
--- a/docs/_packages/oauth.md
+++ b/docs/_packages/oauth.md
@@ -366,9 +366,11 @@ const installer = new InstallProvider({
   clientId: process.env.SLACK_CLIENT_ID,
   clientSecret: process.env.SLACK_CLIENT_SECRET,
   stateStore: {
-    // generateStateParam's first argument is the entire InstallUrlOptions object which was passed into generateInstallUrl method
-    // the second argument is a date object
-    // the method is expected to return a string representing the state
+    /**
+     * generateStateParam's first argument is the entire InstallUrlOptions object which was passed into generateInstallUrl method
+     * the second argument is a date object
+     * the method is expected to return a string representing the state
+     */
     generateStateParam: (installUrlOptions, date) => {
       // generate a random string to use as state in the URL
       const randomState = randomStringGenerator();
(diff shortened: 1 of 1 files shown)


docs/_packages/oauth.md
```diff
@@ -377,13 +379,16 @@ const installer = new InstallProvider({
       // return a state string that references saved options in DB
       return randomState;
     },
-    // verifyStateParam's first argument is a date object and the second argument is a string representing the state
-    // verifyStateParam is expected to return an object representing installUrlOptions
-    verifyStateParam:  (date, state) => {
+    /**
+     * verifyStateParam's first argument is a date object and the second argument is a string representing the state
+     * verifyStateParam is expected to return an object representing installUrlOptions
+     * @note You must set `legacyStateVerification: true` if you want `verifyStateParam` to be called on the server side
```
sidpremkumar, line 385: This is the new line here.

seratch: I don't think this is correct. As you can see [here](https://github.com/slackapi/node-slack-sdk/blob/%40slack/oauth%403.0.0/packages/oauth/src/install-provider.ts#L509-L526), the only necessary options are stateVerification and stateStore.

sidpremkumar: The cookie will never get set if you are writing a custom integration (at least that was my experience). 

Pls correct me if I was doing something wrong though 🙇 , I checked the cookies in chrome and they never never set. 

This is my setup 

1) Get a url with metadata attached: 
```js 
await SlackInstallProvider.generateInstallUrl({
          scopes: ["chat:write"],
          metadata: uniqueId,
        }),
```

2) Pass that to the frontend and use it in a button 

3) Use in redirect

```js
slackRoute.get(
  "/redirect",
  asyncHandler(async (request, response) => {
    try {
      const callbackOptions: CallbackOptions = {
        successAsync: async (installation, options, req, res) => {
          if (!installation.metadata) {
            /**
             * Something went wrong
             */
            logger.error(`Got Slack redirect without metadata!`);
            response.redirect(`${BASE_URL}/home/settings?error=true`);
            return;
          }

          if (!installation.bot?.token) {
            logger.error(`Got Slack redirect without bot token!`);
            response.redirect(`${BASE_URL}/home/settings?error=true`);
            return;
          }

          /**
           * Epic lets save out data
           */
          await SlackIntegrationDB.createOrUpdateSlackIntegration({
            ...custom internal logic
          });

          /**
           * Finally redirect them back to settings
           */
          // @ts-ignore
          res.redirect(`${BASE_URL}/home/settings`);
        },
        failure: (error, installOptions, req, res) => {
          logger.error(`Something went wrong when installing slack`, error);
          // @ts-ignore
          res.redirect(`${BASE_URL}/home/settings?error=true`);
        },
      };
      await SlackInstallProvider.handleCallback(
        request,
        response,
        callbackOptions
      );
    } catch (err) {
      logger.error(`Unhandled error when authing slack`, err);
      response.redirect(`${BASE_URL}/home/settings?error=true`);
    }
  })
);
```


See my browser:
2024-04-24 at 9 22 PM



Let me know if I can provide anything else! Thanks for responding here :)

seratch: The generateInstallUrl method returns only the URL, and this is why the cookies are not set. You need to use handleInstallPath method instead. Would you mind closing this PR?

sidpremkumar: Awesome thanks for the feedback here. Closing

Issue #1751, opened 2024-02-23
Way to disable message content being logged


gaurav512:
Hi,

I am looking for a way to never output / always hide the message contents from the application logs. If such way already exists, I would like to know how.

### Use-case
I am building an app which contains sensitive content in the messages that I do not want to be logged in the application logs. 

My LogLevel is '**INFO**'.
I recently saw some API failure logs in my application, which in turn resulted in the message data being logged in the app logs.

```
Error: A request error occurred: read ECONNRESET
    at requestErrorWithOriginal (/app/node_modules/@slack/web-api/dist/errors.js:33:33)
    at /app/node_modules/@slack/web-api/dist/WebClient.js:433:65
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async run (/app/node_modules/@slack/web-api/node_modules/p-queue/dist/index.js:163:29) {
  code: 'slack_webapi_request_error',
  original: AxiosError: read ECONNRESET
      at Function.AxiosError.from (/app/node_modules/@slack/web-api/node_modules/axios/dist/node/axios.cjs:837:14)
      at ClientRequest.handleRequestError (/app/node_modules/@slack/web-api/node_modules/axios/dist/node/axios.cjs:3083:25)
      at ClientRequest.emit (node:events:513:28)
      at TLSSocket.socketErrorListener (node:_http_client:494:9)
      at TLSSocket.emit (node:events:513:28)
      at emitErrorNT (node:internal/streams/destroy:157:8)
      at emitErrorCloseNT (node:internal/streams/destroy:122:3)
      at processTicksAndRejections (node:internal/process/task_queues:83:21) {
    syscall: 'read',
    code: 'ECONNRESET',
    errno: -104,
    config: {
      transitional: [Object],
      adapter: [Array],
      transformRequest: [Array],
      transformResponse: [Array],
      timeout: 0,
      xsrfCookieName: 'XSRF-TOKEN',
      xsrfHeaderName: 'X-XSRF-TOKEN',
      maxContentLength: -1,
      maxBodyLength: -1,
      env: [Object],
      validateStatus: [Function: validateStatus],
      headers: [Object [AxiosHeaders]],
      baseURL: 'https://slack.com/api/',
      maxRedirects: 0,
      proxy: false,
      method: 'post',
      url: 'views.publish',
      data: <MY MESSAGE DATA>
	  ...
	  ...
```

#### The Slack SDK version
```
[...]

PR #1786, opened 2024-05-03
providing a way to disable message content being logged

Parama92:
###  Summary

In this PR, we aim to prevent the logging of message content when a request error occurs while attempting to invoke any of Slack's web APIs.

It solves [this issue](https://github.com/slackapi/node-slack-sdk/issues/1751).

#### The problem at a glace -
The Error type - `WebAPIRequestError`, contains a property called `original` which holds the raw error received from `axios`.
The `config` property within this `original` is a part of the error data received from `axios`.
This `config` property might in turn hold a `data` property, which is the "message" we are trying to opt-out of showing.

#### Tests conducted -
1. Error log with the opt-out config -
Error with config
Note: The `original` property is missing from the Error object.

2. Error log without adding the config - 
Error without config
Note: The `original` property is present in the Error object along with all the details of the request.

### Requirements

### DIFF
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -2,7 +2,7 @@ require('mocha');
 const fs = require('fs');
 const path = require('path');
 const { Agent } = require('https');
-const { assert } = require('chai');
+const { assert, expect } = require('chai');
 const { WebClient, buildThreadTsWarningMessage } = require('./WebClient');
 const { ErrorCode } = require('./errors');
 const { LogLevel } = require('./logger');
@@ -1609,6 +1609,46 @@ describe('WebClient', function () {
       };
       await client.filesUploadV2(withRequestFileInfoFalse);
       assert.equal(client.getFileInfo.called, false);
+    });
+  });
+
+  describe('has an option to suppress request error from Axios', () => {
+    it('the \'original\' property is attached when the option, attachOriginalToWebAPIRequestError is absent', async () => {
+      const client = new WebClient(token,  {
+        timeout: 10,
+        retryConfig: rapidRetryPolicy,
+      });
+      try {
+        await client.conversations.list()
+      } catch(err) {
+        expect(err).to.haveOwnProperty('original')
+      }
+    });
+
+    it('the \'original\' property is attached when the option, attachOriginalToWebAPIRequestError is set to true', async () => {
+      const client = new WebClient(token,  {
+        timeout: 10,
+        retryConfig: rapidRetryPolicy,
+        attachOriginalToWebAPIRequestError: true,
+      });
+      try {
+        await client.conversations.list()
+      } catch(err) {
+        expect(err).to.haveOwnProperty('original')
+      }
+    });
+
+    it('the \'original\' property is not attached when the option, attachOriginalToWebAPIRequestError is set to false', async () => {
+      const client = new WebClient(token,  {
+        timeout: 10,
+        retryConfig: rapidRetryPolicy,
+        attachOriginalToWebAPIRequestError: false,
+      });
+      try {
+        await client.conversations.list()
+      } catch(err) {
+        expect(err).not.to.haveOwnProperty('original')
+      }
     });
   });
 
--- a/packages/web-api/src/WebClient.spec.js
+++ b/packages/web-api/src/WebClient.spec.js
@@ -1621,7 +1621,9 @@ describe('WebClient', function () {
     })
 
     it('the \'original\' property is attached when the option, attachOriginalToWebAPIRequestError is absent', async () => {
-      const client = new WebClient(token);
+      const client = new WebClient(token, {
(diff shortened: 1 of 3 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -182,6 +188,7 @@ export class WebClient extends Methods {
     rejectRateLimitedCalls = false,
     headers = {},
     teamId = undefined,
+    attachOriginalToWebAPIRequestError = true,
```
filmaj, line 191: We should also add this option to the `WebClientOptions` interface definition on line 61 of this file, ideally with a JSDoc (so that developers can get some basic information about this option as they develop in their IDE). You can reuse the JSDoc you added on line 170 for this if you wish, but I wouldn't link to the repo issue. We want to document this in a way that is clear to developers without having to open a link.

filmaj (changes requested):
Thanks for the PR! I left a comment about documenting the new option in `WebClientOptions`.

We will also require a unit test for this option before we can accept this PR, please 🙇 🙏

Parama92:
Thanks for the feedback! I have addressed the comments and have re-requested a review

packages/web-api/src/WebClient.spec.js
```diff
@@ -1612,6 +1612,46 @@ describe('WebClient', function () {
     });
   });
 
+  describe('has an option to suppress request error from Axios', () => {
```
filmaj, line 1615: Thanks for adding these tests!

One thing we should do is ensure we mock out any calls to the slack.com API. Right now, since your tests are actually invoking e.g. `conversations.list()` method, this is issuing an HTTP request to slack.com/api. Not great, as we are a) unnecessarily creating traffic for Slack and b) have less control over the outcome of our tests. What if, during a test run, the Slack API goes down? Should the tests fail because of some service degradation? Probably not.

To handle this, we use the `nock` module to mock out API calls. Take a look at the nock setup and use on lines 150-175 of this file. Using `nock`, we can finely control a 'fake' response from the Slack API. This way, our tests will avoid both problems a) and b) above!

Additionally, since you have total control over the "Slack API" response using `nock`, you can not only test for the presence (and absence) of the `original` property in the test, you can even make sure that its contents are what you expect (since you control the API response).

filmaj (changes requested):
Looking great! Very close, all I ask is that we mock out Slack API responses in the tests that you added. I provided details in my comment in-line.

Thanks again for working on this!

packages/web-api/src/WebClient.spec.js
Parama92: Ah! I understand. That does make sense.

I appreciate you taking the time to review this, share your thoughts, and explain the importance of each change. I apologize for any misunderstanding or oversight in my initial attempts.

filmaj: No worries at all! Zero judgment on our side, we appreciate you taking the time to send the PR and the improvements; the least we could do is help guide you along in the journey as we try to keep the quality and testing level of the project high working together 😄

filmaj (approved):
The code looks good to me, let's just make sure the linting and tests pass. On your machine locally, from the `web-api` directory, try running `npm it` to install the dependencies as well as run the web-api tests. It seems like a few lint rules failed (spacing and other style considerations).

Parama92:
I have addressed the issues that caused the linter to fail. It is running fine on my local machine now. Do you mind running the workflow once again?

filmaj:
Hmm, seems like the tests are still failing. It looks to me like something is blocking the testing process; see [the CI output here](https://github.com/slackapi/node-slack-sdk/actions/runs/9019146242/job/24783187732?pr=1786). I can also reproduce this behaviour locally on my macbook pro using node v20.

filmaj:
I believe what is happening is that, because in the tests we return an error / a non-200 HTTP response, `web-api` will by default attempt to retry issuing the request (see [this default option](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/WebClient.ts#L178)). The default retry mechanism is "10 retries in 30 minutes" - meaning, if a test returns an error result (like the new tests you added), `web-api` will block the process in order to retry the request, blocking the process for up to 30 mins! I believe this is what is causing the tests to hang.

[As per our documentation on retries](https://slack.dev/node-slack-sdk/web-api#automatic-retries), try setting this option in the tests on WebClient to turn off retries:

```
{
  retryConfig: { retries: 0 },
}
```

filmaj:
It's also a good idea to run `npm test` locally before pushing any changes to make sure everything works as expected on your machine.

Parama92:
Hmm, makes sense. Strange that it seemed to work fine on my local system earlier.
Thanks for the heads-up though! I make the necessary changes and ran `npm test`. I am getting an unrelated error - 
Error

Hopeful that this time it will work out fine! 🤞

filmaj (approved):
Woohoo! Thank you so much for your contribution!

PR #1809, opened 2024-06-10
feat: allow using WebClient APIs without argument

### LINKED ISSUE #1769: web-api: Consider removing need to provide empty object as argument to zero-argument methods
#### Packages:

Select all that apply:

- [x] `@slack/web-api`

#### The Slack SDK version

        "@slack/web-api": "^7.0.2",

#### Node.js runtime version

```
$ node --version
v20.11.1
```

#### OS info

```
$ sw_vers && uname -v
ProductName:            macOS
ProductVersion:         13.6.3
BuildVersion:           22G436
Darwin Kernel Version 22.6.0: Tue Nov  7 21:40:08 PST 2023; root:xnu-8796.141.3.702.9~2/RELEASE_ARM64_T6000
```

#### Steps to reproduce:
Create a WebClient instance and call `.auth.test()` without argument.

```
import { WebClient } from "@slack/web-api";

const api = new WebClient(
            slackBotToken
        );

const botUserInfo: any = await api.auth.test();
```

### Expected result:

```
// Works without errors
const botUserInfo: any = await api.auth.test();
```

`.auth.test()` returns a userinfo of the bot.

### Actual result:
It requires an empty dict as an argument.

```
// Error
const botUserInfo: any = await api.auth.test();  //  error TS2554: Expected 1 arguments, but got 0.

// Works with an empty dict
const botUserInfo: any = await api.auth.test({});
```

### Requirements

I read the requirements and agreed with the rules.

davidlj95:
###  Summary

Suggest an implementation for #1769 by using `void` & union types

Introduced `OptionalArgument` for explicitness

Fixes #1769 

To pay special attention:
 - [Breaking change in empty interface APIs](https://github.com/slackapi/node-slack-sdk/pull/1809/files#r1633611320)
 - [Operation with no type test](https://github.com/slackapi/node-slack-sdk/pull/1809/files#r1633615501)
 - [Add missing type test](https://github.com/slackapi/node-slack-sdk/pull/1809/files#r1633691571)


davidlj95:
> Hey this looks good to me and thank you for the PR!
> 
> One note about #1769 is that we'll want to add this to _all_ methods that accept all optional arguments. Here's a brief search I did just based on the tests where the changes in this PR would apply:

Thanks! 😊 Oh nice then 💪 I just did a few as kind of a demo of what I'd look like. If looks good, will apply it around. Thanks for searching the places where changes are needed

btw I'm thinking about adding some utility like

```typescript
type OptionalArgument<T> = T | void
```

So the intention is more explicit and can easily be refactored around in case we need to detect in which places args are optional

> 
> Would you be able to add the same changes to all of the above methods that support purely optional arguments?

Sure!

packages/web-api/src/types/request/apps.ts
```diff
@@ -1,8 +1,9 @@
 import type { AppID, CursorPaginationEnabled, TokenOverridable, OAuthCredentials } from './common';
 import type { Manifest } from './manifest';
+import { OptionalArgument } from '../helpers';
 
 // https://api.slack.com/methods/apps.connections.open
-export interface AppsConnectionsOpenArguments { }
+export type AppsConnectionsOpenArguments = OptionalArgument<object>;
```
davidlj95, line 6: ⚠️ **BREAKING CHANGE** (technically, not pragmatically)

`OpenIDConnectUserInfoArguments` & `AppsConnectionsOpenArgument` are declared as empty interfaces. 

An empty interface allows for any non nullable value, not an object with no properties. It is well [explained in `typescript-eslint` rule `no-empty-interface`](https://typescript-eslint.io/rules/no-empty-interface/). In order to specify an empty object, [`object` can be used](https://www.typescriptlang.org/docs/handbook/2/functions.html#object). Using that then

This would be a breaking change if someone is using a non-nullable and non-object value as argument to one of those methods. Like `apps.connections.open(42)`. But that shouldn't have worked in the first place

davidlj95: Can use `{}` and ignore `ESLint` rule complaining about it if we want to be sure no breaking change happens (could be done in next major as cleanup?)

packages/web-api/src/types/request/team.ts
```diff
@@ -5,25 +5,27 @@ import {
...
   user?: string;
-}
+}>;
 // https://api.slack.com/methods/team.billing.info
-export interface TeamBillingInfoArguments extends TokenOverridable {}
+export type TeamBillingInfoArguments = OptionalArgument<TokenOverridable>;
 // https://api.slack.com/methods/team.info
-export interface TeamInfoArguments extends TokenOverridable {
+export type TeamInfoArguments = OptionalArgument<TokenOverridable & {
   /**
```
davidlj95, line 29: ☝️ This one has no type tests, but it's optional too

packages/web-api/src/types/request/apps.ts
filmaj: Huh, TIL! I think as long as the previous "expected" usage of these methods (i.e. needing to provide an empty object `{}` as a parameter for these methods that accept optional arguments) are a) enshrined in tests and b) pass, then that's all that matters to me.

packages/web-api/src/types/request/team.ts
filmaj: Can we add missing type tests while we're here?

filmaj (approved):
Thanks for the work and explanation around extending from empty interfaces! I learned something today 😄 

If there is a missing test, then if you could please add it, that would be awesome! Otherwise everything looks good to me.

packages/web-api/src/types/request/apps.ts
davidlj95: Same here tbh! Didn't know about that either, it was ESLint rule complaining that lead me through the rabbit hole.

Noice then 👍

packages/web-api/src/types/request/team.ts
davidlj95: Sure!

davidlj95: Done in f98f005fb4a0e0fc72c3cf82b64b3cbc6b1a919b

packages/web-api/test/types/methods/team.test-d.ts
```diff
@@ -1,40 +1,46 @@
...
-expectError(web.team.billing.info()); // lacking argument
 // -- happy path
 expectAssignable<Parameters<typeof web.team.billing.info>>([{}]); // all optional arguments
+expectAssignable<Parameters<typeof web.team.billing.info>>([]); // no arg is fine
+
+// team.info
+// -- sad path
+// -- happy path
+expectAssignable<Parameters<typeof web.team.info>>([{}]); // all optional arguments
+expectAssignable<Parameters<typeof web.team.info>>([]); // no arg is fine
```
davidlj95, line 28: 🆕 `team.info` API type tests

PR #1819, opened 2024-06-13
types: export helper union types

### LINKED ISSUE #1818: types: export additional union types used internally
This issue comes out of #1227 and has one example of the suggestion to export certain union types as their own standalone type in the draft PR #1228 (in [this commit specifically](https://github.com/slackapi/node-slack-sdk/pull/1228/commits/2eef4825b0852ecfe6333865495ca6e30e60768b)).

Seems like a no brainer! We should export - and reuse - different kinds of union types we use internally. This is more of a refactoring task and so could be a good first issue to pick up. [This commit](https://github.com/slackapi/node-slack-sdk/pull/1228/commits/2eef4825b0852ecfe6333865495ca6e30e60768b) shows off the proposed solution well. It is particularly applicable to the `elements` array on Blocks, which can contain a certain subset of different Block Elements.

Technically the refactor could be a patch version bump and exporting the new types would be a minor version bump. As such, this could go out in a patch or minor new release in the 2.x line of `types`.

filmaj:
###  Summary

This PR fixes #1818. It exports a bunch of helper unions that we use when authoring these types, but that we've heard from the community that they could possibly want to use themselves in their apps (see #1227).

### DIFF
--- a/packages/types/src/block-kit/blocks.ts
+++ b/packages/types/src/block-kit/blocks.ts
@@ -228,7 +228,8 @@ export type RichTextBlockElement = RichTextSection | RichTextList | RichTextQuot
  * WYSIWYG message composer, so all messages sent by end-users will have this format. Use this block to include
  * user-defined formatted text in your Block Kit payload. While it is possible to format text with `mrkdwn`,
  * `rich_text` is strongly preferred and allows greater flexibility.
- * You might encounter a `rich_text` block in a message payload, as a built-in type in workflow apps, or as output of the {@link RichTextInput}.
+ * You might encounter a `rich_text` block in a message payload, as a built-in type in workflow apps, or as output of
+ * the {@link RichTextInput}.
  * @see {@link https://api.slack.com/reference/block-kit/blocks#rich_text Rich text block reference}.
  */
 export interface RichTextBlock extends Block {
--- a/packages/types/src/block-kit/blocks.ts
+++ b/packages/types/src/block-kit/blocks.ts
@@ -1,12 +1,10 @@
 // This file contains objects documented here: https://api.slack.com/reference/block-kit/blocks
 import {
   PlainTextElement,
-  MrkdwnElement,
   TextObject,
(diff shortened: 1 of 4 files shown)


packages/types/src/block-kit/blocks.ts
```diff
@@ -181,15 +209,44 @@ export interface InputBlock extends Block {
...
+  type: 'rich_text',
+  elements: RichTextBlockElement[];
+}
+
+/**
+ * A helper union type of all Block Elements that can be used as an accessory in a {@link SectionBlock}.
+ * @see {@link https://api.slack.com/reference/block-kit/blocks#section Section block reference}.
+ */
+export type SectionBlockAccessory = Button | Checkboxes | Datepicker | ImageElement | MultiSelect | Overflow |
+RadioButtons | Select | Timepicker | WorkflowButton;
```
zimeg, line 248: Great to see these ordered and matching the docs 🙌

zimeg (approved):
Solid improvements as always! Having exposed types for rich text is super helpful, so thank you for typing this 👏

All of these changes look good to me and the JSDoc is awesome 🚀 📚  No blockers but for my own understanding, are we wanting to export types in this granular way in most cases?

seratch (approved):

filmaj:
> are we wanting to export types in this granular way in most cases?

@zimeg Great question; I don't think so, and I would personally default to keeping things internal/non-exported, at least as an initial move, because:

1. It's easier to put something out there than it is to retract; adding something new is additive while taking something away is a breaking change.
2. In terms of TS specifically, it can be harder to know and manage what is unused if we just export everything.
3. In this particular case, we specifically had a request from the community to export this particular type; see this OP in #1226:

> Moving union types to their own exported types. For example, elements supported within the ActionsBlock having a type of ActionsElements. For SectionBlock, SectionElements. Etc.

Of course always open to discussing and changing this inclination if the team feels otherwise.

PR #1821, opened 2024-06-17
feat: user id map in MigrationExchangeResponse as map

### LINKED ISSUE #1584: Type generation creates UseridMap as an interface when it is a Map/Record/object
When generating types for `packages/web-api/src/response/MigrationExchangeResponse.ts`, `UseridMap` is created as an `interface` when really it should be a `Record`.

This means when calling `migration.exchange` the data in `response.user_id_map` needs to be cast before it can used.

```
// Warning: pseudo code alert
type UserIds = Partial<Record<string, string>>;

const migrateUserIds = (userIds: string[]): UserIds => {
  const response = await client.migration.exchange({
    team_id: "WORKSPACEID",
    users: userIds.join(","),
  });
  
  if (!response.ok) return {};
  return return response.user_id_map as UserIds;  // Mmmmm
}
```

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] `@slack/types`
- [ ] I don't know

### Reproducible in:

Whatever is in main at the moment: 57dfddaa5d1442aa3f854f8b51895b0d9c04e63c

#### Steps to reproduce:

```./generate-web-api-types.sh```

### Expected result:

```
# packages/web-api/src/response/MigrationExchangeResponse.ts
// ...
export type UseridMap  = Record<string, string>  // Should probably be a `Partial` or `string | undefined`
```

### Actual result:

```
# packages/web-api/src/response/MigrationExchangeResponse.ts
// ...
export interface UseridMap {
}
```

### PR DESCRIPTION
**Pull request by davidlj95**
###  Summary

Fixes #1584

In https://github.com/slackapi/java-slack-sdk/pull/1325 , JSON samples were updated so that `quicktype` properly infers `user_id_map` property type of `MigrationExchangeResponse` as a type.

Here the type is updated after running the response types generation script

Note that I manually discarded many changes after type generation: fields were added, new responses, ... In order to keep it in scope. Also some changes were related to code formatting and couldn't find the formatter tool & config used for the project if any.

### DIFF
--- a/packages/web-api/src/types/response/MigrationExchangeResponse.ts
+++ b/packages/web-api/src/types/response/MigrationExchangeResponse.ts
@@ -17,9 +17,6 @@ export type MigrationExchangeResponse = WebAPICallResult & {
   ok?:               boolean;
   provided?:         string;
   team_id?:          string;
-  user_id_map?:      UseridMap;
+  user_id_map?:      { [key: string]: string };
   warning?:          string;
 };
-
-export interface UseridMap {
-}

### CODE REVIEW

File: packages/web-api/src/types/response/MigrationExchangeResponse.ts [hunk 1]
```diff
@@ -17,9 +17,6 @@ export type MigrationExchangeResponse = WebAPICallResult & {
   ok?:               boolean;
   provided?:         string;
   team_id?:          string;
-  user_id_map?:      UseridMap;
+  user_id_map?:      { [key: string]: string };
   warning?:          string;
 };
-
-export interface UseridMap {
```
davidlj95, line 24: ⚠️ **BREAKING CHANGE**

If anyone was using that interface. Though given it was not very useful and cast was needed anyway, removing usage would be enough to fix it.

filmaj: How exactly is this a breaking change? Because the values within `user_id_map` were typed to `any` before, and now would be a `string`?

filmaj:
> Note that I manually discarded many changes after type generation

That is a problem, as when we go to generate the responses the next time, we will lose these changes, right?

We have been discussing internally with @seratch about possibly combining the response type generation together with manual curation / integration with manually-crafted types in order to improve the experience (probably best shown off in @seratch's [personal Slack client project's type generation here](https://github.com/seratch/slack-web-api-client/blob/0.12.0/scripts/code_generator.rb#L30-L65): for certain responses containing Block Kit elements, the generated types "link" up the correct, manually-curated `Block` types to the responses).

If anything, I would like for us to use a similar approach to solve this issue: let the `quicktype`-based automation do its things to automatically generate baseline, but sub-optimal, types, and augment them with targeted, manually-written types in (at least at the beginning) a few specific areas that provide the best developer experience value. Something like [this example I put together in a TypeScript Playground](https://www.typescriptlang.org/play/?#code/PTAEBcAsFNQMwPYBskIO4EsB2BzUPotoAnAQ3GgBNQAjDcAZ1CQzgurmIQFtRJxwABwYAuEDnqQArjQB0AYx7AGSUvIDWpQRmBYElaAFoVa9ccrrgNVDWDdS2YINOkCDYGmg1DWnQ2LywOAAnoLQ7sThgghYDNDAAGoY0GgMAPJhWABKUTFxsowAUNgUxHBqsADCqHGgAN6FoKDQ3AgAVhgA-CJNTTQIyNCkWADcjRDQAB7g3b1NDODE2DhjTSFhs3PVCHEAKqHQq6AAbiQ05Bjcs-2Dw2MAvoVT0cTgzVhSvNt7B-XjAKLcQQhUAAXlAAHIIQAacYAWWI6koaCwYMh3ERyKwMPGAAVVNhdlM3uCIYICVgAPoUaY4x4lEjleSwADihBI5CoSRSOQY0VisAaawOswWS1wRwwlFFi2WR3A9CQ0Fm30OhUeIFAsm1tCkbygsEQKHQy1AnlA8jIbFA9g+pBQwVo0HkpCktXNAGkAHJpADqEBgoEifLysEwKFA9rQpGCTEi4CkxFRBtqgi4YVeyQYxSwpSZsA9ehRACFSHFuWhefzakLQMAAFT10AAAQMDEtGGBGBikc8DB4sAAUgBlAAiCHkTBdqJosEgw0ohjQSwEhD4JGg0IgA1A9eA43W0B6ACIxctj0cG03W+EO12e6Q+wPQCPx5OLcMnXwF0uVxRUTAkRbuAO57uMUo9GeEoHoqR5-HMmrmuoRaRja+j2hAsGgBgTBRjGuGgOSDhUjSbzQEq3CEOAACE4zCmEJ5Edg1LEhedETNMkGytBTSPE0moLqA-agD2iDEIR6YkAq4RmnOpCnHwGA4JAFoxHAUqEMy2GonhsZOqaabhFR6qFJq6lYPaSDBFuzwIK8qHcBgkwiXAAawOAaAILRtn2YeoAVlWoZomyRBkOwAW5AKoAAGSgIW6BYKW5bJJWkVxGMiixGRkykECSo9BFIZReCdR1mAgkoeaBBvGmCAZhAoTLFuaCAe5gbIQlEl1VJWbwFwvDxSWZbQIV1awAwUgZnEBhuSJ94WUgXUZtJTCcDwoAhRy4UpYFAqFOqIxAA).

packages/web-api/src/types/response/MigrationExchangeResponse.ts
seratch: I don’t consider this a breaking change. The auto-generated types here have been changed based on quicktype outputs over time. We've been trying to maintain backward compatibility as much as possible, but when an existing type is incomplete and/or invalid, changes like this can happen.

seratch:
Specifically talking about this PR (rather than general discussion on the future), if @davidlj95 didn't include any unrelated changes, that's fine. When we run the generator script, other changes won't conflict with this PR. I understand @davidlj95 might hesitate to include other changes.

If @davidlj95 made some manual changes to the generated outputs, those changes will be overwritten when any of us executes the script next time (as @filmaj mentioned).

Either way, the generator script should generate code in an idempotent manner (as long as the JSON source files remain the same). Therefore, I'd suggest either:
1. Including other changes in this PR
2. @filmaj running the generator script on behalf of @davidlj95 

As for the option 2, I can run the generator tomorrow too (if others don't have time in North American timezone today).

packages/web-api/src/types/response/MigrationExchangeResponse.ts
filmaj: Agreed. Perhaps in the future, if we start paying closer attention to the response types, and start augmenting them with manual types, we could employ a more stringent backwards compatibility policy for the response types.

filmaj:
I regenerated the latest response types in [the branch `migration-exchange-as-map-regenerated`](https://github.com/slackapi/node-slack-sdk/tree/migration-exchange-as-map-regenerated). It didn't seem to modify the response payload 🤔 so maybe the file as-is here was not manually tweaked / it is fine?

packages/web-api/src/types/response/MigrationExchangeResponse.ts
davidlj95: It's technically a breaking change regarding to type definitions as:
1. **Interface `UseridMap` is exported**. It's also included in published `dist/` files. So it's part of public API surface
1. **Interface `UseridMap` gets removed**. Which is a breaking change in case someone was importing it & using it.
1. **Type definition is narrower**. Before it was very open ([any non null type](https://typescript-eslint.io/rules/no-empty-interface/)). Now it's an [indexed type](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures) which is a stricter type. So it's also non backwards compatible. 

But given the type needed a cast anyways and cast was needed it's probably not enough for a new major version bump due to this kind of small breaking change. And more if auto-generated types are following a good-enough approach. 

I'm happy with whatever you decide, reasons sound fair enough to me :)

filmaj (approved):
Thanks for confirming you didn't make any manual changes to the changed file in this PR! That tells me the generation script output for `MigrationExchangeResponse.ts` is correct and idempotent, so I will merge this PR.

Thanks you for the contribution! 🙇

PR #1825, opened 2024-06-18
chore: tweak response types generation script

filmaj:
- dont reclone java-slack-sdk every time
- pass `--prefer-const-values` to not generate `enum`s in place of string values; this should prevent creating enums like `['', 'U000000']` based off of sample .json files from java-slack-sdk


seratch (approved):
This looks great!

PR #1827, opened 2024-06-18
chore: tweak lint config to allow for `eslint --fix` to fix import order

filmaj:
Small maintainer QoL fix: tweaked the eslint config to use a combination of ESLint's stock `sort-imports` rule and `eslint-plugin-import`'s `import/order` rule to allow for `eslint --fix` to fix import ordering. This was driving me a bit crazy _eyetwitch_ and I found [this comment](https://github.com/import-js/eslint-plugin-import/issues/2521#issuecomment-1216630234) that showed off using the two rules in combo.

Also ensured all the `lint` npm run scripts use the `--fix` rule across all monorepo packages.

The only change worth reviewing is within `lint-configs/` - all other file changes are simply style fixes.


packages/cli-test/src/cli/shell.spec.ts
```diff
@@ -1,9 +1,12 @@
...
-import stream from 'stream';
 import EventEmitter from 'events';
+import stream from 'stream';
+
+import { assert } from 'chai';
+import sinon from 'sinon';
+
 import { shell } from './shell';
+
 import type { ShellProcess } from '../utils/types';
```
zimeg, line 10: 😱 Big fan of splitting stdlib from dependencies from local from types if that's what this rule is requiring!

packages/cli-test/src/utils/logger.ts
```diff
@@ -1,4 +1,4 @@
-import { createLogger, format, transports, Logger } from 'winston';
+import { Logger, createLogger, format, transports } from 'winston';
```
zimeg, line 1: Uppercase alphabetization is a big winston 🏆

packages/web-api/src/methods.ts
```diff
@@ -222,272 +488,8 @@ import type {
...
-  AdminWorkflowsSearchResponse,
-  AdminWorkflowsUnpublishResponse,
-  AdminWorkflowsCollaboratorsAddResponse,
-  AdminWorkflowsCollaboratorsRemoveResponse,
-  AdminWorkflowsPermissionsLookupResponse,
-  AdminAppsConfigLookupResponse,
-  AdminAppsConfigSetResponse,
 } from './types/response/index';
 // Request types
-import type {
```
zimeg, line 248: 🗒️ Might be nice to also swap 🪓 the `request` and `response` comments for these moving types!

packages/web-api/src/types/request/chat.ts
```diff
@@ -1,17 +1,18 @@
...
   TokenOverridable,
 } from './common';
-import { OptionalArgument } from '../helpers';
+import type {
+  Block, // TODO: these will be combined into one in a new types release
+  KnownBlock,
+  LinkUnfurls,
+  MessageAttachment,
+  MessageMetadata,
+} from '@slack/types';
```
zimeg, line 15: This ordering seems unexpected to me - I'd've thought `@slack/types` finds a place before `./*` imports - but no blocker if this is what the linter likes 🙏 

Also only noticed this in a few `web-api` files... Not sure why though!

zimeg (approved):
Awesome PR 😭 The mental friction of "where do I put this import" shall no longer be. And this just looks a lot better. Thank you!

Left one comment with confusion on import ordering and one on comments, but neither are blocking! 🚀

packages/web-api/src/types/request/chat.ts
filmaj: I'm also not sure why.... but I almost don't want to ask questions. Just tell me how to order the damn things and don't make me think!

PR #1856, opened 2024-07-29
🔪 cli-test: remove command and custom error classes

filmaj:
let the exceptions fall where they may! wrapping exceptions with TWO additional custom error classes did not provide additional value; the native stack trace support is perfectly fine.

Setting the base for this PR as the branch `cli-test-one-point-oh` as I inch towards a 1.0 for the test utility - that 1.0 release will be needed for the internal CI project.


vegeris (approved):
So the only additional info we append is the function the exception was thrown in, but that info is also covered by the native stacktrace? 

🔪

PR #1858, opened 2024-07-29
feat (cli-test): expose arbitrary apihost flag to underlying SlackCLIProcess class

filmaj:
Inching towards 1.0 release.

Incremental step to refactoring the command options, which will be a third PR I will work on shortly.


vegeris (approved):
Nice

PR #1864, opened 2024-08-06
cli-test: tackle TODOs, revamp arguments and function signatures, expose all CLI options

filmaj:
This PR changes all the function signatures to to be consistent. It also factors out types for arguments where possible, and exhaustively types individual properties representing each CLI command's different options.

Unit test coverage was added for the commands, bringing the module coverage up close to 100%.

# Review Guidance

In depth reviewing of all the `src/cli/commands` files is likely not needed (unless you want to see the detailed changes). Commands were changed in relatively uniform ways:

- All commands now accept a single object argument, almost all of which extend from `ProjectCommandArguments`, which itself is just `SlackCLIGlobalOptions` + an `appPath` argument that points to the location of the app project on the filesystem.
- Individual command methods delegate most of the work to the `SlackCLIProcess` class (`src/cli/cli-process.ts`). This class manages assembling the CLI invocation command string and parses global options.
  - Most command methods chiefly deal in parsing command-specific arguments into command-specific CLI flags.

However, please look at the `src/cli/cli-process.ts` file - do the global option names make sense? I left a comment in this file about the naming of the `instance` property, which maps to `--app` on the CLI - I wonder if it should just be named the same as the flag, but `app` is such a vague property name that I decided to change it to `instance`.

### DIFF
--- a/packages/cli-test/src/cli/cli-process.spec.ts
+++ b/packages/cli-test/src/cli/cli-process.spec.ts
@@ -76,7 +76,7 @@ describe('SlackCLIProcess class', () => {
         await cmd.execAsync();
         sandbox.assert.calledWithMatch(spawnProcessSpy, '--skip-update');
       });
-      it('should default to `--app deployed` but allow overriding that via the `instance` parameter', async () => {
+      it('should default to `--app deployed` but allow overriding that via the `app` parameter', async () => {
         let cmd = new SlackCLIProcess('help');
         await cmd.execAsync();
         sandbox.assert.calledWithMatch(spawnProcessSpy, '--app deployed');
--- a/packages/cli-test/src/cli/cli-process.spec.ts
+++ b/packages/cli-test/src/cli/cli-process.spec.ts
@@ -81,7 +81,7 @@ describe('SlackCLIProcess class', () => {
         await cmd.execAsync();
         sandbox.assert.calledWithMatch(spawnProcessSpy, '--app deployed');
         spawnProcessSpy.resetHistory();
-        cmd = new SlackCLIProcess('help', { instance: 'local' });
+        cmd = new SlackCLIProcess('help', { app: 'local' });
         await cmd.execAsync();
         sandbox.assert.calledWithMatch(spawnProcessSpy, '--app local');
       });
--- a/packages/cli-test/src/cli/cli-process.spec.ts
+++ b/packages/cli-test/src/cli/cli-process.spec.ts
@@ -76,6 +76,38 @@ describe('SlackCLIProcess class', () => {
         await cmd.execAsync();
         sandbox.assert.calledWithMatch(spawnProcessSpy, '--skip-update');
       });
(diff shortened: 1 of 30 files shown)


packages/cli-test/src/cli/cli-process.ts
```diff
@@ -15,6 +15,13 @@ export interface SlackCLIGlobalOptions {
    * `qa` and `apihost` will both supersede this option.
    */
   dev?: boolean;
+  /** @description Ignore warnings and continue executing command. Defaults to `true`. */
+  force?: boolean;
+  /**
+   * @description Application instance to target. Can be `local`, `deployed` or an app ID string.
+   * Defaults to `deployed`.
+   */
+  instance?: string;
```
filmaj, line 24: Since this maps to the `--app` flag, perhaps `instance` is not the right name for this option? WDYT?

packages/cli-test/src/cli/cli-process.ts
```diff
@@ -15,6 +15,13 @@ export interface SlackCLIGlobalOptions {
    * `qa` and `apihost` will both supersede this option.
    */
   dev?: boolean;
+  /** @description Ignore warnings and continue executing command. Defaults to `true`. */
+  force?: boolean;
+  /**
+   * @description Application instance to target. Can be `local`, `deployed` or an app ID string.
+   * Defaults to `deployed`.
+   */
+  instance?: 'local' | 'deployed' | string;
```
vegeris, line 24: :twocents: `app` seems like a reasonable param name to me, intuitive and matches the flag. `instance` seems more vague

packages/cli-test/src/cli/cli-process.ts
```diff
@@ -24,13 +31,15 @@ export interface SlackCLIGlobalOptions {
    * @description Whether the CLI should skip updating (`--skip-update`). Defaults to `true`.
    */
   skipUpdate?: boolean;
-  /**
-   * @description workspace or organization name or ID to scope command to
-   */
+  /** @description Workspace or organization name or ID to scope command to. */
```
vegeris, line 34: Could be more specific for the org case although `workspace or org ID` is also accurate; eg. `The ID of your team. if you are using a Standard Slack plan, this is your workspace ID. If you are using an Enterprise Grid plan, this is the organization ID, even if your app is only granted to a subset of workspaces within the org.`

vegeris (approved):
LGTM :)

packages/cli-test/src/cli/cli-process.spec.ts
```diff
@@ -76,6 +76,38 @@ describe('SlackCLIProcess class', () => {
         await cmd.execAsync();
         sandbox.assert.calledWithMatch(spawnProcessSpy, '--skip-update');
       });
+      it('should default to `--app deployed` but allow overriding that via the `instance` parameter', async () => {
```
vegeris, line 79: nit: `instance` to `app`

PR #1897, opened 2024-08-21
cli-test(feat): expose timeout argument for activity tail and run start/stop methods

filmaj:
Problematic to not be able to customize the timeout of `run` or `activity --tail`, as the default 10 second timeout can be difficult to manage.

### DIFF
--- a/packages/cli-test/src/cli/shell.ts
+++ b/packages/cli-test/src/cli/shell.ts
@@ -173,15 +173,22 @@ export const shell = {
   waitForOutput: async function waitForOutput(
     expString: string,
     proc: ShellProcess,
+    opts?: {
+      /** @description How long to wait for expected output in milliseconds. Defaults to 10 seconds. */
+      timeout?: number;
+    },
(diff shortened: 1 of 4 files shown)


packages/cli-test/src/cli/shell.ts
```diff
@@ -173,15 +173,22 @@ export const shell = {
...
+      /** @description How long to wait for expected output in milliseconds. Defaults to 10 seconds. */
+      timeout?: number;
+    },
   ): Promise<void> {
     const delay = 1000;
+    const timeout = opts?.timeout || timeouts.waitingAction;
     let waitedFor = 0;
     let timedOut = false;
     while (!proc.output.includes(expString)) {
+      console.log('OUTPUT', proc.output);
```
vegeris, line 186: Were these debug statements? Otherwise, based on the linter complaints it sounds like we might want to use something like `logger.info`

vegeris (approved):
LGTM! Should be very helpful

PR #1901, opened 2024-08-28
chore (web-api): update response types w/ latest generated sample responses

filmaj:
Latest changes from the java-slack-sdk's samples.

### DIFF
--- a/packages/web-api/src/types/response/AdminConversationsSearchResponse.ts
+++ b/packages/web-api/src/types/response/AdminConversationsSearchResponse.ts
@@ -21,7 +21,7 @@ export type AdminConversationsSearchResponse = WebAPICallResult & {
 };
 
 export interface Conversation {
-  canvas?:                        Canvas;
+  canvas?:                        ListsClass;
   channel_email_addresses?:       ChannelEmailAddress[];
   channel_manager_count?:         number;
   connected_limited_team_ids?:    string[];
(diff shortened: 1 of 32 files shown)


packages/web-api/src/types/response/AdminConversationsSearchResponse.ts
```diff
@@ -47,14 +47,15 @@ export interface Conversation {
...
+  lists?:                         ListsClass;
   member_count?:                  number;
   name?:                          string;
   pending_connected_team_ids?:    string[];
+  properties?:                    Properties;
   purpose?:                       string;
 }
 
-export interface Canvas {
+export interface ListsClass {
```
WilliamBergamin, line 58: Since this is exported isn't this name change a breaking change? 🤔

WilliamBergamin (changes requested):
Looks good, but I do have one concern about a potential breaking change

packages/web-api/src/types/response/AdminConversationsSearchResponse.ts
filmaj: It technically is, but this happens all the time due to the type generation script, which uses `quicktype` under the hood. We only export the `*Response` interfaces from the main package (see [here](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/types/response/index.ts) and [here](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/index.ts#L34)), so developers would have to explicitly import from `@slack/web-api/src/types/response/SomeAPIResponse` to get at these sub-interfaces. Ideally these would not be exported - I agree with you - but that would take some more work on the type generation.

seratch: It could be, but this is inevitable as long as we auto-generate it using the 3rd party tool. We generally recommend explicitly referring only on XXXResponse types.

WilliamBergamin (approved):

Issue #1904, opened 2024-08-29
types 3.0: breaking change for event payloads


filmaj:
This would be a follow-up issue to tackle once slackapi/bolt-js#1395 is addressed; slackapi/bolt-js#1395 simply adds the existing, likely lacking, types for Events API payloads to the `types` package. This issue would be about _breaking_ those payloads and include that in a new 3.0 release of `@slack/types`.

# Summary

## Improvements to `MessageEvent`

Today, `MessageEvent` is a big union of different events Slack may send to apps, presumably that include some message information. Because it is a union, when using it as a type both in applications as well as frameworks (like bolt), only the _common_ properties across all types in this union can be used. Some properties that would be expected to be present on messages are not common in this union today: `text` and `blocks`, for instance. We should study how this type is used in bolt and elsewhere, and consider a different approach to modeling message events. For example, perhaps a generics-based approach could better serve?

Another problem when working with `MessageEvent` in TypeScript is the fact that one member of the union, the `GenericMessageEvent`, has `subtype: undefined`. This causes major problems when trying to match against the union of all different kinds of messages, since many message subtype events have a defined string literal `subtype`. `subtype: undefined` needs to get the 🔪 .

# TODO

## Review Issues to Identify Pain Points

The following should be studied to determine the scope of what these breaking changes should include, in order to address developer pain points, and folded into the Summary section above in order to plan the changes:

- https://github.com/slackapi/bolt-js/issues/904
- https://github.com/slackapi/bolt-js/issues/796
- https://github.com/slackapi/bolt-js/issues/1432
- https://github.com/slackapi/node-slack-sdk/issues/2025
- https://github.com/slackapi/node-slack-sdk/issues/2026
- https://github.com/slackapi/bolt-js/issues/2122

Issue #1768, opened 2024-04-02
(Set a clear title describing your idea)


slifty:
I'm using `@slack/socket-mode` in a TypeScript project and would love to be able to benefit from type definitions when defining event handlers -- e.g. when I invoke something along the lines of

```
socketModeClient.on(
  'reaction_added',
  async (event) => {
    ...
  }
);
```

I'd like `event` to not be of the `any` type.

It looks like there are [libraries](https://github.com/andywer/typed-emitter) that might support this, for your consideration.

#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [x] `@slack/socket-mode`
- [ ] `@slack/types`
- [ ] I don't know

### Requirements

Please read the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/main/.github/contributing.md) and [Code of Conduct](https://slackhq.github.io/code-of-conduct) before creating this issue or pull request. By submitting, you are agreeing to those rules.

filmaj:
I agree this would be a fantastic enhancement. The issue is that perfect typing for Slack payload events is challenging, as the payloads can be quite contextual. For example, depending on the Slack workspace type (enterprise vs. standalone), the payloads might differ. Additionally, some payloads may or may not contain properties depending on the context, for example 
a [`block_actions` payload contains a `container` property](https://api.slack.com/reference/interaction-payloads/block-actions?ref=bk#fields), which might be a `View` object or might be a containing message.

That said, at least some baseline form of event typing should be doable. I will leave this as an enhancement issue open but can make no promises on a delivery timeline.

PR #1907, opened 2024-08-29
feat: adding event payloads from bolt-js as a starting point

### LINKED ISSUE #1395: types: add event payloads
# Edits

What about adding types for event payloads to the types package? This could be used to then help address a variety of issues across different packages (bolt, socket-mode, deno-sdk):

- Better typing of interactivity handler payloads in slackapi/deno-slack-sdk, see:
  - https://github.com/slackapi/deno-slack-sdk/blob/main/src/functions/interactivity/block_actions_types.ts
  - https://github.com/slackapi/deno-slack-sdk/blob/main/src/functions/interactivity/block_suggestion_types.ts
  - https://github.com/slackapi/deno-slack-sdk/blob/main/src/functions/interactivity/view_types.ts
- https://github.com/seratch/slack-web-api-client/blob/main/src/manifest/events.ts
- https://github.com/slackapi/bolt-js/blob/main/src/types/events/base-events.ts
- https://github.com/slackapi/bolt-js/blob/main/src/types/events/message-events.ts
- https://github.com/slackapi/bolt-js/blob/main/src/types/actions/block-action.ts
- https://github.com/slackapi/bolt-js/blob/main/src/types/options/index.ts
- https://github.com/slackapi/bolt-js/blob/main/src/types/view/index.ts

# Original Message

Is really necessary to use the type generator instead of manually improve types? I've found out that the ModalView is missing some parameters, like `app_installed_team_id`, `id` and `state`. About the `state`, on the places that have references to it, it is just a string, it could be so much better. Depending on your answer, I can help you with the types that I've created.

#### Packages:

Select all that apply:

- [ ] `@slack/web-api`
- [x] `@slack/types`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] I don't know

filmaj:
fixes #1395

# Summary

This PR doesn't do much other than copy, and slightly split up into separate files, the bolt-js types present here: https://github.com/slackapi/bolt-js/tree/main/src/types/events

(I have a WIP PR for bolt-js that passes all tests w/ this `types` PR here: https://github.com/slackapi/bolt-js/pull/2223)

# The Big Idea

Let's move event payloads into this `types` package, since it revisions stuff like "shapes of things." Why? While to date these types have existed in bolt-js, they could be reused in other packages: socket-mode (see https://github.com/slackapi/node-slack-sdk/issues/1768) and deno-slack-sdk, for one (yes, deno packages, too, can consume `@slack/types`!).

This should be a backwards compatible addition, and nothing in bolt should break as a result of this move - step one is just shifting code around. Step two, though, is much more interesting: what sorts of breaking changes can we introduce in a new types v3 package that improves the lives of our developers? Very interesting question that is tracked in https://github.com/slackapi/node-slack-sdk/issues/1904.


packages/types/src/events/reaction.ts
```diff
@@ -0,0 +1,23 @@
+interface ReactionMessageItem {
```
misscoded, line 1: Should this be exported? Also, since this is a `message` type, should it be in message? That message file was the one that gave me the most pause. 😅

packages/types/src/events/shared-channel.ts
```diff
@@ -0,0 +1,135 @@
+interface SharedChannelTeamItem {
```
misscoded, line 1: Not sure, but should all of these be exported?

packages/types/src/events/steps-from-apps.ts
```diff
@@ -0,0 +1,86 @@
+export interface WorkflowDeletedEvent {
```
misscoded, line 1: Should we rename this file `workflow`? These are going away, anyway, so maybe not important at all.

packages/types/src/events/message.ts
```diff
@@ -0,0 +1,368 @@
...
+    reply_users_count: number;
+    latest_reply: string;
+    reply_users: string[];
+  };
+  client_msg_id: string;
+  channel: string;
+  channel_type: ChannelTypes;
+}
+
+interface BotProfile {
```
misscoded, line 288: No idea what this event is, but I assume we should export it. Same as `File` below.

packages/types/src/common/bot-profile.ts
```diff
@@ -0,0 +1,11 @@
+export interface BotProfile {
```
misscoded, line 1: I think this event is also in `message.ts`?

packages/types/src/events/message.ts
```diff
@@ -0,0 +1,368 @@
...
+  app_id: string;
+  team_id: string;
+  icons: {
+    [size: string]: string;
+  };
+  updated: number;
+  deleted: boolean;
+}
+
+interface File {
```
misscoded, line 300: Should this be in `file.ts` instead?

misscoded (approved):
The division of payloads makes sense to me, though the `message` one gave me the most pause since there seems to be some overlap with other things (and a rogue `message` event being housed elsewhere, as well).

Left some comments about non-exported items, and what I think are duplicates. Other than that, LGTM! :shipit:

packages/types/src/events/channel-membership.ts
```diff
@@ -0,0 +1,18 @@
+export interface MemberJoinedChannelEvent {
```
srajiang, line 1: Should this file be member.ts for consistency with most of the other file names? They seem to be derived from the first unique word/wordcombination all the contained types share in common?

srajiang (approved):
Naming schema very sensible, I left a suggestion around the channel-membership file.

packages/types/src/common/bot-profile.ts
filmaj: The idea here is that common things reused across different events (and maybe even more generally - not just events) will exist in `src/common`. This `BotProfile` thing is exported and used in two of the `src/event` files.

packages/types/src/events/channel-membership.ts
filmaj: Sounds good to me! Will update shortly.

packages/types/src/common/bot-profile.ts
misscoded: I guess I don't follow why it'd be defined in two places: isn't that superfluous?

packages/types/src/events/message.ts
filmaj: Whoops, good catch, actually this should import from `src/common/bot-profile`!

packages/types/src/events/message.ts
filmaj: No, because `file.ts` holds the file-related event shapes, while here within `message.ts` we house message event related shapes, which look like contain file-related information (thus the existence of this interface).

I think once I get to the refactoring task (https://github.com/slackapi/node-slack-sdk/issues/1904), I will take a look at the entire types package holistically and see where else we need a `File` interface and make sure it's reused appropriately.

packages/types/src/events/reaction.ts
filmaj: It should not be exported, as it is only used in this file, which models reaction-event payloads.

packages/types/src/events/shared-channel.ts
filmaj: Nope! As soon as we do, we can't change them or rename them without making a breaking change. Anything not exported is only used within the file it exists in. In this case, all of these unexported interfaces are only re-used within the `*Event` payloads present in this file.

packages/types/src/events/reaction.ts
misscoded: Got it! Skimmed these so missed the usage, but that makes sense.

packages/types/src/events/steps-from-apps.ts
filmaj: Yeah I muddled over that - do you think that's better? If so I'll rename it! Indeed, presumably next month when SFA is 🔪 and hopefully I get to a types v3 by then, this will all go away.

packages/types/src/common/bot-profile.ts
filmaj: 100% 😄 the duplicate was a mistake of mine, and it's now removed.

packages/types/src/events/steps-from-apps.ts
misscoded: No need to change it if it's getting tossed, anyway, IMO

PR #2006, opened 2024-09-04
chore: remove eslint, use biome instead

filmaj:
Many lines of code, little actual change. Perhaps the only thing worth taking a look at is [the `biome.json` file](https://github.com/slackapi/node-slack-sdk/pull/2006/files#diff-2bc8a1f5e9380d5a187a4e90f11b4dd36c3abad6aea44c84be354a4f44cdec55) that describes the lint and formatting rules.

# Changes

- gets us out of the current tough situation where eslint v9 is new but we are stuck on v8 due to the airbnb eslint config being archived and unmaintained
- removes eslint, adds biome instead. https://biomejs.dev/ is way faster (eslint executes in dozens of seconds while biome executes in under a second) and only a single dependency compared to the complicated eslint buildchain
- moves all tests from .js to .ts (commonJS to ESM) <-- this took the most work, finally at least a vast majority of tests are typesafe now. caught many errors because of this (like invalid HTTP response mocks and things of the like).


seratch (approved):
Nice! Too many changes to review but they should be safe. Regarding the changes under `packages/web-api/src/types/response/`, it's fine! but we may want to add the additional step that applies biome after generating in the script.

filmaj:
Sounds @seratch , I will follow up with another PR for that shortly, and can test it out by running it in the PR, too.

PR #2008, opened 2024-09-05
ci: add an integration test with bolt-js

filmaj:
WIP


filmaj:
Damn, can't do this until we release a new major version of bolt! To use the latest node SDK packages :(

PR #2033, opened 2024-09-17
feat (web-api / types): Add support for assistant.threads.* API

misscoded:
### Summary

Adds support for `assistant.threads.*` endpoints, including:

- [assistant.threads.setStatus](https://api.slack.com/methods/assistant.threads.setStatus)
- [assistant.threads.setSuggestedPrompts](https://api.slack.com/methods/assistant.threads.setSuggestedPrompts)
- [assistant.threads.setTitle](https://api.slack.com/methods/assistant.threads.setTitle)


packages/types/src/events/assistant.ts
```diff
@@ -0,0 +1,29 @@
+export interface AssistantThreadStartedEvent {
```
misscoded, line 1: @filmaj These are used in the implementation for Bolt, but figured they should go in now. Let me know if you prefer these to be separate!

packages/types/src/events/assistant.ts
```diff
@@ -0,0 +1,29 @@
...
+  type: "assistant_thread_started";
+  assistant_thread: {
+      user_id: string;
+      context: { 
+        channel_id: string; 
+        team_id: string; 
+        enterprise_id: string;
```
seratch, line 8: the `context` property can be empty when an end-user started a convo next to non-channel views such as "Unread", "Threads", and others
```suggestion
        channel_id?: string; 
        team_id?: string; 
        enterprise_id?: string;
```

packages/web-api/src/types/response/AdminAppsActivitiesListResponse.ts
```diff
@@ -8,80 +7,81 @@
 //                                                                                     //
 /////////////////////////////////////////////////////////////////////////////////////////
 
-import type { WebAPICallResult } from '../../WebClient';
+import { WebAPICallResult } from '../../WebClient';
```
seratch, line 10: @filmaj should know these differences. Formatting code after generating them is fine, but the process must be incorporated into the generator script.

seratch (changes requested):

packages/web-api/src/types/request/assistant.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface AssistantThreadsSetStatusArguments extends TokenOverridable {
+  channel_id: string;
+  status: string;
+  thread_ts: string;
+}
+
+// https://api.slack.com/methods/assistant.threads.setSuggestedPrompts
+export interface AssistantThreadsSetSuggestedPromptsArguments extends TokenOverridable {
+  channel_id: string;
+  prompts: {
```
filmaj, line 13: Are 0 prompts acceptable to send in the request? If not, I type arguments that are of the form "at least one" like so:

```typescript
prompts: [SomeType, ...SomeType[]];
```

So if at least 1 prompt must be provided, I'd factor the prompt object w/ `title` and `message` into its own interface and use the above technique to signal to developers "at least 1." See for example [this part of `reminders.ts`](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/types/request/reminders.ts#L13).

filmaj (changes requested):
LGTM, left one suggestion re: factoring out a type and specifying "at least 1" (if applicable).

As for the formatting, I recently changed the linter (#2006), and I don't think I re-ran generating the response types since then. @misscoded can you try resetting your `node_modules` folder and running the lint task locally? `rm -rf node_modules package-lock.json && npm it`. I have a feeling that will fix the response types and tighten up the diff in this PR.

filmaj (approved):

packages/web-api/src/types/response/ReactionsGetResponse.ts
```diff
@@ -21,7 +20,8 @@ export type ReactionsGetResponse = WebAPICallResult & {
 
 export interface Message {
   app_id?: string;
-  blocks?: MessageBlock[];
+  assistant_app_thread?: AssistantAppThread;
+  blocks?: AssistantAppThreadBlock[];
```
misscoded, line 24: This can't be correct, can it? Also, `MessageBlock` below being replaced by an Assistant-specific type. @filmaj @seratch

filmaj: It's just the type generation's obtuse naming. If you look at what the definition of these things are, they're the same as before, but for whatever reason the type generator decided to name the interface after assistant stuff 🤷

seratch: The naming is done by quicktype and it's not controllable. If a 3rd party developer has the exported name in their code, indeed this could be a change. However, most developers won't care what the name of these internal types. Also, there is no way for us to rever this.

misscoded: Alright, it sounds like we're OK with this change then. Thanks for confirming 🙇🏼‍♀️

packages/types/src/events/assistant.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export interface AssistantThreadContextChangedEvent {
+  type: "assistant_thread_context_changed";
+  assistant_thread: {
+    user_id: string;
+    context: { 
+      channel_id: string; 
+      team_id: string; 
+      enterprise_id: string;
```
seratch, line 23: It seems the enterprise_id can be null if the workspace is not under an enterprise grid org.
```suggestion
      enterprise_id: string | null;
```

packages/web-api/src/types/request/assistant.ts
misscoded: Great suggestion – added and confirmed it works as expected in a sample.

PR #2041, opened 2024-09-27
cli-test(feat): Added `datastore` CLI commands

cchensh:
### Summary

This PR adds basic `datastore` CLI commands for e2e test, it includes `datastore get/put/query/delete` 

### DIFF
--- a/packages/cli-test/src/cli/commands/datastore.ts
+++ b/packages/cli-test/src/cli/commands/datastore.ts
@@ -7,11 +7,11 @@ export interface DatastoreCommandArguments {
   /** @description datastore get primary key value*/
   primaryKeyValue: string;
   /** @description datastore put [item details] */
-  putItem: string;
+  putItem: object;
   /** @description datastore query [expression] */
   queryExpression: string;
   /** @description datastore query [expression expression_values] */
-  queryExpressionValues: string;
+  queryExpressionValues: object;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/cli-test/src/cli/commands/datastore.ts
```diff
@@ -0,0 +1,82 @@
+import type { ProjectCommandArguments } from '../../types/commands/common_arguments';
+import { SlackCLIProcess } from '../cli-process';
+
+export interface DatastoreCommandArguments {
+  /** @description datastore get <query> */
+  getQuery: string;
```
cchensh, line 6: Here I'm using general string for datastore commands to make it easier than pre-define attributes such as `datastore`, `item`, and `expression`, etc

filmaj: I feel like this will make it easy to make mistakes. I think we should model these just like we model the actual methods:
- [`apps.datastore.get` method](https://api.slack.com/methods/apps.datastore.get) needs `datastore` string and `id` string
- [`apps.datastore.query` method](https://api.slack.com/methods/apps.datastore.query) needs `datastore` string, and its optional arguments, etc.

filmaj (changes requested):
I think we should model the API arguments as they are documented on our public docs 😞 sorry, I know that's annoying, but if we accept these JSON-payloads-as-strings, it would require a breaking change / major new version to change it to make it more accurate. I think the chance of making a typo with the string-based arguments is too high.

packages/cli-test/src/cli/commands/datastore.ts
cchensh: It makes sense, I'd say for `get` method, `id` is not a must as it's defined in the manifest as primary key. I'd like to change by adding the `datastore` as the new argument here.

filmaj (approved):

PR #2070, opened 2024-10-03
docs: sprucing up docs, esp. reference doc generation

filmaj:
- dynamically load generated sidebar items - requires a bit of gymnastics, because TypeDoc, which imports+parses TypeScript code to generate docs from source code comments, runs in parallel during docusaurus sidebar generation. As a result, sometimes there can be race-condition-like errors where the TypeDoc generation is not complete before the sidebar generation kicks off. Code was added to, basically, wait for the relevant sub-package TypeDoc routines to complete before finalizing the sidebar items.
- add biome / linting to docs folder
- lint during docs CI
- tweak generated reference documentation styles (tables instead of lists where possible)

Doing this as part of a an initial pass to learn the docusaurus ropes as I inch towards applying the same thing to bolt-js docs.

I left some in-line comments with screenshots of relevant changes. The big visual change is using tables for API references instead of bulleted lists: I think this is more visually pleasing and also scalable as we add more content to the docs.


docs/docusaurus.config.js
```diff
@@ -32,10 +55,12 @@ const config = {
       /** @type {import('@docusaurus/preset-classic').Options} */
       ({
         docs: {
+          sidebarItemsGenerator: async ({ defaultSidebarItemsGenerator, ...args }) => {
```
filmaj, line 58: Leveraged this documented approach here: https://typedoc-plugin-markdown.org/plugins/docusaurus/sidebar#injecting-into-autogenerated-sidebar

docs/docusaurus.config.js
```diff
@@ -4,7 +4,30 @@
...
+import { themes as prismThemes } from 'prism-react-renderer';
+import sidebarGenerator from './sidebars.js';
+
+/**
+ * Generate a typedoc / docusaurus config object for use as reference documentation generation options for each package.
+ * @param {string} pkg the sub-package to generate refdoc options for
+ */
+function packageReferenceOptions(pkg) {
+  return {
+    classPropertiesFormat: 'table',
```
filmaj, line 16: The various `*Format` properties and setting them to `table`, I think, is nicer than the default of a bulleted list - makes better use of horizontal space, and as we add more _kinds_ of information to our source code comments / JSdocs, these should show up as separate columns. E.g. if we add an `@examples` JSDoc tag to some method docs, that will show up as a new column in the generated docs.

docs/package.json
```diff
@@ -3,9 +3,11 @@
   "version": "2024.08.01",
   "private": true,
   "scripts": {
+    "lint": "npx @biomejs/biome check .",
+    "lint:fix": "npx @biomejs/biome check --write .",
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
-    "build": "docusaurus build",
+    "start": "TYPEDOC_WATCH=true docusaurus start",
+    "build": "TYPEDOC_WATCH=false docusaurus build",
```
filmaj, line 10: Followed these docs to set up proper TypeScript reference doc re-generation watching: https://typedoc-plugin-markdown.org/plugins/docusaurus/watch-mode

packages/logger/src/index.ts
```diff
@@ -9,59 +9,59 @@ export enum LogLevel {
 }
 
 /**
- * Interface for objects where objects in this package's logs can be sent (can be used as `logger` option).
+ * Logging interface exposing different methods to log at different {@link LogLevel}s.
```
filmaj, line 12: By using `{@link}` inline tags, the generated reference documentation figures out what to link to 😍 
Screenshot 2024-10-03 at 3 49 25 PM

packages/logger/src/index.ts
```diff
@@ -94,53 +95,35 @@ export class ConsoleLogger implements Logger {
     return this.level;
   }
 
-  /**
```
filmaj, line 97: We can forego the docs here because docs generation is smart enough to pull in, as a fallback, the description of this method from the `Logger` interface this class implements!

Screenshot 2024-10-03 at 3 50 52 PM

filmaj:
Odd, docs building seems to fail in CI but runs fine for me 🤔 need to investigate more.

.github/workflows/docs-deploy.yml
```diff
@@ -31,6 +31,22 @@ jobs:
         run: npm ci
         working-directory: ./docs
 
+      - name: Install sub-package dependencies
```
filmaj, line 34: Since the docs sidebar generation relies on importing and parsing TypeScript source of each sub-package we wish to document, we need to pull in the dependencies of each sub-package before building can begin.

docs/docusaurus.config.js
```diff
@@ -32,10 +58,12 @@ const config = {
       /** @type {import('@docusaurus/preset-classic').Options} */
       ({
         docs: {
+          sidebarItemsGenerator: async () => {
```
filmaj, line 61: Leveraged this documented approach here: https://typedoc-plugin-markdown.org/plugins/docusaurus/sidebar#injecting-into-autogenerated-sidebar

docs/docusaurus.config.js
```diff
@@ -70,96 +99,7 @@ const config = {
...
-    [
-      'docusaurus-plugin-typedoc',
-      {
-        id: 'webhook',
-        entryPoints: ['../packages/webhook/src/index.ts'],
-        tsconfig: '../packages/webhook/tsconfig.json',
-        out: "./content/reference/webhook",
-      }
-    ]
+    ...PACKAGES.map((pkg) => ['docusaurus-plugin-typedoc', packageReferenceOptions(pkg)]),
```
filmaj, line 102: DRYing things up here

docs/sidebars.js
```diff
@@ -1,152 +1,80 @@
...
+ */
+async function packageSidebarEntry(pkg) {
+  return {
+    type: 'category',
+    label: `@slack/${pkg}`,
+    link: {
+      type: 'doc',
+      id: `reference/${pkg}/index`,
+    },
+    items: await waitAndImportFile(`./content/reference/${pkg}/typedoc-sidebar.cjs`),
```
filmaj, line 27: One issue I saw playing around with the docs is that sometimes, as part of docs re-generation, I would get `MODULE_NOT_FOUND` errors where the sidebar generation was executing, and the generated typedoc files for each sub-package was missing. I _think_ this has something to do with the docusaurus lifecycle: plugins we use, like typedoc, are still executing will docusaurus is trying to generate the sidebar. This lifecycle is a bit unclear to me. In any case, to prevent this error, I rewrote this part so that `require`ing the relevant sub-package's generated typedoc only happens once the generated file is actually present. It seems to have addressed the issue.

docs/sidebars.js
```diff
@@ -1,152 +1,80 @@
...
       className: 'sidebar-title',
     },
-    'getting-started',
-    'typescript',
-    'packages/web-api',
-    'packages/oauth',
-    'packages/rtm-api',
-    'packages/webhook',
-    'packages/socket-mode',
+    { type: 'doc', id: 'getting-started' },
```
filmaj, line 46: Moving to the more fine-control approach of [providing our own asynchronous sidebar generator function](https://typedoc-plugin-markdown.org/plugins/docusaurus/sidebar#injecting-into-autogenerated-sidebar), for some reason, requires the output of the function to be the more verbose list-of-objects for sidebar items - I'd get errors otherwise. So, had to move away from the the convenience string items 🤷

docs/sidebars.js
```diff
@@ -1,152 +1,80 @@
...
-          label: '@slack/webhook',
-          link: {
-            type: 'doc',
-            id: 'reference/webhook/index',
-          },
-          items: require('./content/reference/webhook/typedoc-sidebar.cjs'),
-        },
-      ],
+      label: 'API Reference',
+      items: apiReferenceItems,
```
filmaj, line 75: This part was also DRYed up a bit

docs/src/css/custom.css
```diff
@@ -4,7 +4,7 @@
  * work well for content-centric websites.
  */
 
- :root {
+:root {
```
filmaj, line 7: Purely linter fixes

docs/src/theme/NotFound/Content/index.js
```diff
@@ -1,32 +1,26 @@
-import React from 'react';
-import clsx from 'clsx';
 import Translate from '@docusaurus/Translate';
 import Heading from '@theme/Heading';
-export default function NotFoundContent({className}) {
+import clsx from 'clsx';
```
filmaj, line 3: Purely linter fixes

docs/src/theme/NotFound/index.js
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
-import {translate} from '@docusaurus/Translate';
-import {PageMetadata} from '@docusaurus/theme-common';
+import { translate } from '@docusaurus/Translate';
```
filmaj, line 1: Linter fixes only here

docs/static/img/style.css
```diff
@@ -1,20 +1,20 @@
 /* Color variables */
```
filmaj, line 1: Linter fixes only here

docs/src/theme/NotFound/Content/index.js
```diff
@@ -1,32 +1,26 @@
...
             </Translate>
           </Heading>
           <p>
-            <Translate
-              id="theme.NotFound.p1"
-              description="The first paragraph of the 404 page">
-              If we've led you astray, please let us know. We'll do our best to get things in order. 
-
+            <Translate id="theme.NotFound.p1" description="The first paragraph of the 404 page">
+              If we've led you astray, please let us know. We'll do our best to get things in order.
```
hello-ashleyintech, line 17: this might be more of a copy nit for a follow up PR, but should we include some sort of link or CTA here so folks can leave feedback on the docs?

docs/docusaurus.config.js
hello-ashleyintech: nice! I agree the table is the way to go 🎉

docs/sidebars.js
hello-ashleyintech: is the file pretty much guaranteed to generate within 50ms or should we increase this for flexibility?

hello-ashleyintech (approved):
excellent work and LGTM 😄 left a few questions, mostly for my learning purposes 🧠

docs/src/theme/NotFound/Content/index.js
filmaj: Great call. @hello-ashleyintech perhaps you can help me, a total React noob, here: I tried wrapping the "please let us know" text in an `<a href=""></a>` tag and link to this repo's issue creation, but I get a, I think, React error 😢 

How.. does on do links in React? 👶 

Screenshot 2024-10-04 at 9 55 36 AM

hello-ashleyintech: ooooh yes! That looks like it might actually be caused by embedding a link in the `<Translate>` tag from Docusaurus. It seems like that tag only supports simple strings. The suggested workaround in the docs is to break this into substrings wrapped in the `<Translate>` tag - maybe to keep that link as part of the translation it could be something like:

```
<Translate>If we've led you astray, </Translate><a href=""><Translate>please let us know</Translate></a><Translate>. We'll do our best to get things in order.</Translate>
```

If that doesn't work though then we might need to just not translate the `please let us know` line for now 😬

lukegalbraithrussell:
Closing for now -- we ran into build issues with this approach. I'll take another crack at this early next year

Issue #2073, opened 2024-10-08
web-api: expose axios interceptors as `WebClient` constructor option


mtjandra:
I'd like to transform outgoing requests from the `@slack/web-api` package similar to other SDKs such as
- Github - [@octokit/rest](https://github.com/octokit/rest.js)
    - can transform outgoing requests uses the concept of [hooks](https://octokit.github.io/rest.js/v21/#hooks)
- Datadog - [@datadog/datadog-api-client](https://github.com/DataDog/datadog-api-client-typescript/)
    - can transform outgoing request by [overriding](https://github.com/DataDog/datadog-api-client-typescript?tab=readme-ov-file#configure-proxy) their [default httpClient](https://github.com/DataDog/datadog-api-client-typescript/blob/c6d3465bd0aeb8545facee8892a4a30e534cfc46/packages/datadog-api-client-common/http/isomorphic-fetch.ts#L13)

This Slack SDK allowa adding additional headers but I require more flexibility. Taking a look at the source it seems it is using axios under the hood ([ref](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/WebClient.ts#L242)). Axios seems to support the concept of [interceptors](https://axios-http.com/docs/interceptors) which would support my use case. 

I'd like to transform all outgoing requests into POST requests with a body representing the original request, add custom headers, then re-route that to a proxy. Is this possible to configure with the current version `@slack/web-api`?

If what I'm asking for is not currently possible, is there anything in the works?

----
**Side note:** was also thinking we could override the method making the outgoing request ([makeRequest](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/WebClient.ts#L564)), although I'm confused why all outgoing requests are [post requests](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/WebClient.ts#L611)

filmaj:
It may be possible in JavaScript, though you'd have to reach in to [the private `axios` instance contained within `WebClient`](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/WebClient.ts#L161-L164), which may be tricky - not sure what the state of the art for ignoring class visibility modifiers is in JS 🤷 

Alternatively, this could be a new feature request: expose some option on the `WebClient` constructor to provide the ability to specify interceptors.

[...]

PR #2076, opened 2024-10-11
feat(web-api): add request interceptor and HTTP adapter config to WebClient [ISSUE-2073]

mtjandra:
# Summary

Resolves https://github.com/slackapi/node-slack-sdk/issues/2073

## Features
- ability to configure one [axios request interceptor](https://axios-http.com/docs/interceptors)
    - useful for clients who would like to manipulate outgoing requests to conform to their custom proxies
- ability to configure an [axios adapter](https://axios-http.com/docs/req_config)
    - useful for client who would like to used their existing http client
        - ex: client may have a dedicated http client configured to be used with their internal proxy

## Contribution Pre-Requisites 

### DIFF
--- a/packages/web-api/src/WebClient.spec.ts
+++ b/packages/web-api/src/WebClient.spec.ts
@@ -972,7 +972,24 @@ describe('WebClient', () => {
   });
 
   describe('requestInterceptor', () => {
-    it('can intercept out going requests', async () => {
+    function configureMockServer(expectedBody: () => Record<string, unknown>) {
+      nock('https://slack.com/api', {
+        reqheaders: {
+          test: 'static-header-value',
+          'Content-Type': 'application/json',
+        },
+      })
+        .post(/method/, (requestBody) => {
+          expect(requestBody).to.deep.equal(expectedBody());
+          return true;
+        })
+        .reply(200, (_uri, requestBody) => {
+          expect(requestBody).to.deep.equal(expectedBody());
+          return { ok: true, response_metadata: requestBody };
+        });
+    }
+
+    it('can intercept out going requests, synchronously modifying the request body and headers', async () => {
       let expectedBody: Record<string, unknown>;
 
       const client = new WebClient(token, {
@@ -995,20 +1012,37 @@ describe('WebClient', () => {
         },
       });
 
-      nock('https://slack.com/api', {
-        reqheaders: {
-          test: 'static-header-value',
-          'Content-Type': 'application/json',
+      configureMockServer(() => expectedBody);
+
+      await client.apiCall('method');
+    });
+
+    it('can intercept out going requests, asynchronously modifying the request body and headers', async () => {
+      let expectedBody: Record<string, unknown>;
+
+      const client = new WebClient(token, {
+        requestInterceptor: async (config: RequestConfig) => {
+          expectedBody = Object.freeze({
+            method: config.method,
+            base_url: config.baseURL,
+            path: config.url,
+            body: config.data ?? {},
+            query: config.params ?? {},
+            headers: structuredClone(config.headers),
+            test: 'static-body-value',
+          });
+
+          await new Promise((resolve) => setTimeout(resolve, 100));
+          config.data = expectedBody;
+
+          config.headers.test = 'static-header-value';
+          config.headers['Content-Type'] = 'application/json';
+
+          return config;
         },
-      })
-        .post(/method/, (requestBody) => {
-          expect(requestBody).to.deep.equal(expectedBody);
-          return true;
-        })
-        .reply(200, (_uri, requestBody) => {
-          expect(requestBody).to.deep.equal(expectedBody);
-          return { ok: true, response_metadata: requestBody };
-        });
+      });
+
+      configureMockServer(() => expectedBody);
 
       await client.apiCall('method');
     });
--- a/packages/web-api/src/WebClient.spec.ts
+++ b/packages/web-api/src/WebClient.spec.ts
@@ -1032,7 +1032,6 @@ describe('WebClient', () => {
             test: 'static-body-value',
           });
 
-          await new Promise((resolve) => setTimeout(resolve, 100));
           config.data = expectedBody;
 
           config.headers.test = 'static-header-value';
--- a/packages/web-api/src/WebClient.spec.ts
+++ b/packages/web-api/src/WebClient.spec.ts
@@ -2,7 +2,13 @@ import fs from 'node:fs';
 import { assert, expect } from 'chai';
 import nock from 'nock';
 import sinon from 'sinon';
-import { type WebAPICallResult, WebClient, WebClientEvent, buildThreadTsWarningMessage } from './WebClient';
+import {
+  type RequestConfig,
+  type WebAPICallResult,
+  WebClient,
+  WebClientEvent,
+  buildThreadTsWarningMessage,
+} from './WebClient';
 import { ErrorCode, type WebAPIRequestError } from './errors';
 import {
   buildGeneralFilesUploadWarning,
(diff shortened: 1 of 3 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -667,13 +688,11 @@ export class WebClient extends Methods {
...
    * multipart/form-data.
-   * @param options - arguments for the Web API method
-   * @param headers - a mutable object representing the HTTP headers for the outgoing request
+   * @param config - The Axios request configuration object
    */
-  private serializeApiCallOptions(
-    options: Record<string, unknown>,
-    headers?: Record<string, string>,
-  ): string | Readable {
+  private serializeApiCallOptions(config: InternalAxiosRequestConfig): InternalAxiosRequestConfig {
```
filmaj, line 693: A nit here: I see you had to alias the original `options` variable to `data`. I sort of agree that "API call options" makes less sense in this context. To me, request `data` makes more sense than request `options` (and maybe even could be confused with HTTP OPTIONS method). I would be in favour of renaming this private method to `serializeApiCallData` to make this clearer.

packages/web-api/src/WebClient.spec.ts
```diff
@@ -964,6 +971,76 @@ describe('WebClient', () => {
     });
   });
 
+  describe('requestInterceptor', () => {
+    it('can intercept out going requests', async () => {
```
filmaj, line 975: Thanks for adding tests! Let's clarify the intention of this test by expanding on the test title. For someone unfamiliar with request interceptors in axios, what does 'intercepting' a request do? Anyone can figure out that the `requestInterceptor` intercepts requests, but to what to end? Asked differently: what is this test asserting?

packages/web-api/src/WebClient.spec.ts
```diff
@@ -964,6 +971,76 @@ describe('WebClient', () => {
...
+          expect(requestBody).to.deep.equal(expectedBody);
+          return { ok: true, response_metadata: requestBody };
+        });
+
+      await client.apiCall('method');
+    });
+  });
+
+  describe('adapter', () => {
+    it('allows for custom handling of requests', async () => {
```
filmaj, line 1018: This test is insufficient for me to understand the difference between request interceptor and adapter. Let's discuss this feature some more. Why would web-api users want to use adapters over a request interceptor?

filmaj (changes requested):
Thanks so much for the PR! I think this is heading in a great direction. I left many questions for my own education and to help inform my future reviews as well.

One higher-level question: what is the intention behind exposing both an `adapter` option as well as `requestInterceptor`? How/why would consumers use either/or? Would they use both in any situation? The request interceptor makes sense to expose, but based on the axios docs, it seems `adapter` is mainly for use for testing (https://axios-http.com/docs/req_config)?

Also, I think we may need to add something to our docs about this new feature. Looking at the right menu bar for the `web-api` docs [here](https://tools.slack.dev/node-slack-sdk/web-api), perhaps a new section on request interceptors / modifiers would be in order? What do you think?

packages/web-api/src/WebClient.spec.ts
mtjandra: In some cases a web-api user will already have a preconfigured client which they would like to use. They may have a predefined axios instance configured with their desired settings to be used with their internal proxy for all external requests. This client is shared across their project, and is re-usable via their dependency injection framework. Their custom http client may have the following properties:
    - instance already setup to use interceptor for tracing (datadog, new relic, etc.)
    - instance already configured with desired timeout, retries etc.
    - instance may be using using a SDK configured with their internal proxy OpenAPI specs


Yes there is overlap in functionality, but interceptor and adaptor can apply for different use cases. For client who do not have a custom axios instance they want to use, they can just setup an interceptor.

Some user's may want to configure both an interceptor and an adapter based on their way of thinking. For example:
- configure a custom interceptor to shape the desired request
- configure a custom adapter to ensure that client with all desired configuration is in use

packages/web-api/src/WebClient.ts
```diff
@@ -130,6 +136,12 @@ export type PageAccumulator<R extends PageReducer> = R extends (
   ? A
   : never;
 
+export type RequestConfig = InternalAxiosRequestConfig;
```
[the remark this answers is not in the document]
mtjandra: I was not sure what the best practice was here. My thinking was that since `axios` is not set as a peer dependency of the `web-api` package, I did not want clients to import an interface from `axios`, but rather keep the imports contained to the WebClient, as shown in the tests [here](https://github.com/slackapi/node-slack-sdk/pull/2076/files#diff-e07240a5e03c68dd380c2aa81e3b829fd336cadb30770b1c0a8fa19af95e24dcR7)

packages/web-api/src/WebClient.ts
```diff
@@ -257,6 +271,13 @@ export class WebClient extends Methods {
     // serializeApiCallOptions will always determine the appropriate content-type
     this.axios.defaults.headers.post['Content-Type'] = undefined;
 
+    // request interceptors have reversed execution order
```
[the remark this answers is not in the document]
mtjandra: That is correct. Previously when `serializeApiCallOptions` was set by [transformRequest](https://github.com/slackapi/node-slack-sdk/pull/2076/files#diff-a395355361c79a96d5caf3f7ac02c9bf16067d13f93da0fe3f68a0d782f075aeL248) it would always be run at the end. This is not the desired behaviour when configuring an interceptor. We would like the custom interceptor to have the final say on the out going request.

packages/web-api/src/WebClient.spec.ts
mtjandra: Note that `octokit` and `datadog` supports this in their on way.  Here is some pseudo code:

For example with [octokit](https://octokit.github.io/rest.js/v21/#hooks) can do the following

```ts
octokit.hook.wrap('request', async (request, options) => {
  const proxyRequest = transformOutgoinRequests(options)
  const proxyResponse = await customAxiosClient.request(proxyRequest, {
    konnectAccessToken: authToken,
  })

  return proxyResponse
})
```

For [datadog](https://github.com/DataDog/datadog-api-client-typescript/blob/c6d3465bd0aeb8545facee8892a4a30e534cfc46/packages/datadog-api-client-common/http/isomorphic-fetch.ts#L13)
 
 
```ts
export class CustomProxyHttpLib implements DatadogHttpLibrary {
  constructor(
    private readonly customAxiosClient,
  ) {}

  public send(request: client.RequestContext): Promise<client.ResponseContext> {
    const proxyRequest = transformOutgoinRequests(request)
    const proxyResponse = await this.customAxiosClient.request(proxyRequest, {
      konnectAccessToken: authToken,
    })
    
    return transformToDDResponseInterface(proxyResponse)
  }
}
```

packages/web-api/src/WebClient.ts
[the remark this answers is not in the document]
filmaj: Nice, OK, makes sense / sounds good. Let's drop a JSDoc above that to capture this sentiment; most IDEs and our reference generation support linking to pre-existing constructs within JSDoc. Maybe we can add something like the following to help create a papertrail between these various constructs:

```javascript
/**
 * An alias to {@link https://github.com/axios/axios/blob/v1.x/index.d.ts#L367 axios' `InternalAxiosRequestConfig`} object, which is the main parameter type provided to axios interceptors and adapters.
 */
```

Or some such... that will get properly parsed by our documentation generator as well as most IDEs / LSP integrations.

packages/web-api/src/WebClient.spec.ts
filmaj: Thanks for taking the time to explain that! That sounds fine and reasonable to me, esp. if these are already well-established and leveraged axios features.

However, this definitely solidifies the need, in my opinion, to add hand-written docs for this feature as part of the deliverable. Other users will love that, but perhaps you as the PR author will not (as it's more work) 😅 I think adding a section covering request interceptors and adapters to the document under `docs/content/packages/web-api.md`. I'd suggest a new section after the "Custom API URL" section (around line 800 in that document). Doesn't have to be long (you can see the other sections in this document are typically one code example and a couple of paragraphs). In fact, I think what you wrote in your comment above is a great start for such docs content. Couple that with links to the relevant axios docs should be plenty to have this featured covered from a documentation perspective.

packages/web-api/src/WebClient.ts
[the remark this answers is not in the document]
mtjandra: JSDoc has been added

packages/web-api/src/WebClient.ts
```diff
@@ -90,6 +94,8 @@ export interface WebClientOptions {
    * @default true
    */
   attachOriginalToWebAPIRequestError?: boolean;
+  requestInterceptor?: RequestInterceptor;
```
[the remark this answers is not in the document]
mtjandra: JSDoc has been added

packages/web-api/src/WebClient.ts
mtjandra: `options` alias has been removed and private method has been renamed to `serializeApiCallData`

packages/web-api/src/WebClient.spec.ts
mtjandra: I've adjusted the test to better reflect real world use case and added documentation under `docs/content/packages/web-api.md`

packages/web-api/src/WebClient.spec.ts
```diff
@@ -964,6 +971,120 @@ describe('WebClient', () => {
...
+            method: config.method,
+            base_url: config.baseURL,
+            path: config.url,
+            body: config.data ?? {},
+            query: config.params ?? {},
+            headers: structuredClone(config.headers),
+            test: 'static-body-value',
+          });
+
+          await new Promise((resolve) => setTimeout(resolve, 100));
```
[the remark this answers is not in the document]
mtjandra: Nice find, seems like it is not, removed. Also removed the `{ synchronous: true }` flag on the configuration of interceptors, seems to be our safest bet

mtjandra:
> Awesome work! Left a few minor documentation suggestions; once applied or rejected, I will merge this in.
> 
> Thank you _so_ much for your work here and patience with me during review, it is greatly appreciated 🙇

Thank you for helping me with this and being responsive! I've addressed your comments and have rebased with `main`

PR #2081, opened 2024-10-29
ci: add windows testing to CI by lifting OS into its own test matrix …

filmaj:
- Run tests on Windows in CI.
- Tweak a few failing tests on Windows as a result
  - the instrumentation of user agent in `web-api` needs changing, as `process.title` can return a wide variety of results, including it can be completely overridden!
  - some `cli-test` functionality is skipped on Windows (graceful shutting down of processes / lack of interrupt signal support on Windows), so the relevant test executing this branch of code also needs to be skipped on windows.

### DIFF
--- a/packages/cli-test/src/cli/commands/platform.spec.ts
+++ b/packages/cli-test/src/cli/commands/platform.spec.ts
@@ -96,10 +96,12 @@ describe('platform commands', () => {
       sandbox.stub(shell, 'kill').rejects();
       await assert.rejects(platform.runStop({ proc: fakeProcess }));
     });
-    it('should reject if waitForShutdown=true and waitForOutput rejects', async () => {
-      sandbox.stub(shell, 'kill').resolves();
-      waitForOutputSpy.rejects();
-      await assert.rejects(platform.runStop({ proc: fakeProcess, waitForShutdown: true }));
+    it('non-Windows only: should reject if waitForShutdown=true and waitForOutput rejects', async () => {
+      if (process.platform !== 'win32') {
+        sandbox.stub(shell, 'kill').resolves();
+        waitForOutputSpy.rejects();
+        await assert.rejects(platform.runStop({ proc: fakeProcess, waitForShutdown: true }));
+      }
     });
     it('should resolve immediately if waitForShutdown=false and shell.kill resolve', async () => {
       sandbox.stub(shell, 'kill').resolves();
--- a/packages/web-api/src/WebClient.spec.ts
+++ b/packages/web-api/src/WebClient.spec.ts
@@ -113,11 +113,11 @@ describe('WebClient', () => {
       } catch (e) {
         // biome-ignore lint/suspicious/noExplicitAny: TODO: type this better, should be whatever error class web-api throws for timeouts
         const error = e as any;
-        assert.isTrue((logger.warn as sinon.SinonStub).calledOnce, 'expected Logger to be called once');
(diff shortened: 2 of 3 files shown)


packages/cli-test/src/cli/commands/platform.spec.ts
```diff
@@ -96,10 +96,12 @@ describe('platform commands', () => {
       sandbox.stub(shell, 'kill').rejects();
       await assert.rejects(platform.runStop({ proc: fakeProcess }));
     });
-    it('should reject if waitForShutdown=true and waitForOutput rejects', async () => {
-      sandbox.stub(shell, 'kill').resolves();
-      waitForOutputSpy.rejects();
-      await assert.rejects(platform.runStop({ proc: fakeProcess, waitForShutdown: true }));
+    it('non-Windows only: should reject if waitForShutdown=true and waitForOutput rejects', async () => {
+      if (process.platform !== 'win32') {
```
filmaj, line 100: The reason we don't run this test on Windows is that [this code-under-test also doesn't run on Windows](https://github.com/slackapi/node-slack-sdk/blob/main/packages/cli-test/src/cli/commands/platform.ts#L149-L153)

packages/web-api/src/WebClient.spec.ts
```diff
@@ -113,11 +113,11 @@ describe('WebClient', () => {
       } catch (e) {
         // biome-ignore lint/suspicious/noExplicitAny: TODO: type this better, should be whatever error class web-api throws for timeouts
         const error = e as any;
-        assert.isTrue((logger.warn as sinon.SinonStub).calledOnce, 'expected Logger to be called once');
         assert.equal(error.code, ErrorCode.RequestError);
         assert.equal(error.original.config.timeout, timeoutOverride);
         assert.equal(error.original.isAxiosError, true);
         assert.instanceOf(error, Error);
+        assert.isTrue((logger.warn as sinon.SinonStub).calledOnce, 'expected Logger to be called once');
```
filmaj, line 120: Moved this assert down, as it flaked a few times on Windows in CI.. next time it does, I want to make sure these other assertions about the state of the error execute.

packages/web-api/src/WebClient.spec.ts
```diff
@@ -69,7 +69,7 @@ describe('WebClient', () => {
 
     it('should succeed when constructing a class that extends WebClient', () => {
       assert.doesNotThrow(() => {
-        class X extends WebClient {}
+        class X extends WebClient { }
```
mwbrooks, line 72: Nit: Is this whitespace required or common convention (asking because I don't know)?
```suggestion
        class X extends WebClient {}
```

packages/web-api/src/WebClient.spec.ts
```diff
@@ -463,14 +463,13 @@ describe('WebClient', () => {
         const scope = nock('https://slack.com', {
           reqheaders: {
             'User-Agent': (value) => {
+              // User Agent value is different across platforms.
+              // on mac this is: @slack:web-api/7.7.0 node/18.15.0 darwin/23.6.0
+              // on windows this is: @slack:web-api/7.7.0 cmd.exe /22.10.0 win32/10.0.20348
```
mwbrooks, line 468: Thanks for the detailed documentation. These kinda of notes help future maintainers a lot!

mwbrooks (approved):
Thanks for adding Windows testing! Let's give this matrix a run! 🧪 ✨

PR #2082, opened 2024-10-31
feat(cli-test): use powershell.exe as shell on windows

filmaj:
What changes if we use powershell when shelling out to the CLI on Windows? Will cut an RC from this branch to test and experiment.


filmaj:
Closing in favour of #2090

PR #2090, opened 2024-11-01
feat!(cli-test): Use child_process spawn arguments properly, fixing JSON encoding on the command line on Windows

filmaj:
### Summary

(FYI a majority of changes in this PR are relatively cosmetic - will post comments for areas of particular interest for the kind reviewer)

There is a (technically) breaking change in this PR: the string arguments passed to `SlackCLIProcess` should be arranged in an array now, and these are passed as the second argument into `child_process.spawn`. This, in combination with the following changes, also fixes the `datastore` commands on Windows:

- On Windows, shelling out to the CLI process is now wrapped in a `cmd.exe /c /s` invocation, which helps to manage how node will strip quotes out of arguments on Windows when passed to `child_process` APIs.
- Additional post-processing of JSON string arguments is done (currently only in the datastore commands) to ensure double quotes within JSON arguments are escaped.

### DIFF
--- a/packages/cli-test/src/cli/cli-process.ts
+++ b/packages/cli-test/src/cli/cli-process.ts
@@ -49,7 +49,7 @@ export class SlackCLIProcess {
   /**
    * @description The CLI command to invoke
    */
-  public command: string;
+  public command: string[];
 
   /**
    * @description The global CLI options to pass to the command
@@ -61,7 +61,11 @@ export class SlackCLIProcess {
    */
   public commandOptions: SlackCLICommandOptions | undefined;
 
-  public constructor(command: string, globalOptions?: SlackCLIGlobalOptions, commandOptions?: SlackCLICommandOptions) {
+  public constructor(
+    command: string[],
+    globalOptions?: SlackCLIGlobalOptions,
+    commandOptions?: SlackCLICommandOptions,
+  ) {
     if (!process.env.SLACK_CLI_PATH) {
       throw new Error('`SLACK_CLI_PATH` environment variable not found! Aborting!');
     }
@@ -75,7 +79,8 @@ export class SlackCLIProcess {
    */
   public async execAsync(shellOpts?: Partial<SpawnOptionsWithoutStdio>): Promise<ShellProcess> {
     const cmd = this.assembleShellInvocation();
-    const proc = shell.spawnProcess(cmd, shellOpts);
+    // biome-ignore lint/style/noNonNullAssertion: the constructor checks for the truthiness of this environment variable
+    const proc = shell.spawnProcess(process.env.SLACK_CLI_PATH!, cmd, shellOpts);
     await shell.checkIfFinished(proc);
     return proc;
   }
@@ -88,7 +93,8 @@ export class SlackCLIProcess {
     shellOpts?: Partial<SpawnOptionsWithoutStdio>,
   ): Promise<ShellProcess> {
     const cmd = this.assembleShellInvocation();
-    const proc = shell.spawnProcess(cmd, shellOpts);
+    // biome-ignore lint/style/noNonNullAssertion: the constructor checks for the truthiness of this environment variable
+    const proc = shell.spawnProcess(process.env.SLACK_CLI_PATH!, cmd, shellOpts);
     await shell.waitForOutput(output, proc, {
       timeout: shellOpts?.timeout,
     });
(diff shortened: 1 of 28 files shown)


packages/cli-test/src/cli/cli-process.ts
```diff
@@ -49,7 +49,7 @@ export class SlackCLIProcess {
   /**
    * @description The CLI command to invoke
    */
-  public command: string;
```
filmaj, line 52: One part of the breaking change: commands are now composed of an array of strings, representing the space-delimited shell invocation.

packages/cli-test/src/cli/commands/datastore.ts
```diff
@@ -14,6 +14,10 @@ export interface DatastoreCommandArguments {
   queryExpressionValues: object;
 }
 
+function escapeJSON(obj: Record<string, unknown>): string {
+  return `"${JSON.stringify(obj).replace(/"/g, '\\"')}"`;
```
filmaj, line 18: This in combination with using using `child_process` `spawn` et al's `args` array of arguments helped with dealing with JSON encoding of command line arguments in a cross-platform way.

packages/cli-test/src/cli/shell.ts
```diff
@@ -252,3 +249,38 @@ export const shell = {
...
+  command: string,
+  args: string[],
+  env: ReturnType<typeof shell.assembleShellEnv>,
+  shellOpts?: Partial<child.SpawnOptionsWithoutStdio>,
+): [string, string[], child.SpawnOptionsWithoutStdio] {
+  if (process.platform === 'win32') {
+    // In windows, we actually spawn a command prompt and tell it to invoke the CLI command.
+    // The combination of windows and node's child_process spawning is complicated: on windows, child_process strips quotes from arguments. This makes passing JSON difficult.
+    // As a workaround, by telling Windows Command Prompt (cmd.exe) to execute a command to completion (/c) and leave spaces intact (/c), combined with feeding arguments as an argument array into child_process, we can get around this mess.
+    const windowsArgs = ['/s', '/c'].concat([command]).concat(args);
```
filmaj, line 266: This is the root of the Windows compatibility hack

packages/cli-test/src/cli/shell.spec.ts
```diff
@@ -61,34 +61,104 @@ describe('shell module', () => {
...
+      );
     });
     it('should raise bubble error details up', () => {
       runSpy.throws(new Error('this is bat country'));
       assert.throw(() => {
-        shell.runCommandSync('about to explode');
+        shell.runCommandSync('about to explode', []);
       }, /this is bat country/);
     });
+    if (process.platform === 'win32') {
```
filmaj, line 81: This fork in the test code is to test stuff on Windows vs. non-Windows properly, since we now include Windows-specific code.

This also explains why the code coverage dropped: since the code coverage is generated on a GitHub ubuntu runner, it won't exercise the Windows-specific code path, so the coverage reporter will report a lower coverage score. But, since our CI now runs tests on Windows, this code path _is_ exercised (you can see that in the GitHub Action output for the windows check [here](https://github.com/slackapi/node-slack-sdk/actions/runs/11635516867/job/32405055723#step:8:207))

packages/cli-test/src/cli/commands/datastore.spec.ts
```diff
@@ -25,35 +25,37 @@ describe('datastore commands', () => {
...
+        id: '1',
       };
-      await datastore.datastoreQuery({ appPath: '/some/path',  datastoreName: 'datastore', queryExpression: 'id = :id', queryExpressionValues: expressObj});
-      sandbox.assert.calledWith(spawnSpy, sinon.match('datastore query'));
+      await datastore.datastoreQuery({
+        appPath: '/some/path',
+        datastoreName: 'datastore',
+        queryExpression: 'id = :id',
+        queryExpressionValues: expressObj,
+      });
```
mwbrooks, line 57: :bliss: Much more readable

packages/cli-test/src/cli/commands/datastore.ts
mwbrooks: It may be worthwhile leaving a code comment to explain that this is to help support cross-platform JSON encoding

packages/cli-test/src/cli/cli-process.ts
mwbrooks: I think this is a learning for all of our back-pockets. When dealing with CLI arguments in code, it's best to represent them as an array rather than a single string. Command-line frameworks like Cobra also always parse the raw command string into an array as well and it's likely because it's a more reliable way to interact with CLI arguments in code.

packages/cli-test/src/cli/shell.ts
```diff
@@ -252,3 +249,38 @@ export const shell = {
...
+function getSpawnArguments(
+  command: string,
+  args: string[],
+  env: ReturnType<typeof shell.assembleShellEnv>,
+  shellOpts?: Partial<child.SpawnOptionsWithoutStdio>,
+): [string, string[], child.SpawnOptionsWithoutStdio] {
+  if (process.platform === 'win32') {
+    // In windows, we actually spawn a command prompt and tell it to invoke the CLI command.
+    // The combination of windows and node's child_process spawning is complicated: on windows, child_process strips quotes from arguments. This makes passing JSON difficult.
+    // As a workaround, by telling Windows Command Prompt (cmd.exe) to execute a command to completion (/c) and leave spaces intact (/c), combined with feeding arguments as an argument array into child_process, we can get around this mess.
```
mwbrooks, line 265: Really great comment and super helpful for all the maintainers to understand how we need to handle Windows Command Prompt.

Should this be a `/s`?
```suggestion
    // As a workaround, by telling Windows Command Prompt (cmd.exe) to execute a command to completion (/c) and leave spaces intact (/s), combined with feeding arguments as an argument array into child_process, we can get around this mess.
```

mwbrooks (commented):
Pfff... amazing work @filmaj! That's some incredible detective work 🕵🏻 to uncover the right combination of flags to keep JSON intact on Windows Command Prompt.

I've left a few minor comments and suggestions from the code review. I'll let the others throw a proper ✅ on it but it looks good on my side.

packages/cli-test/src/cli/shell.ts
filmaj: It's described in the comment - let me know if you have a suggestion on wording or formatting this differently to make it possibly clearer

vegeris: Oh I see 🤡 `execute a command to completion (/c) and leave spaces intact (/s)`

vegeris:
Just wondering, how were the Windows e2e tests verified? (Are we able to run them in CircleCI with a given version of the cli-test package?)

filmaj:
@vegeris I have a PR using this branch of `cli-test` in the relevant repo up here: https://github.com/slackapi/platform-devxp-test/pull/151

The two most recent CircleCI runs for this branch show both the non-windows (`e2e-test`) and windows (`windows-e2e-test`) results: https://app.circleci.com/pipelines/github/slackapi/platform-devxp-test?branch=cli-test-pshell

The `windows-e2e-test` job I triggered manually via the CircleCI UI from the same link: https://app.circleci.com/pipelines/github/slackapi/platform-devxp-test?branch=cli-test-pshell

packages/cli-test/src/cli/shell.ts
vegeris: I found it really helpful to see it spelled out that this means Windows will execute `cmd.exe /c /s "slack cli command here"` and would suggest that get added to the comment as well; for a while I thought it was `slack \c \s datastore etc.`, which didn't work when I tried it on my Windows laptop 😅

vegeris (approved):
This is great!

I take it we're fine to merge this despite the general restrictions this week as an update to the testing package that won't impact customer-facing behaviour?

PR #2096, opened 2024-11-20
fix(cli-hooks): silence node warnings that can break @slack/cli-hooks

mwbrooks:
### Summary

This pull request updates the `@slack/cli-hooks` to silence the node process warning output.

The reason for this update is that warnings can break some hooks, such as `get-manifest`, because the output must be valid JSON and the warning message is plain-text. For other hooks, such as `doctor` the warning is displayed to the user when the CLI command is executed, which can be confusing. I've left the warning for `start` because the user's code may need to print warning messages to the console.

### Example of a node warning message

This is an example using node v23.2.0:

```bash
$ npx slack-cli-get-hooks

(node:52368) ExperimentalWarning: CommonJS module /Users/michael.brooks/.nvm/versions/node/v23.2.0/lib/node_modules/npm/node_modules/debug/src/node.js is loading ES Module /Users/michael.brooks/.nvm/versions/node/v23.2.0/lib/node_modules/npm/node_modules/supports-color/index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
{"hooks":{"doctor":"NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-doctor","check-update":"NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-check-update","get-manifest":"NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-get-manifest","start":"npx -q --no-install -p @slack/cli-hooks slack-cli-start"},"config":{"watch":{"filter-regex":"^manifest\\.json$","paths":["."]},"protocol-version":["message-boundaries"],"sdk-managed-connection-enabled":true},"runtime":"node"}
```

### Expected `get-hooks` JSON

```json
{
   "config" : {
      "protocol-version" : [
         "message-boundaries"
      ],
      "sdk-managed-connection-enabled" : true,
      "watch" : {
         "filter-regex" : "^manifest\\.json$",
         "paths" : [
            "."
         ]
      }
   },
   "hooks" : {
      "check-update" : "NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-check-update",
      "doctor" : "NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-doctor",
      "get-manifest" : "NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-get-manifest",
      "start" : "npx -q --no-install -p @slack/cli-hooks slack-cli-start"
   },
   "runtime" : "node"
}
```

### Reviewers

```bash
# Change into the cli-hooks package
$ cd node-slack-sdk/packages/cli-hooks

# Install the dependencies
$ npm install

# Run the get-hooks and (optionally) format the JSON
$ npx slack-cli-get-hooks | json_pp

# Expect the JSON response above
# - Confirm `NODE_NO_WARNINGS=1` for `check-update`, `doctor`, and `get-manifest`
# - You may see a warning when you run `get-hooks`, this warning can be safely ignored
#   because it will be silenced in `slack.json`

```

### DIFF
--- a/packages/cli-hooks/src/get-hooks.js
+++ b/packages/cli-hooks/src/get-hooks.js
@@ -45,9 +45,9 @@ if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
 export default function getHooks() {
   return {
     hooks: {
-      doctor: 'npx -q --no-install -p @slack/cli-hooks slack-cli-doctor',
-      'check-update': 'npx -q --no-install -p @slack/cli-hooks slack-cli-check-update',
-      'get-manifest': 'npx -q --no-install -p @slack/cli-hooks slack-cli-get-manifest',
+      doctor: 'NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-doctor',
+      'check-update': 'NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-check-update',
+      'get-manifest': 'NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-get-manifest',
       start: 'npx -q --no-install -p @slack/cli-hooks slack-cli-start',
     },
     config: {
--- a/packages/cli-hooks/src/get-hooks.js
+++ b/packages/cli-hooks/src/get-hooks.js
@@ -43,6 +43,8 @@ if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
  * @returns {SDKInterface} Information about the hooks currently supported.
  */
 export default function getHooks() {
+  // NODE_NO_WARNINGS=1 silences node process warnings. These warnings can occur
+  // on some node version (e.g. v23.2.0) and cause invalid JSON hook responses.
   return {
     hooks: {
       doctor: 'NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-doctor',
--- a/packages/cli-hooks/src/protocols.js
+++ b/packages/cli-hooks/src/protocols.js
@@ -42,7 +42,7 @@ export function DefaultProtocol(args) {
   // If the particular hook invocation is requesting manifest generation we
   // ensure any logging is a no-op to prevent littering stdout with logging
   // and confusing the CLI's manifest JSON payload parsing.
-  const loggerMethod = manifestOnly ? () => { } : console.log;
+  const loggerMethod = manifestOnly ? () => {} : console.log;
   return {
     name: DEFAULT_PROTOCOL,
     log: loggerMethod,
--- a/packages/cli-hooks/README.md
+++ b/packages/cli-hooks/README.md
@@ -76,7 +76,7 @@ Below is an example `slack.json` file that overrides the default `start` hook:
 ```json
 {
   "hooks": {
-    "get-hooks": "npx -q --no-install -p @slack/cli-hooks slack-cli-get-hooks",
+    "get-hooks": "NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-get-hooks",
     "start": "npm run dev"
   }
 }
--- a/packages/cli-hooks/src/get-hooks.spec.js
+++ b/packages/cli-hooks/src/get-hooks.spec.js
@@ -6,9 +6,13 @@ import getHooks from './get-hooks.js';
 describe('get-hooks implementation', async () => {
   it('should return scripts for required hooks', async () => {
     const { hooks } = getHooks();
-    assert(hooks.doctor === 'npx -q --no-install -p @slack/cli-hooks slack-cli-doctor');
-    assert(hooks['check-update'] === 'npx -q --no-install -p @slack/cli-hooks slack-cli-check-update');
-    assert(hooks['get-manifest'] === 'npx -q --no-install -p @slack/cli-hooks slack-cli-get-manifest');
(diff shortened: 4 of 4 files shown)


packages/cli-hooks/src/protocols.js
```diff
@@ -42,7 +42,7 @@ export function DefaultProtocol(args) {
   // If the particular hook invocation is requesting manifest generation we
   // ensure any logging is a no-op to prevent littering stdout with logging
   // and confusing the CLI's manifest JSON payload parsing.
-  const loggerMethod = manifestOnly ? () => { } : console.log;
```
mwbrooks, line 45: Linter auto-pilot

packages/cli-hooks/src/get-hooks.js
```diff
@@ -43,11 +43,13 @@ if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
...
+  // NODE_NO_WARNINGS=1 silences node process warnings. These warnings can occur
+  // on some node version (e.g. v23.2.0) and cause invalid JSON hook responses.
   return {
     hooks: {
-      doctor: 'npx -q --no-install -p @slack/cli-hooks slack-cli-doctor',
-      'check-update': 'npx -q --no-install -p @slack/cli-hooks slack-cli-check-update',
-      'get-manifest': 'npx -q --no-install -p @slack/cli-hooks slack-cli-get-manifest',
+      doctor: 'NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-doctor',
+      'check-update': 'NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-check-update',
+      'get-manifest': 'NODE_NO_WARNINGS=1 npx -q --no-install -p @slack/cli-hooks slack-cli-get-manifest',
```
zimeg, line 52: 🙏 Nice find! I hadn't noticed these warnings in the versions we officially support - `18`, `20`, `22` for now - but think this is a nice and quick guard for those on the cutting edge.

zimeg (approved):
LGTM and thanks for updating the `README` too! 📚 

We can merge this whenever, but I'm hoping we can instead separate `stdout` and `stderr` as part of the default hook protocol on the CLI side of things 👀  IMO debug printing possible warnings would be more ideal for finding odd configurations or other errors that might cause later problems with hooks.

That shouldn't block this, but I'm thinking we can revert these changes if that lands while these warnings are only appearing in unsupported Node versions?

I'm also wondering if we improve testing in later PRs to exercise the hooks in multiple versions of multiple runtimes 😉

mwbrooks:
> We can merge this whenever, but I'm hoping we can instead separate `stdout` and `stderr` as part of the default hook protocol on the CLI side of things 👀 IMO debug printing possible warnings would be more ideal for finding odd configurations or other errors that might cause later problems with hooks.

I'm also very much in favour of updating the Slack CLI to separate `stdout` and `stderr`. We may also want the combined output and possibly the error code. 

It would be nice to have a future where all `stdout` and `stderr` are printed to `--verbose` while select output is displayed in the standard output. This would allow developers to debug while keeping the normal CLI output cleaner.

I also share the concern that meaningful warnings could be suppressed. For example, I think `NODE_NO_WARNINGS=1` will also silence `console.warn(...)` which could be used by developers and the Bolt Framework for debugging.

> That shouldn't block this, but I'm thinking we can revert these changes if that lands while these warnings are only appearing in unsupported Node versions?

I'm in favour of reverting this once we improve the CLI as well!

> I'm also wondering if we improve testing in later PRs to exercise the hooks in multiple versions of multiple runtimes 😉

We should. I imagine the hooks package should support the same node versions that the Slack SDK and Bolt Frameworks support?

packages/cli-hooks/src/get-hooks.js
zimeg: The Node [release schedule](https://nodejs.org/en/about/previous-releases) is so interesting to me!

I believe odd versions are treated as a "current" release and act as an upgrading version to the next even version. These odd versions never get long term support 😅

We might want to discuss testing with the latest current version across the SDKs to make sure bumping to the next current LTS goes smooth.

Supporting that version IMO is another question 🤔

zimeg:
> It would be nice to have a future where all `stdout` and `stderr` are printed to `--verbose` while select output is displayed in the standard output. This would allow developers to debug while keeping the normal CLI output cleaner.

Wow! Mirroring `stdout` in debugs is a fantastic idea! Filtering with the tagged `INFO` or `ERROR` will be so helpful 😭

> I also share the concern that meaningful warnings could be suppressed.

I'm hoping to start a PR that separates these outputs soon! I hadn't thought about the `console.warn` suppressions, but these are the edges I fear so much 😉

> I imagine the hooks package should support the same node versions that the Slack SDK and Bolt Frameworks support

Right now I think we're testing these versions alright, but I don't think we test that an actual `npx slack-cli-get-hooks` call returns just the JSON. That might be difficult to test for some hooks that depend on app setups - `start` - but a single test for `get-hooks` might catch regressions from warnings like this!

**This might also not be a problem after hook enhancements!**

I left another comment about **testing** (not necessarily supporting) the current Node version to make upgrades to the next LTS a bit easier, but that might be discussion for future changes! 🤔

This does seem good to merge as is though! Please of course feel free to tag a release too if you'd like. Or let me know if that's a tag I can make. I still plan to follow up with reversions and revisions in the CLI, but let's not block improvement here 💪

Issue #2094, opened 2024-11-03
WebSocket error occurred: Unexpected server response: 409


gdhardy1:
Socket mode receiver fails to make a successful connection due to `WebSocket error occurred: Unexpected server response: 409` error. Retries after this error occurs may eventually lead to a successful connection or worse, rate limiting from Slack servers on the `apps.connections.open` call.

### `@slack/bolt` version

4.0.1

### Your `App` and Receiver Configuration

Socket mode 2.0.2

### Node.js runtime version

v20.17.0

### Steps to reproduce:

This error seems to reproduce only when launching in a distributed environment (i.e. launching 2 or more instances of the app).
Nothing special is being done on start up. It is minimally reproducible using the example from [BoltJS Getting Started](https://tools.slack.dev/bolt-js/getting-started/) docs as a basis.

```javascript
const app = new App({
  socketMode: true,
  token: process.env.BOT_TOKEN,
  signingSecret: process.env.SIGNING_SECRET,
  appToken: process.env.APP_TOKEN,
  logger
});

void (async () => {
  // Start your app
  await app.start(Number(process.env.PORT) || 3000);

  console.log('⚡️ Bolt app is running!');
})();
```
1. Open multiple shell terminals (suggested 5) and change `process.env.PORT` to a different number in each shell
2. Quickly start the bolt app in each shell (roughly simultaneously to mimic deployment as a server group)

### Expected result:

Each instance of the app should start up successfully without the `WebSocket error occurred: Unexpected server response: 409` error

### Actual result:
First we see the Socket Mode client successfully initialized.
```
[14:05:55.636] DEBUG (36123): Initializing SocketModeReceiver
[14:05:55.636] DEBUG (36123): The logLevel given to Socket Mode was ignored as you also gave logger
[14:05:55.637] DEBUG (36123): The logLevel given to WebClient was ignored as you also gave logger
[14:05:55.637] DEBUG (36123): initialized
[14:05:55.637] DEBUG (36123): The Socket Mode client has successfully initialized
[14:05:55.637] DEBUG (36123): apiCall('auth.test') start
[14:05:55.638] DEBUG (36123): http request url: https://slack.com/api/auth.test
[14:05:55.638] DEBUG (36123): http request body: {"token":"[[REDACTED]]"}
[...]

PR #2099, opened 2024-11-22
socket-mode: Handling WS errors during handshake

filmaj:
This PR fixes https://github.com/slackapi/node-slack-sdk/issues/2094

If a websocket error is raised during WS connection establishment, and retries are on, the library today may start a runaway loop of _doubling_ the number of WS connections as it attempts to retry. This was caused by the `SlackWebSocket` class not tracking a fresh new WS connection until the connection was established; if an error occurred during the handshake, a _new_ WS connection would be created and the old one would not be cleaned up. The core of the fix in this PR involves tracking the WS connection at all times.


packages/socket-mode/src/SlackWebSocket.ts
```diff
@@ -111,29 +111,28 @@ export class SlackWebSocket {
       agent: this.options.httpAgent,
     };
 
-    const ws = new WebSocket(this.options.url, options);
+    this.websocket = new WebSocket(this.options.url, options);
 
-    ws.addEventListener('open', (_event) => {
+    this.websocket.addEventListener('open', (_event) => {
       this.logger.debug('WebSocket open event received (connection established)!');
-      this.websocket = ws;
```
filmaj, line 118: Not tracking the current websocket instance was a problem in the case the web socket server returns an error. Since only once the `open` event was emitted would we keep track of the websocket (this line), any errors during the websocket handshake would call into `disconnect` which, crucially, only terminates the underlying network resources for the tracked `this.websocket`. That was the underlying issue that caused the "doubling up" of reconnection attempts.

packages/socket-mode/src/SocketModeClient.ts
```diff
@@ -229,9 +229,13 @@ export class SocketModeClient extends EventEmitter {
     this.logger.debug(`Before trying to reconnect, this client will wait for ${msBeforeRetry} milliseconds`);
     return new Promise((res, _rej) => {
       setTimeout(() => {
-        this.logger.debug('Continuing with reconnect...');
-        this.emit(State.Reconnecting);
-        cb.apply(this).then(res);
+        if (this.shuttingDown) {
```
filmaj, line 232: While not a problem brought up by users, I noticed a problem here when writing integration tests:

- if you turn on the socket mode client and it is seeing errors from the web socket server, and
- in between reconnect attempts you call `disconnect()` to stop the client,

.. then the client will keep trying to reconnect forever via this part of the code. Adding this guard here just in case.

packages/socket-mode/test/integration.spec.js
```diff
@@ -168,6 +169,47 @@ describe('Integration tests with a WebSocket server', () => {
       assert.isTrue(debugLoggerSpy.calledWith(sinon.match('Unable to parse an incoming WebSocket message')));
       await client.disconnect();
     });
+    it('should maintain one serial reconnection attempt if WSS server sends unexpected HTTP response during handshake, like a 409', async () => {
```
filmaj, line 172: This is the test testing for the exponential reconnection doubling spiral bug.

packages/socket-mode/src/SlackWebSocket.ts
gdhardy1: Thank you for finding this. I was able to patch this in and test it while we await a full fix.

packages/socket-mode/src/SocketModeClient.ts
zimeg: `this` makes for an awesome catch 🙏 ✨  Thanks for guarding against it!

packages/socket-mode/src/SlackWebSocket.spec.ts
```diff
@@ -1,14 +1,29 @@
 import { ConsoleLogger } from '@slack/logger';
 import { assert } from 'chai';
 import EventEmitter from 'eventemitter3';
+import proxyquire from 'proxyquire';
 import sinon from 'sinon';
-
-import { SlackWebSocket } from './SlackWebSocket';
+proxyquire.noPreserveCache();
```
zimeg, line 6: Superb test setups all throughout 👏  This adds so much confidence around the changes.

With some of the retry timing we might have to lookout for flakiness of tests due to the CI runners - nothing a rerun might not solve, but with the matrix setups I think our chances of this happening increase eversomuch 👾

packages/socket-mode/test/integration.spec.js
```diff
@@ -90,7 +90,7 @@ describe('Integration tests with a WebSocket server', () => {
       await client.disconnect();
     });
   });
-  describe('catastrophic server behaviour', () => {
```
zimeg, line 93: 🔥!

packages/socket-mode/test/integration.spec.js
```diff
@@ -168,6 +169,52 @@ describe('Integration tests with a WebSocket server', () => {
...
+          }),
+        );
+      });
+      badServer.listen(WSS_PORT);
+      let closed = 0;
+      // the `close` event is raised every time the websocket server returns an error, so let's keep track of how often this event is emited and use that to infer correct reconnection attempt counts / behaviour
+      client.on('close', () => {
+        closed++;
+      });
+      // do not use await here, since `start()` won't return until the connection is established. we are intentionally testing connection establishment failure, so that will never finish. so, let's run this in a rogue "thread", e.g. fire off an async method and let it do its thing!
```
zimeg, line 201: This is awesome 👏

WilliamBergamin (approved):
This looks good to me as well 💯 

Thank you @zimeg for manually stress testing these changes 🙏 

I tested out our [socket-mode-oauth](https://github.com/slackapi/bolt-js/tree/main/examples/socket-mode-oauth) example and seems like the socket gets established properly 👍

zimeg:
@WilliamBergamin 🙏 And thank you for checking with the example! Let's merge this great fix for the next release 🚀

PR #2117, opened 2024-12-05
ci: check for changes to lints separate from writing changes

zimeg:
### Summary

This PR adds the `lint:fix` job to write changes needed for linting to pass. The `npm run lint` command is changed to just check for changes. That'll cause CI to fail if changes are needed!

This matches the changes discussed in https://github.com/slack-samples/bolt-ts-starter-template/pull/131 and follows https://github.com/slackapi/bolt-js/pull/2357.

#### Before changes

https://github.com/slackapi/node-slack-sdk/actions/runs/12128359302/job/33814593433#step:8:11

#### After changes

Found in the action logs of this PR!

### Requirements


.github/workflows/ci-build.yml
```diff
@@ -26,6 +26,11 @@ jobs:
           - webhook
     runs-on: ${{ matrix.os }}
     steps:
+      - name: Configure git settings (Windows)
+        if: matrix.os == 'windows-latest'
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
```
zimeg, line 33: This step is added to not convert files to ending lines with `\r\n`  after a checkout.

Some searching hints at `.gitattributes` being an option for converting files to `\n` endings before a commit, which might be an option if lintings begin to fail due to files being committed with `\r\n`!

packages/cli-hooks/package.json
```diff
@@ -34,7 +34,8 @@
     "prebuild": "shx rm -rf ./coverage",
     "build": "shx chmod +x src/*.js",
     "prelint": "tsc --noemit --module es2022 --maxNodeModuleJsDepth 0 --project ./jsconfig.json",
-    "lint": "npx @biomejs/biome check --write .",
+    "lint": "npx @biomejs/biome check .",
+    "lint:fix": "npx @biomejs/biome check --write .",
```
zimeg, line 38: This change is matched across all packages using `biome` 🙏

packages/logger/tsconfig.json
```diff
@@ -16,22 +16,16 @@
     "moduleResolution": "node",
     "baseUrl": ".",
     "paths": {
-      "*": [
-        "./types/*"
-      ]
+      "*": ["./types/*"]
```
zimeg, line 19: Other changes are made to fix errors from `npm run lint` 🔍

zimeg (commented):
📝 Leaving a few notes for the kind reviewers!

I was running into interesting strangeness in the CI matrix and left a note on changes from that 👁️‍🗨️

.github/workflows/ci-build.yml
hello-ashleyintech: is there a reason why the files' ending lines are converted to this? I'm guessing we're getting rid of those ending lines because it's clashing with the linter's parsing?

packages/logger/tsconfig.json
hello-ashleyintech: what error would be thrown without this change?

hello-ashleyintech (approved):
PR generally LGTM, thanks for doing this! 🙇 left a few questions mostly for curiosity purposes

.github/workflows/ci-build.yml
zimeg: It seems to be a Windows default as far as I can tell 🤔

Some [suggestions](https://github.com/biomejs/biome/issues/4529#issuecomment-2474905129) add the following to `.gitattributes` to prefer `lf` but that might break editing experiences on certain setups that expect the `\r\n`:

```
* text eol=lf
```

And [some repos](https://github.com/golang/go/blob/5213e1e56c17bde612398dac8a2996bf020fe542/.gitattributes) treat text files as binary to never change endings on checkout:

```
* -text
```

That last example is so interesting to me. The linter does expect `\n` in all cases and was erroring if that wasn't found, and I'm hoping this change in CI is an alright enough workaround for matching expected line endings 👾

packages/logger/tsconfig.json
zimeg:
```
$ npx @biomejs/biome check .
./tsconfig.json format ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Formatter would have printed the following content:

    17 17 │       "baseUrl": ".",
    18 18 │       "paths": {
    19    │ - ······"*":·[
    20    │ - ········"./types/*"
    21    │ - ······]
       19 │ + ······"*":·["./types/*"]
    22 20 │       },
    23 21 │       "esModuleInterop": true
    ····· │
    27 25 │       // "resolveJsonModule": true,
    28 26 │     },
    29    │ - ··"include":·[
    30    │ - ····"src/**/*"
    31    │ - ··],
    32    │ - ··"exclude":·[
    33    │ - ····"src/**/*.spec.*"
    34    │ - ··],
       27 │ + ··"include":·["src/**/*"],
       28 │ + ··"exclude":·["src/**/*.spec.*"],
    35 29 │     "jsdoc": {
    36 30 │       "out": "support/jsdoc",


Checked 9 files in 3ms. No fixes applied.
Found 1 error.
check ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ Some errors were emitted while running checks.


```

An expection of an inlined list. What a beautiful error message this is 🤧

My output even shows `diff` with red and green 🚙 💨  TBH I would've kept the separate lines but I agree with the linter 🙏

zimeg: This is even an error that was changed with automagic 🪄  Quite impressive! ✨

zimeg:
@hello-ashleyintech 🙏 Thank you for another fast review!

I'm glad a few updates from the linter were caught here, and I'm wondering if future updates from the great @dependabot might cause errors here? 👁️‍🗨️ 

For now, I think we're alright to merge these changes, but open to following up with other changes around line endings or similar! 🚢

PR #2118, opened 2024-12-06
Release: @slack/webhook@7.0.4, @slack/web-api@7.8.0

hello-ashleyintech:
### Summary

This PR releases the following new versions:
- @slack/webhook@7.0.4
- @slack/web-api@7.8.0

### DIFF
--- a/packages/webhook/package.json
+++ b/packages/webhook/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@slack/webhook",
-  "version": "7.0.3",
+  "version": "7.1.0",
   "description": "Official library for using the Slack Platform's Incoming Webhooks",
   "author": "Slack Technologies, LLC",
   "license": "MIT",
--- a/packages/webhook/package.json
+++ b/packages/webhook/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@slack/webhook",
-  "version": "7.1.0",
+  "version": "7.0.4",
   "description": "Official library for using the Slack Platform's Incoming Webhooks",
   "author": "Slack Technologies, LLC",
   "license": "MIT",
--- a/packages/web-api/package.json
+++ b/packages/web-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@slack/web-api",
-  "version": "7.7.0",
+  "version": "7.8.0",
   "description": "Official library for using the Slack Platform's Web API",
   "author": "Slack Technologies, LLC",
   "license": "MIT",


packages/webhook/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@slack/webhook",
-  "version": "7.0.3",
+  "version": "7.1.0",
```
zimeg, line 3: We might be alright to keep this as a patch bump after landing [the dependency bump](https://github.com/slackapi/node-slack-sdk/milestone/69?closed=1)?

zimeg (commented):
Quick question before merging!

packages/webhook/package.json
zimeg: I often put package updates as patch changes but please let me know if `semver:minor` is the way here 🙏

hello-ashleyintech: This one is a bit weird because it's a dependency bump, but also supposed to also resolve a security vuln 🤔 But I guess that vuln is considered a bug, so perhaps patch makes more sense here?

hello-ashleyintech: I have pondered this and agree that patch makes more sense, especially when thinking in terms of dependabot PRs. I'll update the webhook milestone!

zimeg: ✨ 🐛 🔍  I like to think a `security` vuln. makes it a fancy patch?

I was thinking about @dependabot too and was also curious how updates have been in past minor vs. patch changes, but [`@slack/webhook`](https://www.npmjs.com/package/@slack/webhook?activeTab=versions) doesn't hint at much here!

zimeg (approved):
Thanks for the fast follows and seeing these updates through! 🙏

zimeg:
The `@slack/oauth` package just popped into mind which [uses](https://github.com/slackapi/node-slack-sdk/blob/96d846a79878f0d33893d14a9942ed9c4f678f2d/packages/oauth/package.json#L35) `@slack/web-api` - bumping `@slack/web-api` might be needed too? 😓

Not sure if this was planned for a separate PR, but I think it could be changed here too. Though separate PRs feels a bit more stable? Your call!

hello-ashleyintech:
@zimeg for public transparency will also leave this comment here- I will merge this PR, release `web-api` and `webhooks`, and then on Monday I'm going to dig into updating those dependencies and releasing - so for Bolt, `socket-mode`, and `oauth` (thus far)! I'll look into other places as well where we may need to update :raised_hands:

PR #2147, opened 2025-02-11
feat(cli-test): include a verbose global option for cli commands

zimeg:
### Summary

This PR adds a **verbose** option to `@slack/cli-test` to add additional output information to specific commands.

### Preview

```diff
    const createOutput = await SlackCLI.create({
      template: 'slack-samples/deno-starter-template',
      branch: 'main',
      appPath,
+     verbose: true
    });
```

### Requirements

### DIFF
--- a/packages/cli-test/src/cli/cli-process.ts
+++ b/packages/cli-test/src/cli/cli-process.ts
@@ -43,7 +43,7 @@ export interface SlackCLIGlobalOptions {
   verbose?: boolean;
 }
 
-export type SlackCLIHostTargetOptions = Pick<SlackCLIGlobalOptions, 'qa' | 'dev' | 'apihost' | 'verbose'>;
+export type SlackCLIHostTargetOptions = Pick<SlackCLIGlobalOptions, 'qa' | 'dev' | 'apihost'>;
 
 export type SlackCLICommandOptions = Record<string, string | boolean | number | undefined>;
 
--- a/packages/cli-test/src/cli/cli-process.ts
+++ b/packages/cli-test/src/cli/cli-process.ts
@@ -43,7 +43,7 @@ export interface SlackCLIGlobalOptions {
   verbose?: boolean;
 }
 
-export type SlackCLIHostTargetOptions = Pick<SlackCLIGlobalOptions, 'qa' | 'dev' | 'apihost'>;
+export type SlackCLIHostTargetOptions = Pick<SlackCLIGlobalOptions, 'qa' | 'dev' | 'apihost' | 'verbose'>;
 
 export type SlackCLICommandOptions = Record<string, string | boolean | number | undefined>;
 
--- a/packages/cli-test/src/cli/cli-process.ts
+++ b/packages/cli-test/src/cli/cli-process.ts
@@ -39,6 +39,8 @@ export interface SlackCLIGlobalOptions {
   team?: string;
   /** @description Access token to use when making Slack API calls. */
   token?: string;
+  /** @description Print debug logging and additional CLI information. */
+  verbose?: boolean;
 }
 
 export type SlackCLIHostTargetOptions = Pick<SlackCLIGlobalOptions, 'qa' | 'dev' | 'apihost'>;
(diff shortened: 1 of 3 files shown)


packages/cli-test/src/cli/cli-process.ts
```diff
@@ -144,6 +146,9 @@ export class SlackCLIProcess {
       if (opts.token) {
         cmd = cmd.concat(['--token', opts.token]);
       }
+      if (opts.verbose) {
+        cmd = cmd.concat(['--verbose']);
```
WilliamBergamin, line 150: Using `concat` alines with the syntax of this function and makes it readable, but looking at this line alone made me ask why `push` was not used instead 🤔

WilliamBergamin (approved):
lgtm thanks for working on this 💯

packages/cli-test/src/cli/cli-process.ts
zimeg: IIRC `push` and `concat` are doing similar things but with different returns and underlying happenings - instead of adding to the existing array, `concat` creates a new one and returns this as a value while `push` returns the new length instead 📚 ✨

FWIW I think `push` is a reasonable substitution, but `concat` gives me confidence in fewer side effects.

zimeg:
@WilliamBergamin and thanks for a fast review!

I'm looking into implementing this in our testing suites now - once I'm feeling good about that I'll merge this and will follow up with a release of [`@slack/cli-test@2.1.0`](https://github.com/slackapi/node-slack-sdk/milestone/127) 🚢 💨

mwbrooks (approved):
👌🏻 Beauty

packages/cli-test/src/cli/cli-process.ts
```diff
@@ -39,9 +39,11 @@ export interface SlackCLIGlobalOptions {
   team?: string;
   /** @description Access token to use when making Slack API calls. */
   token?: string;
+  /** @description Print debug logging and additional CLI information. */
+  verbose?: boolean;
 }
 
-export type SlackCLIHostTargetOptions = Pick<SlackCLIGlobalOptions, 'qa' | 'dev' | 'apihost'>;
+export type SlackCLIHostTargetOptions = Pick<SlackCLIGlobalOptions, 'qa' | 'dev' | 'apihost' | 'verbose'>;
```
zimeg, line 46: 📝 This was added to support the login helper commands:

```diff
    const { authTicket, authTicketSlashCommand, output } = await SlackCLI.auth.loginNoPrompt({
      apihost,
+     verbose: true,
    });
```

zimeg: I'm not thrilled with having this added to the `SlackCLIHostTargetOptions` type, but I think it makes sense as a possible argument.

Super open to suggestions here, but will plan on merging this soon-ish otherwise!

zimeg (commented):
@WilliamBergamin @mwbrooks thank y'all both for the reviews! 🙏 ✨ 

I made one more change after some findings in local testing that would be helpful to include, but I'm not certain on the approach. If this still looks good I'll plan on merging and following up with a release soon!

packages/cli-test/src/cli/cli-process.ts
WilliamBergamin: Hmm this does feel off 🤔 Maybe we can rename `SlackCLIHostTargetOptions` to `AuthLoginNoPromptOptions`

Or create a new type named `AuthLoginNoPromptGlobalOptions` that is `Pick<SlackCLIGlobalOptions, 'verbose'>` and make the args of loginNoPrompt be something like 
```ts
type AuthLoginNoPromptGlobalOptions = Pick<SlackCLIGlobalOptions, 'verbose'>;
type AuthLoginNoPromptOptions = SlackCLIHostTargetOptions & AuthLoginNoPromptGlobalOptions;

loginNoPrompt: async function loginNoPrompt(args?: AuthLoginNoPromptOptions): Promise<{...
```

zimeg: @WilliamBergamin Nice call! When making this update I noticed a few other commands have a similar pattern for specific, additional, arguments.

The d689fa3932a8ec86c429ca188be6f8607a0318a3 change creates similar arguments for each of the `login` helpers, which seems better to me 🎉 

Let me know what you think though! I left the `logout` arguments unchanged, but these don't need to be updated to use `verbose` 😉

WilliamBergamin (approved):
Nice 💯  LGTM

packages/cli-test/src/cli/cli-process.ts
WilliamBergamin: Looks good 💯  ship it!

PR #2148, opened 2025-02-12
fix(cli-test): remove the trailing newline from parsed cli outputs

zimeg:
### Summary

This PR removes the trailing newline found in CLI outputs **before** logging to `error` or `verbose` since the logger adds a newline itself!

No change is made to the captured shell outputs, besides the existing `removeANSIcolors` function, to avoid breaking outputs that expect certain linebreaks 🙏 ✨ 

### Preview

The following example uses changes in #2147 to showcase linebreaks:

#### Before changes

before

#### After changes

after

### Requirements

### DIFF
--- a/packages/cli-test/src/cli/shell.ts
+++ b/packages/cli-test/src/cli/shell.ts
@@ -42,17 +42,17 @@ export const shell = {
       // Listen to data event that returns all the output and collect it
       // biome-ignore lint/suspicious/noExplicitAny: stdout can accept a variety of data
       childProcess.stdout.on('data', (data: any) => {
-        const output = this.removeANSIcolors(data.toString()).replace(/\r?\n$/, '');
+        const output = this.removeANSIcolors(data.toString());
         sh.output += output;
-        logger.verbose(`Output: ${output}`);
+        logger.verbose(`Output: ${output.replace(/\r?\n$/, '')}`);
       });
 
       // Collect error output
       // biome-ignore lint/suspicious/noExplicitAny: stderr can accept a variety of data
       childProcess.stderr.on('data', (data: any) => {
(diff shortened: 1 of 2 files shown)


packages/cli-test/src/cli/shell.ts
```diff
@@ -42,15 +42,17 @@ export const shell = {
       // Listen to data event that returns all the output and collect it
       // biome-ignore lint/suspicious/noExplicitAny: stdout can accept a variety of data
       childProcess.stdout.on('data', (data: any) => {
-        sh.output += this.removeANSIcolors(data.toString());
-        logger.verbose(`Output: ${this.removeANSIcolors(data.toString())}`);
+        const output = this.removeANSIcolors(data.toString());
+        sh.output += output;
+        logger.verbose(`Output: ${output.replace(/\r?\n$/, '')}`);
```
mwbrooks, line 47: Taking it to the next level with `\r?` 🪟

mwbrooks (approved):
💅🏻 So much nicer!

packages/cli-test/src/cli/shell.ts
zimeg: Of course - always keeping the cartridge return in mind! But only sometimes 😉

zimeg:
@mwbrooks Thanks a ton for another fast review!

I noticed test coverage dropped a bit with this initial change, but I'm glad that called out a gap in the tests! A few cases were added to cover how outputs are handled with a focus on these **not** being adjusted (aside from ANSI removals).

With that, I think now's a nice time to merge! I'll follow up later with #2147 and then the promised release and then the actual updates to our tests 🎁 ✨

PR #2176, opened 2025-03-13
feat(web-api): add configs to toggle absolute url usage in dynamic api calls

zimeg:
### Summary

This PR adds a new `allowAbsoluteUrls` option to the `WebClient` constructor that can toggle if the `slackApiUrl` is **always** the prefix of the request URL when calling Slack API methods **or** if an absolute URL can override this.

To keep the current behavior, `allowAbsoluteUrls` defaults to `true` 📚

### Preview

The following code attempts to make a request to "https://slack.com/api/https://example.com/method"!

```javascript
const client = new webapi.WebClient(process.env.SLACK_TOKEN, {
  allowAbsoluteUrls: false,
});

const _response = await client.apiCall("https://example.com/method", { /* ... */ })
```

```
[DEBUG]  web-api:WebClient:0 initialized
[DEBUG]  web-api:WebClient:0 apiCall('https://example.com/method') start
[DEBUG]  web-api:WebClient:0 http request url: https://slack.com/api/https://example.com/method
```

### Reviewers

The above example can be used with Slack API methods in "dot" notation - `chat.postMessage` - and different values of the `allowAbsoluteUrls` option 👾

For fast reference, it might be neat to use the absolute "https://slack.com/api/chat.postMessage" URL.

### Requirements


docs/content/packages/web-api.md
```diff
@@ -131,6 +131,20 @@ const conversationId = '...';
...
+[`allowAbsoluteUrls`](/reference/web-api/interfaces/WebClientOptions#allowabsoluteurls)
+option to `false` sends all requests to the `slackApiUrl` option value:
+
+```javascript
+const { WebClient } = require('@slack/web-api');
+
+const web = new WebClient(token, {
+  allowAbsoluteUrls: false,
+});
+```
```
zimeg, line 146: Open to suggestions as always, but wanted to make sure a note of this option is included in these details 📚 ✨

docs

packages/web-api/src/WebClient.ts
```diff
@@ -618,7 +646,9 @@ export class WebClient extends Methods {
     // TODO: better input types - remove any
     const task = () =>
       this.requestQueue.add(async () => {
-        const requestURL = url.startsWith('https' || 'http') ? url : `${this.axios.getUri() + url}`;
+        const requestURL = url.startsWith('https' || 'http') && this.allowAbsoluteUrls
+          ? url
+          : `${this.axios.getUri() + url}`;
```
zimeg, line 651: 📝 The `requestURL` variable is only used in debug logs, `axios` details configured separately, though this did seem close to what the logic of `allowAbsoluteUrls` might do!

zimeg (commented):
📝 A few quick notes on these changes for the kind reviewers!

docs/content/packages/web-api.md
zimeg: 🔗 And a link for reference: https://tools.slack.dev/node-slack-sdk/web-api#call-a-method

packages/web-api/src/WebClient.ts
```diff
@@ -618,7 +646,8 @@ export class WebClient extends Methods {
     // TODO: better input types - remove any
     const task = () =>
       this.requestQueue.add(async () => {
-        const requestURL = url.startsWith('https' || 'http') ? url : `${this.axios.getUri() + url}`;
+        const requestURL =
+          url.startsWith('https' || 'http') && this.allowAbsoluteUrls ? url : `${this.axios.getUri() + url}`;
```
WilliamBergamin, line 650: The complexity of this line seems to be growing, could it be more readable it we broke it up with an `if` statement 🤔 ?

WilliamBergamin (changes requested):
These changes look good 💯 

I'm just not clear on when a developer would want to do this, why would they be passing and override URL if it will have not effect 🤔 instead of this addition should we advise them to not pass the override value if they don't want to override the URL?

packages/web-api/src/WebClient.ts
zimeg: I was thinking the same... So much is happening here 😓 

I'll explore the `if` statements, but do also like the `const` allowed in the current setup 👾

zimeg:
@WilliamBergamin Thanks for checking out the changes! ✨ 

I don't believe this option is needed for scripts or apps making known API calls, but this option can be useful in setups that might pass unknown methods to `WebClient.apiCall`

The same guards against absolute URLs could definitely be added to application code, prefixing all `.apiCall` requests with the default API URL as you suggest, but IMO this implementation might be forgotten in some places which makes me think having an SDK implementation is preferred 👁️‍🗨️

packages/web-api/src/WebClient.ts
WilliamBergamin: Maybe breaking the logic out into its own function could make sense

```ts
private deriveRequestUrl(url: string): string {
    if (this.allowAbsoluteUrls) {
        return url.startsWith('https' || 'http') ? url : `${this.axios.getUri() + url}`
    }
    return `${this.axios.getUri() + url}`;
}
```

WilliamBergamin (commented):
I see how the changes could be useful for security use cases 🤔  

I've added one more comment about breaking the `requestUrl` logic into a method to increase readability, let me know what you think

packages/web-api/src/WebClient.ts
zimeg: Similar logic was moved into a `deriveRequestUrl` function in 5f98b568289def1af5e8049da62e856f87509e3c! I forget `private` functions can be useful sometimes - it's a great call 😅

zimeg:
@WilliamBergamin 🙏 Thanks for suggesting the improvement to code structures. I'm hoping this'll make later changes around this logic a bit quicker to understand!

Tests continue to pass with that change too, but please let me know if other changes are requested!

WilliamBergamin (approved):
Nice 💯 this looks good

packages/web-api/src/WebClient.ts
```diff
@@ -744,7 +744,8 @@ export class WebClient extends Methods {
    * @param url - The resource to POST to. Either a Slack API method or absolute URL.
    */
   private deriveRequestUrl(url: string): string {
-    if (url.startsWith('https' || 'http') && this.allowAbsoluteUrls) {
+    const isAbsoluteURL = url.startsWith('https://') || url.startsWith('http://');
```
zimeg, line 747: @codecov caught that this check wasn't what we were expecting! 🤖 ✨  

The `if` statement before wasn't taken unless the URL started as `https` - the updated check now checks HTTP separate and uses the `://` incase a collection of `http` Slack methods existed 👁️‍🗨️

zimeg:
@WilliamBergamin Thanks so much calling out the changes of this PR 🙏

I did a bit more testing and found one thing strange with `http` URLs fixed b154b6b910ef11002a206f200c7b055b0072e36c. AFAICT this change only updates the request URL shown in logs since `axios` handles actual request logic, which is why the tests were passing 📺 ✨

Now is seeming like an alright time to merge this! I'll open a release PR for `@slack/web-api@7.9.0` but wait until tomorrow to release and make adjacent updates if this is all alright. Please let me know if other changes are needed before that though! 🔍

PR #2231, opened 2025-04-09
Feat work objects

vegeris:
### Summary

* [ ] I've read and understood the [Contributing Guidelines](https://github.com/slackapi/node-slack-sdk/blob/main/.github/contributing.md) and have done my best effort to follow them.
* [ ] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).


packages/types/src/events/message-metadata.ts
```diff
@@ -21,20 +21,20 @@ export interface MessageMetadataUpdatedEvent {
...
-  metadata: MessageMetadata;
+  metadata: EventMessageMetadata;
 }
 
 export interface MessageMetadataDeletedEvent {
   type: 'message_metadata_deleted';
   channel_id: string;
   event_ts: string;
-  previous_metadata: MessageMetadata;
+  previous_metadata: EventMessageMetadata;
```
vegeris: MessageMetadata has been updated to support both event and entity metadata, but we don't want to surface the entity metadata in these events

packages/types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@slack/types",
-  "version": "2.17.0",
+  "version": "2.17.0-featWorkObjectsBeta.2",
```
zimeg:
```suggestion
  "version": "2.17.0",
```

~🪓 todo: Before merging let's revert this!~

😳 edit: Oop perhaps not for this package!

packages/web-api/src/methods.ts
```diff
@@ -1642,6 +1644,18 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
     update: bindApiCall<ChatUpdateArguments, ChatUpdateResponse>(this, 'chat.update'),
   };
 
+  public readonly entity = {
```
zimeg: 🧮 suggestion: Is it possible to add this after perhaps `emoji`?

packages/web-api/src/types/request/manifest.ts
```diff
@@ -540,4 +540,5 @@ type ManifestEvent =
   | 'workflow_published'
   | 'workflow_step_deleted'
   | 'workflow_step_execute'
-  | 'workflow_unpublished';
+  | 'workflow_unpublished'
+  | 'entity_details_requested';
```
zimeg: 🧮 quibble: Can this too be added earlier in the list for alphabetics?

packages/web-api/src/methods.ts
```diff
@@ -1642,6 +1644,18 @@ export abstract class Methods extends EventEmitter<WebClientEvent> {
...
+    /**
+     * @description
+     * Provide information about the entity to be displayed in the flexpane.
+     * A link to the API documentation will be added when this API is publicly released.
```
zimeg:
```suggestion
     * @description Provide information about the entity to be displayed in the flexpane.
     * @see {@link https://docs.slack.dev/reference/methods/entity.presentDetails}
```

📚 suggestion: Perhaps this is too optimistic but we might want to inline the description before merging?

packages/web-api/src/types/request/chat.ts
```diff
@@ -255,14 +256,9 @@ export interface SourceAndUnfurlID {
...
 
-// https://docs.slack.dev/reference/methods/chat.unfurl
-export type ChatUnfurlArguments = {
-  /**
-   * @description URL-encoded JSON map with keys set to URLs featured in the the message, pointing to their unfurl
-   * blocks or message attachments.
-   */
-  unfurls: LinkUnfurls;
-} & UnfurlTarget &
+// https://api.slack.com/methods/chat.unfurl
```
zimeg:
```suggestion
// https://docs.slack.dev/reference/methods/chat.unfurl
```

🪓 quibble: To prefer the redirect in most cases!

packages/web-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@slack/web-api",
-  "version": "7.11.0",
+  "version": "7.11.0-featWorkObjectsBeta.2",
```
zimeg:
```suggestion
  "version": "7.11.0",
```

🪓 todo: We ought revert this before merge!

packages/web-api/package.json
```diff
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "@slack/logger": "^4.0.0",
-    "@slack/types": "^2.17.0",
+    "@slack/types": "2.17.0-featWorkObjectsBeta.2",
```
zimeg:
```suggestion
    "@slack/types": "^2.18.0",
```

🔮 todo(follow-up): This I think is fine to merge as a prerelease but we should release `@slack/types@2.18.0` and fast follow with such update. I think removing this otherwise will break tests!

packages/web-api/src/types/request/entity.ts
```diff
@@ -0,0 +1,37 @@
+import type { EntityActionButton, EntityMetadata } from '@slack/types';
+import type { TokenOverridable } from './common';
+
+// https://api.slack.com/methods/entity.presentDetails
```
zimeg:
```suggestion
// https://docs.slack.dev/reference/methods/entity.presentDetails
```

📚 quibble: Optimistic linking to perhaps reduce future change?

packages/web-api/src/types/request/entity.ts
```diff
@@ -0,0 +1,37 @@
...
+   * */
+  user_auth_required?: boolean;
+  /**
+   * @description A custom URL to which users are directed for authentication if required.
+   * Example: "https://example.com/onboarding?user_id=xxx"
+   * */
+  user_auth_url?: string;
+  /** @description Error response preventing flexpane data from being returned. */
+  error?: {
+    // Error status indicating why the entity could not be presented.
```
zimeg: 📚 question: Should these also be written in `jsdoc` notation? I am unsure what appears in reference otherwise...

packages/types/src/message-metadata.ts
```diff
@@ -8,15 +11,302 @@ export interface MessageMetadata {
...
+   * @description An array of work object entities.
+   */
+  entities?: EntityMetadata[];
 }
 
+// ------------------------------
+// Event metadata
+// ------------------------------
+
+export interface EventMessageMetadata extends Omit<MessageMetadata, 'entities'> {}
```
zimeg: ⚠️ question: I'm not so familiar with the changes to message metadata, but would composing this interface make the `MessageMetadata` use more clear with something like:

```js
export interface MessageMetadata extends EventMessageMetadata | EntityMetadata {}
```

packages/web-api/test/types/methods/chat.test-d.ts
```diff
@@ -441,6 +441,25 @@ expectAssignable<Parameters<typeof web.chat.postMessage>>([
...
 ]);
+expectAssignable<Parameters<typeof web.chat.postMessage>>([
+  {
+    channel: 'C1234',
+    text: 'hello',
+    thread_ts: '1234.56',
+    metadata: {
+      entities: [
+        {
+          entity_type: EntityType.File,
```
zimeg:
```suggestion
          entity_type: 'slack#/entities/file',
```

suggestion(non-blocking): Unsure if this improves the test, but I find explicit values clear at a glance!

packages/web-api/src/types/request/chat.ts
```diff
@@ -286,6 +282,25 @@ export type ChatUnfurlArguments = {
...
   };
 
+// `unfurls` param of the `chat.unfurl` API
+type BlockKitUnfurls = {
+  /**
+   * @description Object with keys set to URLs featured in the message, pointing to their unfurl
+   * blocks or message attachments.
+   */
+  unfurls: LinkUnfurls;
+};
```
zimeg: ⭐ praise: Thanks for making this more clear in types and documentation!

packages/types/src/message-metadata.ts
zimeg: 🐔 ramble: This makes me wonder too if `MessageMetadata` should remain unchanged and instead a new union is added for both "entity" and traditional message metadata, but please let me know if this doesn't make sense!

🦠 thought: I wonder this since some downstream packages might be expecting the current strict typing but perhaps this isn't a concern after all...

https://github.com/slackapi/node-slack-sdk/blob/ff03f7812c678bdc5cea5eace75db34631a88dda/packages/webhook/src/IncomingWebhook.ts#L116-L120

packages/types/src/message-metadata.ts
```diff
@@ -8,15 +11,302 @@ export interface MessageMetadata {
...
+  /**
+   * @description Entity type.
+   */
+  entity_type: EntityType | string;
+  /**
+   * @description Schema for the given entity type.
+   */
+  entity_payload: {
+    attributes: EntityAttributes;
+    fields?: FileEntityFields | TaskEntityFields | IncidentEntityFields | ContentItemEntityFields;
```
zimeg: ⭐ praise: I'm a fan of this union implementation with the enum for `entity_type` above!

zimeg: 👁️‍🗨️ question: For my own understanding, does "slack#/entities/item" require a `fields` type?

zimeg: 🧮 suggestion(non-blocking): I might kind request alphabetics for these orderings too-

packages/types/src/message-metadata.ts
zimeg: 🪬 ramble: A quick finding that the exported `MessageMetadata` appears as input to other methods too-

https://github.com/slackapi/node-slack-sdk/blob/ff03f7812c678bdc5cea5eace75db34631a88dda/packages/web-api/src/types/request/chat.ts#L209

- `chat.scheduleMessage`
- `chat.stopStream`
- `chat.chat.update`

packages/web-api/src/types/request/chat.ts
vegeris: Thanks!

packages/types/src/message-metadata.ts
vegeris: > does "slack#/entities/item" require a fields type?

It does not 🙂 'item' is the generic type so we don't define any common fields 

> alphabetics for these orderings too

Updated ✅

packages/web-api/test/types/methods/chat.test-d.ts
vegeris: No strong feelings; this was demonstrating that the enum could be used here

packages/types/src/message-metadata.ts
vegeris: Good callout! This is definitely one of the areas I wanted to get feedback on. I originally updated `MessageMetadata` to support both event and entity metadata because the naming seemed most appropriate, being more generic. But there's that risk of breaking something by changing the name of the expected type for these APIs/events that don't support entity metadata to `EventMessageMetadata` (this would apply to the APIs listed above, I didn't catch those!)

It sounds like the naming isn't a big factor so I'm down to keep `MessageMetadata` unchanged and introduce a separate union type for event + entity metadata!

Edit: Ah, I guess if we have a separate type, the input to chat.postMessage would be affected rather than those other APIs. I guess either way, the metadata input type to one or more currently used APIs would change

packages/web-api/src/types/request/entity.ts
vegeris: Whoops! Yes it should 🙂

packages/types/src/message-metadata.ts
vegeris: Update in progress (keeping MessageMetadata unchanged) 🚧

packages/web-api/src/types/request/chat.ts
```diff
@@ -286,6 +283,29 @@ export type ChatUnfurlArguments = {
...
+/**
+ * @description The `metadata` param of the `chat.unfurl` API.
+ */
+type EntityMetadataUnfurls = {
+  /**
+   * @description Unfurl metadata featuring an array of entities to attach to the message based on URLs featured in the message.
+   */
+  metadata: Partial<MessageMetadata> & {
+    entities: EntityMetadata[];
+  };
```
vegeris, line 318: (One can also provide event metadata in chat.unfurl and it'll be persisted on the message but it's not a main use case)

packages/web-api/src/types/request/chat.ts
```diff
@@ -191,7 +193,7 @@ export type ChatPostMessageArguments = TokenOverridable &
   Authorship &
   Parse &
   LinkNames &
-  Metadata &
+  EntityAndEventMessageMetadata &
```
vegeris: Updated chat.postMessage input; it now accepts both event and entity data (both can be supplied, or one or the other) 

(but other APIs still only accept event metadata)

packages/types/src/message-metadata.ts
vegeris: I refactored to add a separate `EntityAndEventMessageMetadata` object that can be consumed in places where both event and entity metadata are supported (but both are optional), like chat.postMessage

(beta release for the types package needs to updated for the tests to pass in web-api)

vegeris: Released new beta versions for the types and web-api packages ✅

packages/web-api/package.json
vegeris: Ah, good point 😅 So if I understand correctly, we would be okay to merge this PR with the web-api package pointing to the types package pre-release, and as a fast follow we would release a minor version of the types package, and then the web-api with that dependency updated (after the work objects Houston toggle is full released)

packages/web-api/src/types/request/chat.ts
```diff
@@ -191,7 +197,7 @@ export type ChatPostMessageArguments = TokenOverridable &
   Authorship &
   Parse &
   LinkNames &
-  Metadata &
+  ChatPostMessageMetadata &
```
WilliamBergamin, line 208: ❓ Just want to validate that I'm understanding correctly, this updates `ChatPostMessageArguments` from 

```ts
interface ChatPostMessageArguments {
  ...
  metadata?:  {
    event_type: string;
    event_payload: {
      [key: string]: string | number | boolean | MessageMetadataEventPayloadObject | MessageMetadataEventPayloadObject[];
    };
}
```
To 
```ts
interface ChatPostMessageArguments {
  ...
  metadata?:  {
    event_type?: string;
    event_payload?: {
      [key: string]: string | number | boolean | MessageMetadataEventPayloadObject | MessageMetadataEventPayloadObject[];
    };
    entities?: EntityMetadata[];
}
```

Since this type is used to define and input for a function, making `event_type` and `event_payload` optional should not be a breaking change?

packages/types/src/message-metadata.ts
```diff
@@ -20,3 +26,287 @@ export interface MessageMetadata {
...
+// ------------------------------
+// Work object metadata
+// ------------------------------
+
+/**
+ * @description Application-specific entity and/or event data to attach to a Slack message.
+ * Provide 'entities' to set work object entity metadata.
+ * Provide 'event_type' and 'event_payload' to set event metadata.
+ */
+export interface EntityAndEventMessageMetadata extends Partial<MessageMetadata> {
```
WilliamBergamin: Since this seems to only be used to define `ChatPostMessageMetadata` in `packages/web-api/src/types/request/chat.ts`, what would you think about removing `EntityAndEventMessageMetadata` and instead defining it in `ChatPostMessageMetadata`, this would follow a similar pattern as `EntityMetadataUnfurls`

```ts
export interface ChatPostMessageMetadata {
  /**
   * @description Object representing message metadata, entity and/or event data to attach to a Slack message.
   * Provide 'entities' to set work object entity metadata.
   * Provide 'event_type' and 'event_payload' to set event metadata.
   */
  metadata?: Partial<MessageMetadata> & {
    /**
     * @description An array of work object entities.
     */
    entities?: EntityMetadata[];
  };
}
```

packages/web-api/src/types/request/chat.ts
WilliamBergamin: ++ on praise for this 🙏

packages/web-api/src/types/request/chat.ts
```diff
@@ -104,6 +106,10 @@ export interface BroadcastedThreadReply extends ThreadTS {
 // or not broadcasted. Broadcasted replies are necessarily threaded, so `thread_ts` becomes required.
 type ReplyInThread = WithinThreadReply | BroadcastedThreadReply;
 
+export interface ChatPostMessageMetadata {
+  /** @description Object representing message metadata, which will be made accessible to any user or app. */
+  metadata?: EntityAndEventMessageMetadata;
+}
```
WilliamBergamin, line 120: This is related to another comment, on defining `EntityAndEventMessageMetadata` inside `ChatPostMessageMetadata`, this would also align with the patterns used by `EntityMetadataUnfurls`
```suggestion
export interface ChatPostMessageMetadata {
  /**
   * @description Object representing message metadata, entity and/or event data to attach to a Slack message.
   * Provide 'entities' to set work object entity metadata.
   * Provide 'event_type' and 'event_payload' to set event metadata.
   */
  metadata?: Partial<MessageMetadata> & {
    /**
     * @description An array of work object entities.
     */
    entities?: EntityMetadata[];
  };
}
```

packages/web-api/src/types/request/chat.ts
WilliamBergamin: QQ: In `ChatPostMessageMetadata`, `entities` is optional, just want to confirm that here is should be required

packages/web-api/src/types/request/chat.ts
```diff
@@ -286,6 +287,29 @@ export type ChatUnfurlArguments = {
...
+   * @description Object with keys set to URLs featured in the message, pointing to their unfurl
+   * blocks or message attachments.
+   */
+  unfurls: LinkUnfurls;
+};
+
+/**
+ * @description The `metadata` param of the `chat.unfurl` API.
+ */
+type EntityMetadataUnfurls = {
```
WilliamBergamin: Thanks for breaking this up into its own entity 🚀 

To align with the patterns from `Metadata` and `ChatPostMessageMetadata`, what do you think about naming this `ChatUnfurlMetadata` and making it an interface? 🤔

 ```suggestion
interface ChatUnfurlMetadata {
```

**Review by WilliamBergamin — COMMENTED** [review]
Great work putting all this together 🙏  types are hard!!

I've left a few comment and questions let me know what you think

File: packages/web-api/src/types/request/chat.ts (hunk 15 above)
Comment (vegeris) [reply packages/web-api/src/types/request/chat.ts:318]: Yeah, if `metadata` is provided to `chat.unfurl`, `entities` is required. For `chat.postMessage`, `metadata` can be provided with event metadata only

File: packages/web-api/src/types/request/chat.ts (hunk 17 above)
Comment (vegeris) [reply packages/web-api/src/types/request/chat.ts:208]: Yup; what I'm trying to show is that one can provide event and / or entity metadata in `chat.postMessage`

```
interface ChatPostMessageArguments {
  ...
  metadata?:  {

   // if providing event metadata, both of these params are still required (but they're optional in the type bc one can provide entity metadata instead)
    event_type: string;
    event_payload: {
      [key: string]: string | number | boolean | MessageMetadataEventPayloadObject | MessageMetadataEventPayloadObject[];
    };


AND / OR

   // if providing entity metadata, this param is required (but it's optional in the type bc one can provide event metadata instead)
    entities: EntityMetadata[];
}
```

> Since this type is used to define and input for a function, making event_type and event_payload optional should not be a breaking change?

Passing an object into the metadata param with just event_type and event_payload is still valid and should still be accepted by the SDK; if users are typing the object as `Metadata` though they would need to change that to `ChatPostMessageMetadata`. Similar thing came up for the [Java update](https://github.com/slackapi/java-slack-sdk/pull/1512#discussion_r2441425964); changing the type of an existing field does seem like a breaking change 😅 Will work object support require a major version bump because of this?

File: packages/types/src/message-metadata.ts (hunk 18 above)
Comment (vegeris) [reply packages/types/src/message-metadata.ts]: It's currently only used in the input for `chat.postMessage` but I was thinking that we might want to support entity metadata in other places in the future. If we do want to add that support though, new types could be defined specific to those APIs or events (which would be consistent with how the inputs for `chat.unfurl` are defined). I'll move this into `packages/web-api/src/types/request/chat.ts`

File: packages/web-api/src/types/request/chat.ts (hunk 20 above)
Comment (vegeris) [reply packages/web-api/src/types/request/chat.ts]: Sure; with regard to interface vs. type I'm not clear on when we would prefer one over the other aside from when we need to eg. use a type for a union (I see a mix of both in these files)

Comment (vegeris) [reply packages/web-api/src/types/request/chat.ts]: For consistency I could also rename `type BlockKitUnfurls` to `interface ChatUnfurlUnfurls` (lol, since `unfurls` is the param name)

**Comment by vegeris** [comment]
I'm not sure what's going on with the failing tests in the cli-test package;
```
 Exception during run: src/cli/shell.spec.ts(21,55): error TS2345: Argument of type 'SpawnSyncReturns<Buffer<ArrayBufferLike>>' is not assignable to parameter of type 'SpawnSyncReturns<string | NonSharedBuffer>'.
...
```

It seems like the object being passed in matches the [interface](https://docs.deno.com/api/node/child_process/~/SpawnSyncReturns)

```
    runOutput = { pid: 1337, output: [], stdout: Buffer.from("[]"), stderr: Buffer.from("[]"), status: 0, signal: null };
    runSpy = sandbox.stub(child, 'spawnSync').returns(runOutput);
```

Edit: resolved!

File: packages/web-api/src/types/request/chat.ts (hunk 20 above)
Comment (vegeris) [reply packages/web-api/src/types/request/chat.ts]: (Updated!)

File: packages/web-api/src/types/request/chat.ts (hunk 17 above)
Comment (WilliamBergamin) [reply packages/web-api/src/types/request/chat.ts:208]: >  Will work object support require a major version bump because of this?

The the Java SDK I need to take a look there might be a way around this

File: packages/web-api/package.json [hunk 21]
```diff
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "@slack/logger": "^4.0.0",
-    "@slack/types": "^2.17.0",
+    "@slack/types": "2.18.0",
```
WilliamBergamin: Nice 💯

WilliamBergamin:
```suggestion
    "@slack/types": "^2.18.0",
```

WilliamBergamin (changes requested):
Quick change actually

packages/web-api/package.json
vegeris: Thanks!

PR #2302, opened 2025-07-19
chore: adds webhook package to dependabot.yml

lukegalbraithrussell:
### Summary

This adds the webhook package to dependabot.yml. Expect lots of dev dependencies updating this week 🥳 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -140,14 +140,6 @@ updates:
   - package-ecosystem: "npm"
     directory: "/packages/webhook"
     groups:
-      slack:
-        patterns:
-          - "@slack/*"
-      dev-sinon:
-        dependency-type: "development"
-        patterns:
-          - "@types/sinon"
-          - "sinon"
       dev-chai:
         dependency-type: "development"
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -138,28 +138,6 @@ updates:
       interval: "weekly"
 
   - package-ecosystem: "npm"
-    directories:
-      - "/"
-      - "/docs"
-      - "/examples/*"
-    groups:
-      docusaurus:
-        patterns:
-          - "@docusaurus/*"
-          - "docusaurus-plugin-typedoc"
-          - "typedoc"
-          - "typedoc-plugin-markdown"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -137,6 +137,30 @@ updates:
...
+    directory: "/packages/webhook"
+    groups:
+      slack:
+        patterns:
+          - "@slack/*"
+      dev-sinon:
+        dependency-type: "development"
+        patterns:
+          - "@types/sinon"
+          - "sinon"
```
zimeg:
```suggestion
```

🪓 We might've found a way to test without this in the more recent changes:

https://github.com/slackapi/node-slack-sdk/blob/d83f4809d6c1c51577222f2ad83259674d7b1c35/packages/webhook/package.json

.github/dependabot.yml
```diff
@@ -137,6 +137,30 @@ updates:
     schedule:
       interval: "weekly"
 
+  - package-ecosystem: "npm"
+    directory: "/packages/webhook"
+    groups:
+      slack:
+        patterns:
+          - "@slack/*"
```
zimeg:
```suggestion
```

🪓 Removing this is no blocker, but we depend on just `@slack/types` for webhooks!

zimeg (approved):
@lukegalbraithrussell Thank you so much for catching this! 🤖 ✨ 

I left a few comments on `groups` we can perhaps remove with these package updates, though these aren't big blockers.

I'm now also curious about a `labels` tag for updates to packages and examples, but we should save this for a follow up: 😉 

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#labels--

zimeg:
🗣️ The updates to come are exciting for me also - we should be developing with the latest package versions I think.

zimeg:
@lukegalbraithrussell Thanks again for making these changes!

Some updates might bring breaking changes to our tests, which I'm open at the start of next week to help with, but please feel free to merge when the time is right 🚢 💨

zimeg:
@lukegalbraithrussell The start of this week has arrived and I think now's the solid time for these kind updates 🎁 ✨

PR #2308, opened 2025-07-22
Docs: autogenerated typedoc reference 

lukegalbraithrussell:
### Summary

Previously, we used a docusaurus typedoc plugin that was hard to navigate with our multiple packages. It also doesn't work for our planned setup of only keeping markdown files in this repo. 

each package has a `docs` script in its `package.json` that generates the reference docs for that package. 

This PR does not change the current reference; it's preparing for our unified docs site. 

TODO:
* webhook reference (waiting on `typescript` dependency bump)

After merge
* add `docs` script to site build workflow
* figure out if we want to check in the reference docs or not


.github/maintainers_guide.md
```diff
@@ -50,38 +52,40 @@ Releasing can feel intimidating at first, but rest assured: if you make a mistak
 
 4. For each package to be released, run `npm run test` to verify that tests are passing and code is free of linting errors.
 
-5. On your new branch, bump the version(s) in `package.json` (see [Versioning and Tags](https://github.com/slackapi/node-slack-sdk/blob/main/.github/maintainers_guide.md#-versioning-and-tags))
+5. Generate the reference docs for that package by running `npm run docs`.
```
lukegalbraithrussell, line 55: I think this makes the most sense since the maintainer will be in the proper package subdirectory from the previous step, but may be wrong!

docs/sidebars.js
```diff
@@ -63,98 +63,16 @@ const sidebars = {
...
-            type: 'doc',
-            id: 'reference/webhook/index',
-          },
-          items: require('./content/reference/webhook/typedoc-sidebar.cjs'),
-        },
-      ],
+      {
+        type: 'autogenerated',
+        dirName: 'reference',
+      },
```
zimeg, line 74: ⭐ Automagic all around!

packages/web-api/typedoc.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "https://typedoc.org/schema.json",
+  "entryPoints": ["src/index.ts"],
+  "plugin": ["typedoc-plugin-markdown"],
+  "outputs": [
+    {
+      "name": "markdown",
+      "path": "../../docs/content/reference/webhooks"
```
zimeg:
```suggestion
      "path": "../../docs/content/reference/web-api"
```

👁️‍🗨️ Small change perhaps?

packages/cli-test/typedoc.json
```diff
@@ -0,0 +1,25 @@
...
+  "useCodeBlocks": true,
+  "logLevel": "Error",
+  "readme": "none",
+  "gitRevision": "main",
+  "excludePrivate": true,
+  "excludeProtected": true,
+  "excludeInternal": true,
+  "hideGenerator": true,
+  "jsDocCompatibility": true,
+  "titleLink": false,
```
zimeg:
```suggestion
  "titleLink": "",
```

No blocker but this might be expecting a string! 🧵

I'm not noticing a difference between an empty string and `false` though 😉

packages/cli-test/package.json
```diff
@@ -31,10 +31,13 @@
     "lint:fix": "npx @biomejs/biome check --write .",
     "prepare": "npm run build",
     "mocha": "cross-env SLACK_CLI_PATH=/doesnt/matter mocha --config ./test/.mocharc.json src/*.spec.ts src/**/*.spec.ts src/**/**/*.spec.ts",
-    "test": "npm run lint && npm run build && c8 --config ./test/.c8rc.json npm run mocha"
+    "test": "npm run lint && npm run build && c8 --config ./test/.c8rc.json npm run mocha",
+    "docs": "npx typedoc --plugin typedoc-plugin-markdown"
```
zimeg: 🔍 Nit to prefer scripts are kept alphabetical, but no blocker!

packages/cli-test/typedoc.json
```diff
@@ -0,0 +1,25 @@
...
+    {
+      "name": "markdown",
+      "path": "../../docs/content/reference/cli-test"
+    }
+  ],
+  "skipErrorChecking": true,
+  "useCodeBlocks": true,
+  "logLevel": "Error",
+  "readme": "none",
+  "gitRevision": "main",
```
zimeg, line 15: 🗣️ This seems reasonable to me for links that match:

```
Defined in: [src/WebClient.ts:133](https://github.com/slackapi/node-slack-sdk/blob/main/packages/web-api/src/WebClient.ts#L133)
```

Updating this to use the package version might be useful too since we're running this with the release process, but we can revisit this later I think:

https://typedoc.org/documents/Options.Input.html#includeversion

```
Defined in: [src/WebClient.ts:133](https://github.com/slackapi/node-slack-sdk/blob/49f8acc669f55335541561c50559aaf769d1efff/packages/web-api/src/WebClient.ts#L133)
```

packages/cli-test/package.json
```diff
@@ -31,10 +31,13 @@
...
   },
   "dependencies": {
     "tree-kill": "^1.2.2",
+    "typedoc": "^0.28.7",
+    "typedoc-plugin-markdown": "^4.7.1",
```
zimeg:
```suggestion
```

@lukegalbraithrussell Is it possible to save these as `devDependencies` instead with:

```
$ npm install --save-dev typedoc typedoc-plugin-markdown
```

We might want to avoid packaging these with the rest of the package on release if possible 👁️‍🗨️

zimeg (changes requested):
@lukegalbraithrussell Super impressive changeset and I hold excitement! 📚 ✨ 

Requesting a few changes before merging this but am wanting to hear what you think about a few things too:

- Can `typedoc` be kept with development dependencies?
- Do we want to use versioned permalinks for code references?

Some other suggestions are more preference but please do keep me updated 👾

packages/cli-test/package.json
lukegalbraithrussell: yep!

packages/cli-test/typedoc.json
lukegalbraithrussell: the one thing is that this would make bookmarks to those links out of date without the user always noticing (or maybe that's a me thing)

packages/cli-test/package.json
lukegalbraithrussell: fixed!

packages/cli-test/typedoc.json
lukegalbraithrussell: i'm going to just remove this; this was part of me trying every single param to hide some top-matter that wasn't going away

lukegalbraithrussell:
@zimeg thanks for your feedback! Fixed the issues

packages/cli-test/typedoc.json
zimeg: Hmm I was imagining bookmarks would be more common to a page like:

- https://tools.slack.dev/node-slack-sdk/reference/web-api/classes/WebClient

Which will be refreshed with each tagged version 🚀

Between versions the code referenced on a specific line of `main` might change without the reference in docs updating otherwise, and might cause these links to point to missing lines or files altogether!

That's a bit more fragile to me and might be confusing if the reference links don't line up with the tagged version, but that might also be motivation to release more often... 👁️‍🗨️

lukegalbraithrussell: oh! i see. okay i'm sold

PR #2313, opened 2025-07-24
docs: add generated reference files

lukegalbraithrussell:
### Summary

This is a follow up to #2308 . adds the actual reference files by deleting the line in .gitignore


docs/content/reference/web-api/interfaces/TeamDomainChangeEvent.md
zimeg, line 1: 👁️‍🗨️ Unrelated to this PR, but this file makes it clear that keeping dependencies at the latest versions is meaningful for reference.

A change landed in `@slack/types@2.15.0` that updated this from `team_domain_changed` to be the expected [`team_domain_change`](https://docs.slack.dev/reference/events/team_domain_change) event and at the moment `@slack/web-api` uses:

https://github.com/slackapi/node-slack-sdk/blob/22117c2cac162c51dc938881d60f7ae9a9833176/packages/web-api/package.json#L52

zimeg (approved):
@lukegalbraithrussell LGTM! And an incredible thanks for tackling this.

With quick reading this all is so solid but I remain open to follow ups as needed for both `jsdoc` and file organization alike. For now I think it'd be awesome to ship this 🚢 💨

docs/content/reference/cli-test/variables/SlackProduct.md
```diff
@@ -0,0 +1,47 @@
+[@slack/cli-test](../index.md) / SlackProduct
+
+# Variable: SlackProduct
+
+```ts
+const SlackProduct: object;
+```
+
+Defined in: [utils/constants.ts:52](https://github.com/slackapi/node-slack-sdk/blob/main/packages/cli-test/src/utils/constants.ts#L52)
```
zimeg, line 9: 👁️‍🗨️ This link might be finding `main` instead of a specific commit, though I do think we can revisit this for neartime releases!

Edit: Added relevant `typedoc` configs following 🔍 

https://github.com/slackapi/node-slack-sdk/blob/22117c2cac162c51dc938881d60f7ae9a9833176/packages/cli-test/typedoc.json#L15-L16

PR #2315, opened 2025-07-29
feat: add custom manifest filename ability.

malewis5:
### Summary

Allow developers to configure the name of their local manifest.json file. This allows for them to have things like: manifest.json, manifest.local.json, etc. 

This makes it easier to have your codebase be a source of truth for the app manifest when working on multiple branches/apps (dev, staging, prod, etc.).

### DIFF
--- a/packages/cli-hooks/src/get-manifest.js
+++ b/packages/cli-hooks/src/get-manifest.js
@@ -24,11 +24,35 @@ if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
  * @returns {object} Parsed values of the project manifest.
  */
 export default function getManifestData(searchDir) {
-  const manifestJSON = readManifestJSONFile(searchDir, 'manifest.json');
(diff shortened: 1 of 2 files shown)


WilliamBergamin:
The [CLI](https://github.com/slackapi/slack-cli) interacts with other packages like [Bolt python](https://github.com/slackapi/bolt-python) and the [Deno SDK](https://github.com/slackapi/deno-slack-sdk) similarly to how it interacts with this one. Before we introduce changes like these we need to ensure the solution can be applied across all packages. 

Unfortunately, I don't think we're ready to review this PR at this stage. We may want to reconsider how these values are passed to the hooks, placing them in a configuration file comes with its own set of complexities and may not yield the best developer experience.

I've created #2317 in order to keep track, discuss and align on a solution for this feature request

Overall this is a neat prototype @malewis5 thanks for putting it together 💯

malewis5:
Sounds good, thank you @WilliamBergamin! I'll continue to follow along.

PR #2355, opened 2025-08-28
cli-test(fix): wait for cli run start trace instead of activity output

zimeg:
### Summary

This PR uses `SLACK_TRACE_PLATFORM_RUN_START` variable as a signal of an active running app instead of the exact output:

> Connected, awaiting events

Fixes an issue where `bolt` apps were not detected as running.

### Requirements


packages/cli-test/src/cli/commands/platform.test.ts
```diff
@@ -86,6 +87,12 @@ describe('platform commands', () => {
         sinon.match.string,
         sinon.match.array.contains(['run', '--cleanup', '--hide-triggers']),
       );
+      sandbox.assert.calledWith(
+        waitForOutputSpy,
+        SlackTracerId.SLACK_TRACE_PLATFORM_RUN_START,
```
zimeg, line 93: 🗣️ note: This is preferred over `SLACK_TRACE_PLATFORM_RUN_READY` since the "ready" variable appears once activity logs stream for CLI managed connections. It does not appear for `bolt` apps at this time.

👁️‍🗨️ thought: I'm unsure if we should add that trace to the CLI outputs in a follow up PR? For now I think using **a** trace is still an improvement but I'm somewhat worried about tests skipping to the next step before `run` is settled for some apps... 🦕

zimeg (commented):
👾 Rambling thoughts on this change but I am open to follow up or updates otherwise!

ewanek1:
LGTM! Like the more reliable detection of running Slack apps 👏 👏

packages/cli-test/src/cli/commands/platform.test.ts
mwbrooks: > I'm unsure if we should add that trace to the CLI outputs in a follow up PR?

Just checking if I understand correctly - are you thinking that the CLI would print the `SLACK_TRACE_PLATFORM_RUN_READY` trace for sdk-managed socket connections?

mwbrooks (approved):
✅ Very healthy fix for Bolt tests! Sorry that this review took 3 weeks! :shame:

zimeg:
@ewanek1 @mwbrooks Thanks both for the kind reviews. And never worries on a rushed review for this! 👾 

> Just checking if I understand correctly - are you thinking that the CLI would print the `SLACK_TRACE_PLATFORM_RUN_READY` trace for sdk-managed socket connections?

Exactly this! I'm not finding such printed at this time, but I'm also not sure if that'd be possible to know for SDK managed connections...

Somewhat following that, I'm hesitant to release this change since it might disrupt our E2E tests if workflows are started **after** "PLATFORM_RUN_START" and **before** "PLATFORM_RUN_READY" but we can perhaps wait for the second trace in an explicit step?

WilliamBergamin (approved):

zimeg:
@ewanek1 @mwbrooks @WilliamBergamin Thanks for the reviews as our understanding of tests have improved! 🧪 ✨ 

Let's merge this now for use in our tests downstream. We might find the following trace interesting to explore in adjacent PRs for SDK managed connections - perhaps within Bolt internals?

```
SLACK_TRACE_PLATFORM_RUN_READY
```

🔌 Which is different from the "start" trace that begins when the CLI delegates! Otherwise we'll continue to wait for certain logs to appear in output. No regression but less stable for changes to the code in question 💾

PR #2361, opened 2025-09-08
web-api(fix):Update SearchMessagesArguments to support cursor pagination

vegeris:
### Requirements ()


packages/web-api/src/types/request/search.ts
```diff
@@ -12,9 +12,23 @@ interface Searchable extends OptionalTeamAssignable, SortDir {
   sort?: 'score' | 'timestamp';
 }
 
+interface SearchMessagesCursorPagination {
+  /**
+   * @description Paginate through collections of data by setting the `cursor` parameter to `*` for the first "page"
+   * or a `next_cursor` attribute returned by a previous request's `response_metadata`.
+   * Use the `count` parameter to set the number of items to return per page rather than `limit`.
```
zimeg, line 19: 🧮 praise: Thank you for including this.

packages/web-api/src/types/request/search.ts
```diff
@@ -12,9 +12,23 @@ interface Searchable extends OptionalTeamAssignable, SortDir {
...
 export interface SearchAllArguments extends TokenOverridable, TraditionalPagingEnabled, Searchable {}
 // https://docs.slack.dev/reference/methods/search.files
 export interface SearchFilesArguments extends TokenOverridable, TraditionalPagingEnabled, Searchable {}
 // https://docs.slack.dev/reference/methods/search.messages
-export interface SearchMessagesArguments extends TokenOverridable, TraditionalPagingEnabled, Searchable {}
+export interface SearchMessagesArguments
+  extends TokenOverridable,
+    TraditionalPagingEnabled,
+    Searchable,
+    SearchMessagesCursorPagination {}
```
zimeg, line 34: 👁️‍🗨️ suggestion(non-blocking): I don't believe a similar `cursor` is used elsewhere? Might be nice to inline this IMO but no blocker!

zimeg (approved):
@vegeris LGTM! This seems good to merge when the time is right but it might be nice to add a type test for this before? 🧪 

https://github.com/slackapi/node-slack-sdk/blob/893a37dd3da2a66cc6cfbbd9d66e57eeab22295a/packages/web-api/test/types/methods/search.test-d.ts#L28-L37

I left a few other comments too!

packages/web-api/src/types/request/search.ts
vegeris: This is how Biome wants it to look 😛 

Yeah, I think an API-specific interface makes sense here. The same `count` param from the TraditionalPagingEnabled interface is reused for cursor pagination

PR #2362, opened 2025-09-09
feat(web-api): add chat.{appendStream|startStream|stopStream} methods

zimeg:
### Summary

This PR adds the `chat.{appendStream|startStream|stopStream}` methods to `@slack/web-api` 👾 ✨

### Preview

```js
const response1 = await client.chat.startStream({
  channel: "C0123456789",
});
const response2 = await client.chat.appendStream({
  channel: "C0123456789",
  ts: "0123456789.00001",
  markdown_text: "hello",
});
const response3 = await client.chat.stopStream({
  channel: "C0123456789",
  ts: "0123456789.00001",
});
```

### Requirements


packages/web-api/src/types/response/ChatStopStreamResponse.ts
```diff
@@ -0,0 +1,9 @@
+import type { WebAPICallResult } from '../../WebClient';
+export type ChatStopStreamResponse = WebAPICallResult & {
```
mwbrooks, line 2: suggestion: Looking at the reference, we'll also want a `message` object type that includes a `streaming_state` string.

mwbrooks (commented):
🙌🏻 Looking awesome! We paired on this and while reviewing the PR, we noticed that the `chat.stopStream` response should also include a `message` object. I've left a comment about it in the code.

mwbrooks (approved):
✅ Approving! We've decided to follow-up on `streaming_state` because it must be added to both the `chat.stopStream` method and the `message_changed` event.

zimeg:
@mwbrooks Sweet! Thanks for keeping note of what's needed next 🏁 ✨

I'll merge this now to keep our changesets clear, but think the `streaming_state` parameter in the method response is useful before prereleases and such. I'm also curious if this appears in responses from the start stream method too but haven't found options other than:

- "in_progress" for the `message_changed` event
- "completed" from the `chat.stopStream` method

PR #2363, opened 2025-09-10
build: use latest biome schema version

zimeg:
### Summary

This PR prefers the `latest` biome schema version to avoid warnings in build scripts.

#### Before changes

```
$ npm run lint

> @slack/web-api@7.10.0 lint
> npx @biomejs/biome check .

biome.json:2:14 deserialize ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ⚠ The configuration schema version does not match the CLI version 2.2.0

    1 │ {
  > 2 │   "$schema": "https://biomejs.dev/schemas/2.0.5/schema.json",
      │              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    3 │   "extends": ["../../biome.json"],
    4 │   "overrides": [

  ℹ   Expected:                     2.2.0
      Found:                        2.0.5


  ℹ Run the command biome migrate to migrate the configuration file.


Checked 430 files in 63ms. No fixes applied.
Found 1 warning.
```

#### After changes

```
$ npm run lint

> @slack/web-api@7.10.0 lint
> npx @biomejs/biome check .

Checked 430 files in 63ms. No fixes applied.
```

### Notes

This might encourage us to keep following the latest version of `biome` with a minimum version requirement?

https://github.com/slackapi/node-slack-sdk/blob/893a37dd3da2a66cc6cfbbd9d66e57eeab22295a/packages/cli-hooks/package.json#L58

Although that isn't [recommended](https://biomejs.dev/internals/versioning/)...

> Due to the nature of these changes, it’s **highly recommended** to save the exact version in your package.json, instead of using range operators.

I'd be interested in revisiting how we include some dependencies like `tsc` and `biome` to use pinned versions later or instead of this change since unexpected updates to these would be an unfortunate reason to error?

### Requirements

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.0.5/schema.json",
+  "$schema": "https://biomejs.dev/schemas/latest/schema.json",
   "assist": {
     "actions": {
       "source": {
@@ -10,9 +10,7 @@
   "formatter": {
     "enabled": true,
     "formatWithErrors": false,
-    "includes": [
-      "**"
-    ],
+    "includes": ["**"],
     "attributePosition": "auto",
     "indentStyle": "space",
     "indentWidth": 2,
--- a/packages/cli-hooks/biome.json
+++ b/packages/cli-hooks/biome.json
@@ -1,4 +1,4 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.0.5/schema.json",
+  "$schema": "https://biomejs.dev/schemas/latest/schema.json",
   "extends": ["../../biome.json"]
 }
--- a/packages/cli-test/biome.json
+++ b/packages/cli-test/biome.json
@@ -1,4 +1,4 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.0.5/schema.json",
+  "$schema": "https://biomejs.dev/schemas/latest/schema.json",
   "extends": ["../../biome.json"]
 }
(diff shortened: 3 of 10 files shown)


WilliamBergamin (approved):
Seems like a reasonable change 💯 

If the packages biome file extend `../../biome.json` do we need to define the schema in all child files?

zimeg:
@mwbrooks @WilliamBergamin Thanks both for the reviews! I agree this is an alright workaround for warnings but we should still investigate a more stable upgrade path...

> Generally, I lean toward pinning all dependencies to specific versions.

For development dependencies I agree too! Minimum versions are nice for production dependencies to unlock the newest features in application code without causing conflicts between sub-dependencies too I believe?

> If the packages biome file extend `../../biome.json` do we need to define the schema in all child files?

I found my editor doesn't recognize a standalone `biome.json` file without the schema at the moment and might not read from "extends" either 👁️‍🗨️ 

For now I'll merge this but am once again curious about more stable versioning for follow up! 🎁

PR #2364, opened 2025-09-10
feat(web-api): add message to chat.stopStream response type

zimeg:
### Summary

This PR adds `message` to the response types of `chat.stopStream` with the **blocks** from `@slack/types` in place of autogenerated builds for this time.

### Reviewers

Please do compare these values and types to actual responses! 👁️‍🗨️ 

### Requirements

### DIFF
--- a/packages/web-api/src/types/response/ChatStopStreamResponse.ts
+++ b/packages/web-api/src/types/response/ChatStopStreamResponse.ts
@@ -1,3 +1,4 @@
+import type { Block, KnownBlock } from '@slack/types';
 import type { WebAPICallResult } from '../../WebClient';
 export type ChatStopStreamResponse = WebAPICallResult & {
   channel?: string;
@@ -6,4 +7,18 @@ export type ChatStopStreamResponse = WebAPICallResult & {
   ok?: boolean;
   provided?: string;
   ts?: string;
+  message?: ChatStopStreamResponseMessage;
 };
+
+export interface ChatStopStreamResponseMessage {
+  subtype?: string;
+  text?: string;
+  user?: string;
+  streaming_state?: string;
+  type?: string;
+  ts?: string;
+  bot_id?: string;
+  thread_ts?: string;
+  parent_user_id?: string;
+  blocks?: (Block | KnownBlock)[];
+}


packages/web-api/src/types/response/ChatStopStreamResponse.ts
```diff
@@ -1,3 +1,4 @@
+import type { Block, KnownBlock } from '@slack/types';
 import type { WebAPICallResult } from '../../WebClient';
 export type ChatStopStreamResponse = WebAPICallResult & {
```
mwbrooks, line 3: suggestion: In the payload example that we captured earlier today, I'm also seeing a `response_metadata`. So we may want to add a `response_metadata?: ResponseMetadata;`.

packages/web-api/src/types/response/ChatStopStreamResponse.ts
```diff
@@ -6,4 +7,18 @@ export type ChatStopStreamResponse = WebAPICallResult & {
   ok?: boolean;
   provided?: string;
   ts?: string;
+  message?: ChatStopStreamResponseMessage;
 };
+
+export interface ChatStopStreamResponseMessage {
+  subtype?: string;
```
mwbrooks, line 14: suggestion: While the project isn't the best at it, perhaps we can alphabetize these arguments? The same for `ChatStopStreamResponse`.

packages/web-api/src/types/response/ChatStopStreamResponse.ts
zimeg: 🧮 We might find this included with the [`WebAPICallResult`](https://docs.slack.dev/tools/node-slack-sdk/reference/web-api/interfaces/WebAPICallResult) union at the moment I believe? But FWIW I like making these explicit parameters and am so open to revisiting this at some point.

mwbrooks: Ahh perfect! 👌🏻 No need to make it explicit then!

packages/web-api/src/types/response/ChatStopStreamResponse.ts
zimeg: @mwbrooks Am open to this but was matching the response ordering itself. I'm not sure how stable this is or if it's worth matching here...

Later additions for sure are not as clear without such a pattern!

mwbrooks (approved):

zimeg:
@mwbrooks Thanks so much for the kind and fast review!

I'll hold off on alphabetics in this changeset but we ought find similarities across these responses soon... 🚀 ✨

PR #2370, opened 2025-09-13
feat(types): add feedback_buttons and icon_button blocks as context_actions block elements

zimeg:
### Summary

This PR adds the following to `@slack/types`:

- `feedback_buttons` block action
- `icon_button` block action
- `context_actions` block

### Requirements

### DIFF
--- a/packages/types/src/block-kit/block-elements.ts
+++ b/packages/types/src/block-kit/block-elements.ts
@@ -139,6 +139,52 @@ export interface EmailInput extends Actionable, Dispatchable, Focusable, Placeho
 }
 
 /**
+ * @description Buttons to indicate positive or negative feedback.
+ */
+export interface FeedbackButtons extends Actionable {
+  /**
+   * @description The type of block. For a feedback buttons block, `type` is always `feedback_buttons`.
+   */
+  type: 'feedback_buttons';
+  /**
+   * @description A button to indicate positive feedback.
+   */
+  positive_button: {
+    /**
+     * @description Defines an object containing some text.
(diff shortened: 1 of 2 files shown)


packages/types/src/block-kit/block-elements.ts
```diff
@@ -161,6 +207,38 @@ export interface FileInput extends Actionable {
...
+export interface IconButton extends Actionable, Confirmable {
+  /**
+   * @description The type of element. In this case `type` is always `file_input`.
+   */
+  type: 'icon_button';
+  /**
+   * @description The icon to show.
+   * @example trash
+   */
+  icon: string;
```
mwbrooks, line 222: question: Today, only `trash` is supported. Do we typically list the supported enum values here?

mwbrooks (approved):
✅ Looks good and matches the spec!

packages/types/src/block-kit/block-elements.ts
zimeg: @mwbrooks I was so curious about this, but wanted to start more general and with an example to guard against future changes.

So open to updating this if being strict seems better of course 👁️‍🗨️ ✨

mwbrooks: I actually think keeping it open is better as well, because it allows devs to use the SDK while the API evolves with more values!

zimeg:
@mwbrooks Once again, thank you tons for the kind review! I'm wanting to merge this for tests with **typescript** so let's 🚢 💨

PR #2371, opened 2025-09-16
feat(types): add streaming_state to bot_message event

zimeg:
### Summary

This PR adds `streaming_state` to the [`bot_message`](https://docs.slack.dev/reference/events/message/bot_message) message event subtype.

### Reviewers

Confirm that this field appears in message changed event 🤖 

### Requirements

### DIFF
--- a/packages/types/src/events/message.ts
+++ b/packages/types/src/events/message.ts
@@ -68,6 +68,7 @@ export interface BotMessageEvent {
   event_ts: string;
   channel: string;
   channel_type: ChannelTypes;
+  streaming_state?: 'in_progress' | 'completed' | 'errored';
   ts: string;
   text: string;
   bot_id: string;


packages/types/src/events/message.ts
```diff
@@ -68,6 +68,7 @@ export interface BotMessageEvent {
   event_ts: string;
   channel: string;
   channel_type: ChannelTypes;
+  streaming_state?: 'in_progress' | 'completed' | 'errored';
```
mwbrooks, line 71: ✅ Matches the schema! 👌🏻 

nit: We may want to pull the values out into an enum at the top of the file, similar to  `ChannelTypes`.

zimeg: > ❓ Should this belong to more than just the `BotMessageEvent`? For example, the `MessageChangedEvent`?
> 
> nit: We may want to pull the values out into an enum at the top of the file, similar to  `ChannelTypes`.

@mwbrooks I was curious about this as well - at the moment I find a "bot_message" subtype within a "message_changed" event as the **message** but I'm not sure if it can be found in other events...

https://github.com/slackapi/node-slack-sdk/blob/4b8bfecff24bc80c36ab2537138617b5aca5a9d7/packages/types/src/events/message.ts#L230-L240

Am curious about this and think we could adjust to a separate enum if this is found elsewhere? 👻

mwbrooks: note: @zimeg and I chatted offline to make sure that `message_changed` events from bots will include the `streaming_state` arg. This arg isn't available on non-bot message changes, so it's looking good.

note: We won't bother turning the values into an enum until we repeat it in 2 places. 👌🏻

mwbrooks:
Merge away @zimeg! 🚀

zimeg:
@mwbrooks Thanks so much for sharing in these discussions and the kind review! I like what we have for now but am so open to revisiting in a change of events.

PR #2377, opened 2025-09-23
feat(web-api): add `recipient_team_id` and `recipient_user_id` to `chat.startStream` method

mwbrooks:
### Summary

This pull request add 2 optional types to the `chat.startStream` method:

* `recipient_team_id` is an optional Team ID that is required when `recipient_user_id` exists
* `recipient_user_id` is an optional User ID that is required when `recipient_team_id` exists

The developer must provide these optional params when starting a conversation stream outside of a DM.

### Reviewers

I didn't add additional logic to enforce the co-dependence between the params. As far as I can tell, the other types let the API return an error instead of adding that logic inside the SDK.


mwbrooks:
Thanks for the quick review @zimeg! 🙇🏻 I'll hit merge on this one and look at repeating it for the Python SDK next. 🐍

PR #2379, opened 2025-09-24
feat(web-api): add a chat stream method to the web client

zimeg:
### Summary

This PR adds a `WebClient#chatStream` method that creates chat streams 🤖

### Preview

```js
const streamer = client.chatStream({
  channel: "C0123456789",
  thread_ts: "1700000001.123456",
  recipient_team_id: "T0123456789",
  recipient_user_id: "U0123456789",
  buffer_size: 500,
});
await streamer.append({
  markdown_text: "**hello wo",
});
await streamer.append({
  markdown_text: "rld!**",
});
await streamer.stop();
```

### Testing

WIP-

### Notes

- Will save documentation generation for a later commit since the `diff` is significant at the moment I find.
- A default text buffer is used to avoid reaching rate limits in frequent updates.
- Super open to thoughts on the approach and implementation - does this seem like an alright change?
- Planning to follow up with tests after discussion to avoid extra changes too 👾 

### Requirements

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -513,10 +513,10 @@ export class WebClient extends Methods {
   /**
    * Stream markdown text into a conversation.
    *
-   * @description The "stream" method starts a new chat stream in a coversation that can be appended to. After appending an entire message, the stream can be stopped with concluding arguments such as "blocks" for gathering feedback.
+   * @description The "chatStream" method starts a new chat stream in a coversation that can be appended to. After appending an entire message, the stream can be stopped with concluding arguments such as "blocks" for gathering feedback.
    *
    * @example
-   * const streamer = await client.stream({
+   * const streamer = await client.chatStream({
    *   channel: "C0123456789",
    *   thread_ts: "1700000001.123456",
    *   recipient_team_id: "T0123456789",
@@ -534,7 +534,7 @@ export class WebClient extends Methods {
    * @see {@link https://docs.slack.dev/reference/methods/chat.appendStream}
    * @see {@link https://docs.slack.dev/reference/methods/chat.stopStream}
    */
-  public async stream(params: ChatStartStreamArguments): Promise<ChatStreamer> {
+  public async chatStream(params: ChatStartStreamArguments): Promise<ChatStreamer> {
     const streamer = new ChatStreamer(this);
     await streamer.start(params);
     return streamer;
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -17,7 +17,7 @@ import isElectron from 'is-electron';
 import isStream from 'is-stream';
 import pQueue from 'p-queue';
 import pRetry, { AbortError } from 'p-retry';
-import ChatStreamer from './chat-stream';
+import { ChatStreamer, type ChatStreamerOptions } from './chat-stream';
 import {
   httpErrorFromResponse,
   platformErrorFromResult,
@@ -516,7 +516,7 @@ export class WebClient extends Methods {
    * @description The "chatStream" method starts a new chat stream in a coversation that can be appended to. After appending an entire message, the stream can be stopped with concluding arguments such as "blocks" for gathering feedback.
    *
    * @example
-   * const streamer = await client.chatStream({
+   * const streamer = client.chatStream({
    *   channel: "C0123456789",
    *   thread_ts: "1700000001.123456",
    *   recipient_team_id: "T0123456789",
@@ -534,10 +534,12 @@ export class WebClient extends Methods {
    * @see {@link https://docs.slack.dev/reference/methods/chat.appendStream}
    * @see {@link https://docs.slack.dev/reference/methods/chat.stopStream}
    */
-  public async chatStream(params: ChatStartStreamArguments): Promise<ChatStreamer> {
-    const streamer = new ChatStreamer(this);
-    await streamer.start(params);
-    return streamer;
+  public chatStream(params: Omit<ChatStartStreamArguments & ChatStreamerOptions, 'markdown_text'>): ChatStreamer {
+    const { buffer_size, ...args } = params;
+    const options: ChatStreamerOptions = {
+      buffer_size,
+    };
+    return new ChatStreamer(this, this.logger, args, options);
   }
 
   /**
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -17,7 +17,7 @@ import isElectron from 'is-electron';
 import isStream from 'is-stream';
 import pQueue from 'p-queue';
 import pRetry, { AbortError } from 'p-retry';
-
+import ChatStreamer from './chat-stream';
 import {
   httpErrorFromResponse,
   platformErrorFromResult,
@@ -35,8 +35,8 @@ import { getUserAgent } from './instrument';
 import { getLogger, type Logger, LogLevel } from './logger';
 import { Methods } from './methods';
 import { type RetryOptions, tenRetriesInAboutThirtyMinutes } from './retry-policies';
+import type { ChatStartStreamArguments } from './types/request';
 import type { CursorPaginationEnabled } from './types/request/common';
-
 import type {
   FilesCompleteUploadExternalArguments,
   FilesGetUploadURLExternalArguments,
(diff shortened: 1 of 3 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -510,6 +510,36 @@ export class WebClient extends Methods {
...
+   * await streamer.append({
+   *   markdown_text: "rld!**",
+   * });
+   * await streamer.stop();
+   *
+   * @see {@link https://docs.slack.dev/reference/methods/chat.startStream}
+   * @see {@link https://docs.slack.dev/reference/methods/chat.appendStream}
+   * @see {@link https://docs.slack.dev/reference/methods/chat.stopStream}
+   */
+  public async stream(params: ChatStartStreamArguments): Promise<ChatStreamer> {
```
mwbrooks: suggestion: Personally, I'd rename this to `chatStream` because it matches the `chat.startStream` and `chat.postMessage` naming. Also, internally it looks like you've already felt like `ChatStreamer` was a good name. 😄

mcodik: +1 chatStream. we might stream to canvases or huddles some day

packages/web-api/src/WebClient.ts
```diff
@@ -510,6 +510,36 @@ export class WebClient extends Methods {
...
+    const streamer = new ChatStreamer(this);
+    await streamer.start(params);
```
mcodik: startStream turns off the loading animation, we should defer it until we have the first bytes to stream.

packages/web-api/src/chat-stream.ts
```diff
@@ -0,0 +1,177 @@
...
+    if (params.token) {
+      this.token = params.token;
+    }
+    if (!this.channel) {
+      throw new Error('failed to append stream: channel not found');
+    }
+    if (!this.streamTs) {
+      throw new Error('failed to append stream: ts not found');
+    }
+    const response = await this.client.chat.appendStream({
```
mcodik: plz buffer some characters here, plus some ways for developers to tune the buffer size

packages/web-api/src/WebClient.ts
zimeg: @mwbrooks @mcodik I knew "stream" felt off for some reason... These are great suggestions and we now have https://github.com/slackapi/node-slack-sdk/pull/2379/commits/6a7e7ba2699d480ea5a3b0aca002bdde1e150bff that changes this to `WebClient#chatStream` for fast use 👾 ✨

packages/web-api/src/WebClient.ts
zimeg: @mcodik Great callout! The changes of https://github.com/slackapi/node-slack-sdk/pull/2379/commits/826e46cf7a401f3e97f7dcaeec479389c14e0751 address this but also make the `ChatStreamer` class more opinionated to no longer expose a "start" method.

Instead we expose "append" and "stop" with the actual "chat.startStream" method being called when sending the first of the markdown text!

I find this alright to use in development but am open to adjustments however needed 🤖

packages/web-api/src/chat-stream.ts
zimeg: @mcodik The `WebClient#chatStream` method now configures this with changes of https://github.com/slackapi/node-slack-sdk/pull/2379/commits/826e46cf7a401f3e97f7dcaeec479389c14e0751:

```js
const streamer = client.chatStream({
  channel: "C0123456789",
  thread_ts: "1700000001.123456",
  recipient_team_id: "T0123456789",
  recipient_user_id: "U0123456789",
  buffer_size: 500,                  // This defaults to "250" characters at the moment
});
```

mcodik (approved):
looks great thanks!

packages/web-api/src/chat-stream.ts
```diff
@@ -0,0 +1,180 @@
...
+   *   channel: "C0123456789",
+   *   thread_ts: "1700000001.123456",
+   *   recipient_team_id: "T0123456789",
+   *   recipient_user_id: "U0123456789",
+   * };
+   * const streamer = new ChatStreamer(client, logger, args, { buffer_size: 500 });
+   * await streamer.append({
+   *   markdown_text: "**hello world!**",
+   * });
+   * await streamer.stop();
```
mwbrooks, line 48: praise: Love seeing the example for the generated docs!

packages/web-api/src/chat-stream.ts
```diff
@@ -0,0 +1,180 @@
...
+   * @example
+   * const client = new WebClient(process.env.SLACK_BOT_TOKEN);
+   * const logger = new ConsoleLogger();
+   * const args = {
+   *   channel: "C0123456789",
+   *   thread_ts: "1700000001.123456",
+   *   recipient_team_id: "T0123456789",
+   *   recipient_user_id: "U0123456789",
+   * };
+   * const streamer = new ChatStreamer(client, logger, args, { buffer_size: 500 });
```
mwbrooks: suggestion: 512 🤩 
```suggestion
   * const streamer = new ChatStreamer(client, logger, args, { buffer_size: 512 });
```

packages/web-api/src/chat-stream.ts
```diff
@@ -0,0 +1,180 @@
+import type { Logger } from '@slack/logger';
+import type { ChatAppendStreamArguments, ChatStartStreamArguments, ChatStopStreamArguments } from './types/request';
+import type { ChatAppendStreamResponse, ChatStartStreamResponse, ChatStopStreamResponse } from './types/response';
+import type WebClient from './WebClient';
+
+export interface ChatStreamerOptions {
+  /**
+   * @description The length of markdown_text to buffer in-memory before calling a method. Increasing this value decreases the number of method calls made for the same amount of text, which is useful to avoid rate limits.
+   * @default 250
```
mwbrooks: suggestion: the nerdy side of me would really enjoy this as 256 (2^8) for absolutely no reason except it feels good. 😄
```suggestion
   * @default 256
```

packages/web-api/src/chat-stream.ts
```diff
@@ -0,0 +1,180 @@
...
+   * @see {@link https://docs.slack.dev/reference/methods/chat.startStream}
+   * @see {@link https://docs.slack.dev/reference/methods/chat.appendStream}
+   * @see {@link https://docs.slack.dev/reference/methods/chat.stopStream}
+   */
+  constructor(client: WebClient, logger: Logger, args: ChatStartStreamArguments, options: ChatStreamerOptions) {
+    this.args = args;
+    this.client = client;
+    this.logger = logger;
+    this.options = {
+      buffer_size: options.buffer_size ?? 250,
```
mwbrooks: suggestion: Ignore me if you want. 😅 
```suggestion
      buffer_size: options.buffer_size ?? 256,
```

packages/web-api/src/chat-stream.ts
```diff
@@ -0,0 +1,180 @@
...
+   *   markdown_text: "**hello wo",
+   * });
+   * await streamer.append({
+   *   markdown_text: "rld!**",
+   * });
+   * await streamer.stop();
+   * @see {@link https://docs.slack.dev/reference/methods/chat.appendStream}
+   */
+  async append(
+    params: Omit<ChatAppendStreamArguments, 'channel' | 'ts'>,
```
mwbrooks: praise: Nice touch that we're just passing in the appendSteam args.

❓ question: I noticed that the constructor calls it `args` while the append calls it `params`. Is there a reason?

packages/web-api/src/chat-stream.ts
```diff
@@ -0,0 +1,180 @@
...
+        ...this.args,
+        token: this.token,
+        ...params,
+        markdown_text: this.buffer,
+      });
+      this.buffer = '';
+      this.streamTs = response.ts;
+      this.state = 'in_progress';
+      return response;
+    }
```
mwbrooks, line 169: question: Why did you choose to `client.chat.startStream(...)` in both the `.stop(...)` and `.flushBuffer(...)` instead of the constructor? Just curious.

mwbrooks (commented):
Exciting work @zimeg! I've left a few comments and I'll take it for a spin now 🧪

mwbrooks (approved):
✅ Nice work @zimeg! Throwing an approval on here - you can work through my feedback and decide what to keep or not.

🧪 I've manually tested locally with a sample app and it works very well. I played with buffer sizes between 250-512 and didn't notice any major difference, so I think 250 (256?) is a good default!

packages/web-api/src/chat-stream.ts
zimeg: 🧮 Super appreciate the suggestion! We add this in 771a6efca2c5c802dee201566197dcfc5ac292ec.

packages/web-api/src/chat-stream.ts
zimeg: 🤓 praise: I like this too! 771a6efca2c5c802dee201566197dcfc5ac292ec

packages/web-api/src/chat-stream.ts
zimeg: @mwbrooks Instead this commit 771a6efca2c5c802dee201566197dcfc5ac292ec 🤖

packages/web-api/src/chat-stream.ts
zimeg: @mwbrooks Thanks for calling this out! I agree that a more general typing is nice to have here! 👾 

> I noticed that the constructor calls it args while the append calls it params. Is there a reason?

A partial reason might be quick naming, but I've added https://github.com/slackapi/node-slack-sdk/pull/2379/commits/fcbce940aeec6f80bb0ac3d58f7c73a65bfeb4bb and https://github.com/slackapi/node-slack-sdk/pull/2379/commits/0f852cdf70408142b37b21a30b57359dd0522af5 to be more consistent in how method arguments mapping to `args` and `params` including options 🙏 ✨

packages/web-api/src/chat-stream.ts
zimeg: @mwbrooks This skips an extra `.appendStream` method call if `streamer.stop()` is called on an existing stream I am hoping. We also avoid adding this to the constructor to delay this as an asynchronous process to favor loading states:

> startStream turns off the loading animation, we should defer it until we have the first bytes to stream.

🔗 https://github.com/slackapi/node-slack-sdk/pull/2379#discussion_r2376599744

PR #2382, opened 2025-09-29
fix(web-api): update `thread_ts` arg to be required for `chat.startStream` method

mwbrooks:
### Summary

This PR updates the `chat.startStream` arguments to now require `thread_ts`. We are changing this because the Slack API has been updated to now require the `thread_ts` when starting a stream. This means all DMs and in-channel conversations with your bot will now have responses replied in-thread.

After merging, we'll need to update our sample apps to always provide `thread_ts`.

### DIFF
--- a/packages/web-api/test/types/methods/chat.test-d.ts
+++ b/packages/web-api/test/types/methods/chat.test-d.ts
@@ -588,6 +588,16 @@ expectAssignable<Parameters<typeof web.chat.scheduledMessages.list>>([]); // no 
 // -- sad path
 expectError(web.chat.startStream()); // lacking argument
 expectError(web.chat.startStream({})); // empty argument
+expectError(
+  web.chat.startStream({
+    channel: 'C1234', // missing thread_ts
+  }),
+);
+expectError(
+  web.chat.startStream({
+    thread_ts: '1234.56', // missing channel
+  }),
+);
 // -- happy path
 expectAssignable<Parameters<typeof web.chat.startStream>>([
   {


packages/web-api/src/types/request/chat.ts
```diff
@@ -218,12 +218,7 @@ export type ChatScheduledMessagesListArguments = OptionalArgument<
     Partial<Channel>
 >;
 
-export interface ChatStartStreamArguments
-  extends TokenOverridable,
-    Channel,
-    Partial<ThreadTS>,
-    Partial<MarkdownText>,
-    Unfurls {
+export interface ChatStartStreamArguments extends TokenOverridable, Channel, Partial<MarkdownText>, ThreadTS, Unfurls {
```
mwbrooks, line 221: note: `npm test` linter requires this to be a single-line. I ordered it alphabetically.

mwbrooks (commented):
Notes for the kind reviewers! 🙇🏻

packages/web-api/test/types/methods/chat.test-d.ts
zimeg, line 590:
```suggestion
expectError(web.chat.startStream({})); // empty argument
expectError(web.chat.startStream({
  channel: 'C1234', // missing thread_ts
}));
```

🧪 suggestion(non-blocking): Asserting that the previous behavior errors might be nice to have, but no blocker for this PR!

zimeg (approved):
@mwbrooks Thanks so much for a fast update on this once more! ⭐ 

I marked this as a `patch` change since I believe aligning to the API methods can be considered a bugfix with invalid arguments erroring later anyways...

We should add this to a prerelease soon IMHO! 🚢 💨

packages/web-api/test/types/methods/chat.test-d.ts
mwbrooks: Great suggestion @zimeg! Commit ef9bfc62 adds tests for `missing thread_ts` and `missing channel`. I couldn't bring in your suggestion because the linter wanted a different format 🤖 📏

mwbrooks:
Thanks for the quick review @zimeg and solid suggestion on covering the edge-case tests. 🙇🏻 I like that you've added the `patch` label. I was leaving it off because this is a PR into a branch, but personally I like seeing semver on all PRs 😄

Issue #2359, opened 2025-09-04
Removing support for Node 18 and other upcoming breaking changes


zimeg:
👋 Hello! This issue announces support for Node 18 is being removed within a few months to add support and develop with changes from more recent Node versions. Ongoing changes and discussion can be followed along here!

## 📆 Timeline

As overview for what's ahead, the following timeline might be useful:

- [x] Upstream Node 18 reaches EOL: 2025-03-27.
- [x] Notice of removing support: 2025-09-03.
- [x] Development releases shared: 2025-10-03 or sooner.
- [ ] Public release of new package versions: 2025-12-02 or later.
- [ ] Previous versions become unsupported: 2026-01-01 or later.

**References**

- Node support schedule: https://nodejs.org/en/about/previous-releases
- Slack support schedule: https://docs.slack.dev/tools/node-slack-sdk/support-schedule

## 👾 Planned changes

Breaking changes planned for this next major version follow.

### Supported Node version

The Node 18 release will no longer be supported or used in testing. Future development will use features from the latest releases for the following versions:

- Node 20 - "Maintenace": What we use in testing.
- Node 22 - "Active": What we use in testing.
- Node 24 - "Current": What we will recommend applications use.

### Using the fetch API to send requests

The [`axios`](https://axios-http.com) package has been used to send requests with both the `@slack/web-api` and `@slack/webhook` packages. We'll look to replace this with the [`undici`](https://undici.nodejs.org/#/) package to align with the Node implementation. Related: #1525 #1761.

### Using request and response in installation callbacks

The installer handler callback arguments from `@slack/oauth` are changing to `Request` and `Response` objects as a more compatible standard. For: #2270.

### Unnesting responses from the filesUploadV2 method

The shape of responses from the `filesUploadV2` method of `@slack/web-api` will be flattened to return file indicies at the top of the response. From: #1803.

### Converting code samples and examples to ECMAScript modules

Examples documented for these packages will use [ECMAScript modules](https://nodejs.org/docs/latest-v22.x/api/esm.html#modules-ecmascript-modules).

### Removing deprecated features

These next major versions will remove deprecated features from packages.

### Removing unsupported packages from the main branch

The `@slack/client`, `@slack/events-api`, and `@slack/interactive-messages` packages are being removed from the main branch after being EOL for a handful of years.

### Converting test runners to default scaffoldings

[...]

PR #2403, opened 2025-10-08
feat(web-api): add support for AbortSignal

### LINKED ISSUE #1761: Please support  `AbortSignal`
[AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) is a standard interface for canceling requests, and [axois](https://axios-http.com/docs/cancellation) also support it. It would be good if we can manage slack requests using standard interface.

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] `@slack/types`
- [ ] I don't know

### Requirements

LinusU:
### Summary

This adds support for aborting in-flight requests, using the standardized [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal)

Fixes #1761

I've designed this as an additional parameter instead of mixing it with the body of the specific endpoint. I believe that this is a sounder approach as there will never be any conflicts, and it also avoids the current problem of `token` also being present in the request body as well as in the header when specified this way.

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -343,7 +343,11 @@ export class WebClient extends Methods {
    * @param method - the Web API method to call {@link https://docs.slack.dev/reference/methods}
    * @param options - options
    */
-  public async apiCall(method: string, options: Record<string, unknown> = {}): Promise<WebAPICallResult> {
+  public async apiCall(
+    method: string,
+    options: Record<string, unknown> = {},
+    config?: { signal?: AbortSignal },
+  ): Promise<WebAPICallResult> {
     this.logger.debug(`apiCall('${method}') start`);
 
     warnDeprecations(method, this.logger);
@@ -369,6 +373,7 @@ export class WebClient extends Methods {
         ...options,
       },
       headers,
+      config,
     );
     const result = await this.buildResult(response);
     this.logger.debug(`http request result: ${JSON.stringify(result)}`);
@@ -678,6 +683,7 @@ export class WebClient extends Methods {
     url: string,
     body: Record<string, unknown>,
     headers: Record<string, string> = {},
+    options?: { signal?: AbortSignal },
   ): Promise<AxiosResponse> {
     // TODO: better input types - remove any
     const task = () =>
@@ -686,6 +692,7 @@ export class WebClient extends Methods {
           // biome-ignore lint/suspicious/noExplicitAny: TODO: type this
           const config: any = {
             headers,
+            signal: options?.signal,
             ...this.tlsConfig,
           };
           // admin.analytics.getFile returns a binary response
@@ -763,6 +770,9 @@ export class WebClient extends Methods {
           // biome-ignore lint/suspicious/noExplicitAny: errors can be anything
           const e = error as any;
           this.logger.warn('http request failed', e.message);
+          if (e.name === 'CanceledError' && options?.signal?.reason) {
+            throw new AbortError(options.signal.reason);
+          }
           if (e.request) {
             throw requestErrorWithOriginal(e, this.attachOriginalToWebAPIRequestError);
           }
--- a/.changeset/web-api-abort-signal.md
+++ b/.changeset/web-api-abort-signal.md
@@ -1,0 +1,5 @@
+---
+"@slack/web-api": minor
+---
+
+Add support for [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal).
--- a/packages/web-api/src/WebClient.test.ts
+++ b/packages/web-api/src/WebClient.test.ts
@@ -2048,6 +2048,62 @@ describe('WebClient', () => {
       }
     });
   });
+
+  describe('accepts an AbortSignal to cancel requests', () => {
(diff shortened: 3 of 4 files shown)


LinusU:
@salesforce-cla signed ✅

zimeg:
Hey @LinusU! 👋 Thanks so much for sending this contribution in 👾 ✨ 

At the moment we're exploring another approach of a custom fetch attribute as an option to the `WebClient` alongside efforts of #2359.

I don't have more to share for `@slack/web-api` at the moment, but an [example](https://github.com/slackapi/node-slack-sdk/tree/56a6183f91437c75fc414fad81d26011d70f2ff5/packages/webhook#send-requests-with-a-custom-fetch-adapter) from `@slack/webhook` might resemble:

https://github.com/slackapi/node-slack-sdk/blob/56a6183f91437c75fc414fad81d26011d70f2ff5/packages/webhook/README.md?plain=1#L95-L133

Let's keep this open while that's explored, but I'm curious if the `fetch` approach might also be suitable for you?

LinusU:
Hey @zimeg, thanks for taking a look at this! 👋 

> [...] but I'm curious if the fetch approach might also be suitable for you?

I would love for this package to be based on fetch instead of Aaxios, and I took special care when writing this to not expose any Axios-specific behaviour (e.g. that is why I do the little dance to throw the correct error, instead of the Axios-specific cancelled error). So this should pair up great together with the fetch move!

With that said, I do *not* think that being able to pass in a custom fetch-function would solve the problem here. We are using AbortSignals across large parts of our app, not just as a tool to set a specific timeout on http requests.

I need to be able to pass `signal` individually to each api call, not at the creation of `WebClient`.

-----

Side note: the example you linked will cancel every single request made 400 milliseconds after the `ProxyAgent` was constructed, not after each request has ran for 400ms?

LinusU:
ping @zimeg, have you had a chance to read my last comment? 🙏

LinusU:
ping @zimeg, have you had a chance to look at this? 🙏

packages/web-api/src/WebClient.ts
```diff
@@ -762,6 +769,9 @@ export class WebClient extends Methods {
           // biome-ignore lint/suspicious/noExplicitAny: errors can be anything
           const e = error as any;
           this.logger.warn('http request failed', e.message);
+          if (e.name === 'CanceledError' && options?.signal?.reason) {
+            throw new pRetry.AbortError(options.signal.reason);
+          }
```
zimeg, line 775: 🔬 question: Is `CanceledError` the most common error code here? I understand we can catch related errors in following changes, but I notice separate codes on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) docs:

- AbortError
- TimeoutError

And it's not obvious to me if that's exhaustive or if it's missing `CanceledError` and more! 🤓

packages/web-api/src/WebClient.ts
```diff
@@ -762,6 +769,9 @@ export class WebClient extends Methods {
           // biome-ignore lint/suspicious/noExplicitAny: errors can be anything
           const e = error as any;
           this.logger.warn('http request failed', e.message);
+          if (e.name === 'CanceledError' && options?.signal?.reason) {
+            throw new pRetry.AbortError(options.signal.reason);
```
zimeg: 🎁 thought: We throw new `AbortError` above and I'm uncertain if we want to use `new` here or continue referencing the `pRetry` package? The tests above do show this is caught as expected which is great.

zimeg (commented):
@LinusU Thank you for keeping these changes alive and relevant 💌 

I haven't had a good chance to test this yet but I do now understand it's a separate change from what I suggest earlier and think your approach of adding another argument is most solid:

```diff
  public async apiCall(
    method: string,
    options: Record<string, unknown> = {},
+   config?: { signal?: AbortSignal },
  ): Promise<WebAPICallResult> {
```

Leaving a few comments from code review which is looking great! I do notice a merge conflict appeared and I left a few questions for this meantime 🚢

packages/web-api/src/WebClient.ts
LinusU: `CanceledError` here is from `axios`, so it will _always_ be `CanceledError` on every platform, since it's an error constructed in "user space" code.

As long as axios is used, this will work. And as I said earlier, this intentionally does _not_ expose `CancelError`, so that a switch to `fetch` would be seamless! 🚀

packages/web-api/src/WebClient.ts
LinusU: I've removed the `pRetry.` part now, to make it match the rest of the code. However, it's the same class, just exposed in two different places, so no runtime difference ☺️

LinusU:
Sorry for the super late reply 🙈 I totally missed that you answered me the same day!

I've rebased on latest `main`, and also changed `pRetry.AbortError` -> `AbortError`

Tests are still passing locally for me, and I would still love to see this merged! 🙏

PR #2451, opened 2025-12-10
feat(web-api): Add support for admin.users.getExpiration

### LINKED ISSUE #2420: Add support for the admin.users.getExpiration API
There's a new API that we should add support for; [admin.users.getExpiration](https://docs.slack.dev/reference/methods/admin.users.getExpiration/)

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] `@slack/oauth`
- [ ] `@slack/socket-mode`
- [ ] `@slack/types`
- [ ] I don't know

### Requirements

misscoded:
### Summary

Adds support for `admin.users.getExpiration` (and updates generated responses). See [documentation](https://docs.slack.dev/reference/methods/admin.users.getExpiration/).

Resolves #2420.


packages/web-api/src/types/response/ChatStopStreamResponse.ts
```diff
@@ -1,24 +1,16 @@
...
+
 import type { WebAPICallResult } from '../../WebClient';
 export type ChatStopStreamResponse = WebAPICallResult & {
-  channel?: string;
   error?: string;
   needed?: string;
   ok?: boolean;
   provided?: string;
-  ts?: string;
-  message?: ChatStopStreamResponseMessage;
```
zimeg, line 10: 🪓 issue(non-blocking): This should be included in the Java generated logs but I'm finding tests aren't covering this for me. Let's keep note to follow up on this - I think using generated types is correct while "message" is optional.

👁️‍🗨️ ramble: But I wouldn't be opposed to reverting changes of this file if that seems uncertain?

zimeg (approved):
@misscoded LGTM and thanks so much for including the generated responses with this change 🙏 ✨ 

I left a comment about one of those changes, but no blockers for this method 🤖

packages/web-api/src/types/response/ChatStopStreamResponse.ts
misscoded: @zimeg Why would this suddenly go missing?

zimeg: @misscoded Agh I missed this - apologies for slow response! But I think some of those arguments were added manually for a release before the Java SDK had matching attributes.

misscoded: Ah, ok! We historically have opted not to do that (for this exact reason 😄), and just use the generic Web API response type until it's updated, then swap them out. Have we changed the approach to this then, or was this a one-off? Because if types that are available for folks to use change and go missing, that's technically a breaking change.

WilliamBergamin (approved):

packages/web-api/src/types/response/ChatStopStreamResponse.ts
WilliamBergamin: To prevent breaking changes should we simply leave those in for now?

zimeg: @WilliamBergamin @misscoded An adjacent PR #2696 shows the same response generation in scheduled commits so https://github.com/slackapi/node-slack-sdk/pull/2451/commits/b8f667c49e7fd0bb2aeb27364093b2bff6b9561a removes other methods from these changes for separate investigations 👾 

IIRC some of these missing responses was indeed one-off and will take note of the generic values to begin method implementation 🧠 ✨

PR #2470, opened 2026-01-16
feat: accept chunks as arguments to chat stream helper 

srtaalej:
### Summary
This PR enables chunks processing in the chat stream helper 

```js
    const streamer = new ChatStreamer(client, client.logger, {
      channel: CHANNEL_ID,
      thread_ts: threadTs,
      recipient_team_id: 'TEAM_ID,
      recipient_user_id: 'USER_ID',
    }, { buffer_size: 1 });

    await streamer.append({
      chunks: [
        {
          type: 'markdown_text',
          text: '**Hello!** I am starting to process your request...\n\n'
        }]
    });

    await streamer.stop({
      chunks: [
        {
          type: 'markdown_text',
          text: '\n\n---\n\n✅ **All tasks completed successfully!**\n',
        },
        {
          type: 'markdown_text',
          text: 'Thank you for using the streaming API with chunks.',
        },
      ],
    });
```

### DIFF
--- a/packages/web-api/src/chat-stream.ts
+++ b/packages/web-api/src/chat-stream.ts
@@ -14,19 +14,12 @@ export interface ChatStreamerOptions {
 
 export class ChatStreamer {
   private buffer = '';
-
   private client: WebClient;
-
   private logger: Logger;
-
   private options: Required<ChatStreamerOptions>;
-
   private state: 'starting' | 'in_progress' | 'completed';
-
   private streamArgs: ChatStartStreamArguments;
-
   private streamTs: string | undefined;
-
   private token: string | undefined;
 
   /**
@@ -132,7 +125,6 @@ export class ChatStreamer {
       throw new Error(`failed to stop stream: stream state is ${this.state}`);
     }
     const { markdown_text, chunks, ...opts } = args ?? {};
-
     if (opts.token) {
       this.token = opts.token;
     }
@@ -150,24 +142,21 @@ export class ChatStreamer {
       this.streamTs = response.ts;
       this.state = 'in_progress';
     }
-
-    const flushings: AnyChunk[] = [];
-
+    const chunksToFlush: AnyChunk[] = [];
     if (this.buffer.length > 0) {
-      flushings.push({
+      chunksToFlush.push({
         type: 'markdown_text',
         text: this.buffer,
       });
     }
     if (chunks) {
-      flushings.push(...chunks);
+      chunksToFlush.push(...chunks);
     }
-
     const response = await this.client.chat.stopStream({
       token: this.token,
       channel: this.streamArgs.channel,
       ts: this.streamTs,
-      chunks: flushings,
+      chunks: chunksToFlush,
       ...opts,
     });
     this.state = 'completed';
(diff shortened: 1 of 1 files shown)


mwbrooks (commented):
🙌🏻 Exciting, thanks for opening this PR @srtaalej!

It looks like you've also added the new `task_update` check along with the `plan_update` 😄 

🧪 @zimeg and I will take a look at the PR soon. In the meantime, it looks like there are some failed tests that will need to be resolved.

packages/web-api/src/chat-stream.ts
```diff
@@ -158,12 +166,13 @@ export class ChatStreamer {
   private async flushBuffer(
     args: Omit<ChatStartStreamArguments | ChatAppendStreamArguments, 'channel' | 'ts'>,
   ): Promise<ChatStartStreamResponse | ChatAppendStreamResponse> {
+    const finalArgs = this.buffer.length > 0 && !args.chunks ? { ...args, markdown_text: this.buffer } : args;
```
zimeg: 👀 note: We might want to add existing `buffer` to a **new** chunk with just markdown text, if the buffer has anything, **then** add the chunks argument. This would help make sure we're not appending while text is in the buffer!

packages/web-api/src/chat-stream.ts
```diff
@@ -89,8 +90,11 @@ export class ChatStreamer {
     if (args.token) {
       this.token = args.token;
     }
-    this.buffer += args.markdown_text;
-    if (this.buffer.length >= this.options.buffer_size) {
+    if (args?.markdown_text) {
+      this.buffer += args.markdown_text;
+      args.markdown_text = undefined;
```
mwbrooks: note: I pushed commit fc5ed1ab6bd948daf2b258e87bfaad82ec0d4490 to unset the `args.markdown_text` after adding it to the buffer.

If this is not done, then the `markdown_text` will be added to the API request, which is also included in the `chunks: [{ type: "markdown_text", "text": "...." }]`.

@zimeg Does this look correct to you? I noticed this wasn't explicitly done in the Python SDK.

packages/web-api/src/chat-stream.ts
```diff
@@ -144,12 +149,25 @@ export class ChatStreamer {
...
+    if (args?.chunks) {
+      flushings.push(...args.chunks);
+    }
+
     const response = await this.client.chat.stopStream({
       token: this.token,
       channel: this.streamArgs.channel,
       ts: this.streamTs,
+      chunks: flushings,
       ...args,
```
mwbrooks: note: I made commit fc5ed1ab6bd948daf2b258e87bfaad82ec0d4490 which added back `..args,`. If we don't have this, then we may drop additional args like `token`. This also has the side effect of adding `markdown_text` as an arg, which is why we now must unset it. .cc @zimeg

packages/web-api/src/WebClient.spec.ts
```diff
@@ -1228,7 +1228,11 @@ describe('WebClient', () => {
           ok: true,
           ts: '123.123',
         })
-        .post('/api/chat.stopStream', { channel: 'C0123456789', ts: '123.123', markdown_text: 'nice!' })
+        .post('/api/chat.stopStream', {
+          channel: 'C0123456789',
+          ts: '123.123',
+          chunks: JSON.stringify([{ type: 'markdown_text', text: 'nice!' }]),
```
mwbrooks, line 1234: note: I made commit fc5ed1ab6bd948daf2b258e87bfaad82ec0d4490 which updates this test to support `chunks` instead of `markdown_text` because the streaming helper will now convert all markdown text into a chunk.

packages/web-api/src/chat-stream.ts
zimeg: @mwbrooks Thanks so much for catching this! We might want to consider a "destructured" approach to separate `markdown_text` and also `chunks` from the remaining options:

```js
const { markdown_text, chunks, ...opts } = args;
```

This might help us avoid overriding new chunks with `...args` as well, or `chunks` can be a following parameter?

packages/web-api/src/chat-stream.ts
zimeg: @mwbrooks A change similar to other comments might be useful to destructure `markdown_text` and `chunks` from the rest of `args` before a chance to flush the buffer?

```js
const { markdown_text, chunks, ...opts } = args;
```

IIRC the arguments alongside `**kwargs` in the Python [implementation](https://github.com/slackapi/python-slack-sdk/blob/610de5b0c649f929933546415f35f9a56558e83b/slack_sdk/web/chat_stream.py#L73-L75) do this without added lines!

packages/web-api/src/chat-stream.ts
```diff
@@ -144,12 +150,25 @@ export class ChatStreamer {
       this.streamTs = response.ts;
       this.state = 'in_progress';
     }
+
+    const flushings: AnyChunk[] = [];
```
zimeg: 🙊 quibble: This variable was changed alike to "chunks to flush" in updates to the Python SDK that we might mirror here? Also unrelated but empty lines might be nice to remove before merging-

PR #2480, opened 2026-01-27
feat(cli-hooks): add default app and manifest watch config

zimeg:
### Summary

This PR adds a default `app` and `manifest` option to file watching for server restarts and app reinstalls.

### Requirements

### DIFF
--- a/packages/cli-hooks/src/get-hooks.js
+++ b/packages/cli-hooks/src/get-hooks.js
@@ -48,7 +48,7 @@ if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
 /**
  * Information about the manifest files to watch for app reinstalls.
  * @typedef ManifestConfigWatch
- * @property {string} filter-regex - Regex pattern for finding filtered files.
+ * @property {string} [filter-regex] - Regex pattern for finding filtered files.
  * @property {string[]} paths - Specific locations to begin searching for files.
  */
 


packages/cli-hooks/src/get-hooks.js
```diff
@@ -52,8 +66,14 @@ export default function getHooks() {
     },
     config: {
       watch: {
-        'filter-regex': '^manifest\\.json$',
-        paths: ['.'],
+        app: {
+          paths: ['.'],
+          'filter-regex': '.js$',
```
mwbrooks: suggestion: I think we should also support `.ts` for TypeScript developers using Bolt JS?

mwbrooks (commented):
Exciting times!

packages/cli-hooks/src/get-hooks.js
zimeg: @mwbrooks We might use the built changes from `tsc` for now, though these changes might be good to document as well!

A note on this is mentioned now in the `CHANGELOG` and also updates to the sample app:

https://github.com/slackapi/node-slack-sdk/blob/cc8e5b729000514b67e4d5c36a9178b984c07e6a/.changeset/green-items-write.md?plain=1#L41

👾 https://github.com/slack-samples/bolt-ts-starter-template/pull/178

zimeg:
🏷️ Follows: https://github.com/slackapi/slack-cli/pull/310

.changeset/green-items-write.md
```diff
@@ -0,0 +1,50 @@
...
+{
+  "manifest": {
+    "source": "local"
+  }
+}
+```
+
+### Custom Configurations
+
+You can override these defaults in your `.slack/hooks.json` file to reduce the paths searched or change the file patterns. Read [Watch Configurations](https://docs.slack.dev/tools/slack-cli/reference/hooks/#watch-configurations) for more options.
```
srtaalej, line 46: 🤩

packages/cli-hooks/src/get-hooks.spec.js
```diff
@@ -21,4 +21,19 @@ describe('get-hooks implementation', async () => {
...
+    assert(config.watch.app);
+    assert.equal(config.watch.app['filter-regex'], '\\.js$');
+    assert.deepEqual(config.watch.app.paths, ['.']);
+  });
+
+  it('should return watch config for manifest files', async () => {
+    const { config } = getHooks();
+    assert(config.watch);
+    assert(config.watch.manifest);
+    assert.deepEqual(config.watch.manifest.paths, ['manifest.json']);
```
srtaalej, line 37: ty for good tests 👾!

packages/cli-hooks/src/get-hooks.js
```diff
@@ -32,8 +32,22 @@ if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
...
+ * Information about file changes for CLI actions.
  * @typedef SDKConfigWatch
+ * @property {AppConfigWatch} app - Watch config for server restarts.
+ * @property {ManifestConfigWatch} manifest - Watch config for app reinstalls.
+ */
+
+/**
+ * Information about the app files to watch for server restarts.
+ * @typedef AppConfigWatch
+ * @property {string} filter-regex - Regex pattern for finding filtered files.
```
mwbrooks: suggestion: The spec states that `filter-regex` is option, so perhaps we can set that in jsdoc as well.
```suggestion
 * @property {string} [filter-regex] - Regex pattern for finding filtered files.
```

packages/cli-hooks/src/get-hooks.js
```diff
@@ -32,8 +32,22 @@ if (fs.realpathSync(process.argv[1]) === fileURLToPath(import.meta.url)) {
...
+ * @property {string[]} paths - Specific locations to begin searching for files.
+ */
+
+/**
+ * Information about the manifest files to watch for app reinstalls.
+ * @typedef ManifestConfigWatch
  * @property {string} filter-regex - Regex pattern for finding filtered files.
```
mwbrooks: suggestion: Updating `filter-regex` to be optional to match the spec.
```suggestion
 * @property {string} [filter-regex] - Regex pattern for finding filtered files.
```

packages/cli-hooks/src/get-hooks.js
```diff
@@ -52,8 +66,13 @@ export default function getHooks() {
     },
     config: {
       watch: {
-        'filter-regex': '^manifest\\.json$',
-        paths: ['.'],
+        app: {
+          'filter-regex': '\\.js$',
```
mwbrooks, line 70: nit: Non-blocker. Not sure if we want to include, this but we could ignore `node_modules` to avoid restarting the server when the developer runs `npm install`.

mwbrooks (approved):
✅ Love this improvement @zimeg! It's so exciting to see this land!

🧪 Local testing works great on my end! 👾 

🛠️ I left a few minor, non-blocking suggestions.

✏️ Love the very detailed changeset! This is what I'd like to see in more of our PRs for developers!

packages/cli-hooks/src/get-hooks.js
zimeg: @mwbrooks Our current CLI implementations don't support "not" paths AFAICT. We might want to follow up with this, since changes to the `test` directories might also be odd to cause restarts! 🤖

zimeg (commented):
@srtaalej @mwbrooks Thanks both so much for the kind reviews and catches 👾 ✨ 

I'm excited to merge this for an upcoming release, and hope we can continue to support good defaults for most developers!

packages/cli-hooks/src/get-hooks.js
zimeg: @mwbrooks Amazing catch! Thank you for sharing a sharp eye 👁️‍🗨️

PR #2493, opened 2026-01-30
docs: polishing source wording

lukegalbraithrussell:
### Summary

Makes it so it all matches the api docs

TODO: update generated reference docs. workspaces update changes some slugs (now full paths, not just subpackage paths), so need to decide if we're okay with all those slug changes appearing here

### DIFF
--- a/packages/types/src/chunk.ts
+++ b/packages/types/src/chunk.ts
@@ -17,7 +17,7 @@ export interface MarkdownTextChunk extends Chunk {
 }
 
 /**
- * An updated title of plans for task and tool calls.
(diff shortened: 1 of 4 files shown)


packages/types/src/chunk.ts
```diff
@@ -17,7 +17,7 @@ export interface MarkdownTextChunk extends Chunk {
 }
 
 /**
- * An updated title of plans for task and tool calls.
+ * Used for displaying an updated title of a plan.
```
mwbrooks, line 20: ✏️  ✨

PR #2513, opened 2026-02-12
fix(types): narrow tag_color to supported color literals

kang8:
## Summary

- Narrow `tag_color` type from `string` to `'red' | 'yellow' | 'green' | 'gray' | 'blue'` in `EntityTypedField`, `EntityStringField`, and `EntityCustomField`
- Per [Slack docs](https://docs.slack.dev/messaging/work-objects/#supported-properties-for-a-field): can only be set when the type is `string`, allows the string to be highlighted in one of the supported colors

image

### DIFF
--- a/packages/types/src/message-metadata.ts
+++ b/packages/types/src/message-metadata.ts
@@ -201,7 +201,7 @@ export interface EntityTypedField {
   slack_file?: SlackFile;
   alt_text?: string;
   edit?: EntityEditSupport;
-  tag_color?: string;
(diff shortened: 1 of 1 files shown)


zimeg:
Hey @kang8! 👋 Thanks so much for sending this in!

Before we make this change I'm checking with how the team wants to handle `enum` arguments over generic `string`. We've so far used generic strings to remain flexible to upstream changes, but I understand this can add effort to app development...

We'll share an update once alignment is reached 🪬

PR #2566, opened 2026-04-29
ci: simplify test scripts and separate coverage from test runs

WilliamBergamin:
### Summary

- **Separated test coverage from regular test runs** each package now has a `test` script and a `test:coverage` script. Previously most packages aliased test to test:unit which always included coverage instrumentation.
- **Removed the test:node18 scripts** these were workarounds for Node 18 lacking --test-reporter support.
- **Simplified CI workflow** replaced the split Run tests (Node 18/20) / Run tests (Node 22+) steps with a single Run tests step for most matrix entries, plus a Run test coverage step only on the latest supported Node version.                                                                                          
- Added glob-bin as a root devDependency to replace bash -O globstar glob expansion in test scripts (cross-platform compatibility on Windows CI). 
   
### Requirements 


.changeset/simple-tests-shine.md
zimeg, line 1: 🔭 note: I'm so curious for what happens to this file in upcoming releases!

.github/workflows/ci-build.yml
```diff
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+env:
+  LATEST_SUPPORTED_NODE: "24.x"
```
zimeg, line 10: 🧠 praise: Super nice to surface this clear!

.github/workflows/ci-build.yml
```diff
@@ -64,15 +67,14 @@ jobs:
         run: npm run lint
       - name: Build docs
```
zimeg, line 68:
```suggestion
      - name: Build docs
        if: matrix.node-version == env.LATEST_SUPPORTED_NODE
```

⌛ thought: We can save ~20 seconds for most runs with this? But perhaps it's adjacent to this PR!

packages/cli-hooks/package.json
```diff
@@ -37,9 +37,8 @@
   "scripts": {
     "build": "shx chmod +x src/*.js",
     "prelint": "tsc --noemit --module es2022 --maxNodeModuleJsDepth 0 --project ./jsconfig.json",
-    "test": "npm run test:unit",
-    "test:node18": "bash -c 'node --test --test-reporter=spec src/*.test.js'",
-    "test:unit": "node --experimental-test-coverage --test-reporter=spec --test-reporter-destination=stdout --test-reporter=lcov --test-reporter-destination=lcov.info --test-reporter=junit --test-reporter-destination=test-results.xml --test src/*.test.js"
+    "test": "bash -c 'node --test-reporter=spec --test-reporter-destination=stdout --test-reporter=junit --test-reporter-destination=test-results.xml --test src/*.test.js'",
```
zimeg, line 40:
```suggestion
    "test": "bash -c 'node --test src/*.test.js'",
```

🧪 thought(non-blocking): I like the pattern of a minimal `test` script and a more complex `test:coverage` but am uncertain which flags remain needed for the first?

packages/web-api/AGENTS.md
```diff
@@ -225,7 +225,7 @@ expectAssignable<Parameters<typeof web.chat.appendStream>>([
 
 ```bash
 npm test --workspace=packages/web-api         # all tests
-npm run test:unit --workspace=packages/web-api # unit only
+npm test --workspace=packages/web-api            # unit only
```
zimeg: 🔭 thought: One of these might be good to remove since it's unclear which test suite is used at a glance?

package.json
```diff
@@ -24,6 +24,7 @@
     "lint": "npx @biomejs/biome check packages",
     "lint:fix": "npx @biomejs/biome check --write packages",
     "test": "npm test --workspaces --if-present",
+    "test:coverage": "npm run test:coverage --workspaces --if-present",
```
zimeg, line 27: 🏆 praise: These patterns are joined well here!

zimeg (approved):
@WilliamBergamin Amazing improvements to CI health! I like the pattern of checking coverage for the latest version so much 🧪 ✨ 

I'm approving with a few notes and question about dependencies noted? If this all seems expected please do merge:

> Added glob-bin as a root devDependency to replace bash -O globstar glob expansion in test scripts (cross-platform compatibility on Windows CI).

I added https://github.com/slackapi/node-slack-sdk/pull/2566/commits/af3165f0379896fb44f33746551ff65583ec56be during review that hinted toward this package, but was this still used?

🚢 💨

.github/workflows/ci-build.yml
WilliamBergamin: Good catch 💯 I agree we can add this to speed up CI

.changeset/simple-tests-shine.md
WilliamBergamin: Same! originally I thought of adding a patch to each package but that would be inaccurate because this does not actually have any effect on what the user consumes

packages/web-api/AGENTS.md
WilliamBergamin: Good catch 💯

packages/cli-hooks/package.json
WilliamBergamin: These flags are interdependent. We can't remove the spec reporter without losing local visibility, and we can't remove the JUnit reporter without breaking the CI upload step.

PR #2572, opened 2026-05-01
chore: drop Node.js 18 support for relevant packages

WilliamBergamin:
### Summary

Node.js 18 reached end-of-life on April 30, 2025. This raises the minimum supported Node.js version to 20 for `@slack/cli-hooks`, `@slack/cli-test`, `@slack/logger`, `@slack/oauth`, `@slack/socket-mode`, `@slack/types`,`@slack/web-api` and `@slack/webhook`, ahead of the upcoming axios-to-fetch migration.                                                                 
- Updated engines.node and @types/node from 18 to 20 in both packages
- Split the CI test step so Node 18 runs only the packages that still support it
- All other packages remain unchanged and continue to support Node 18  

### Requirements 

### DIFF
--- a/packages/cli-hooks/package.json
+++ b/packages/cli-hooks/package.json
@@ -20,7 +20,7 @@
     "src/start.js"
   ],
   "engines": {
-    "node": ">= 18",
+    "node": ">= 20",
     "npm": ">= 8.6.0"
   },
   "repository": {
--- a/packages/cli-hooks/package.json
+++ b/packages/cli-hooks/package.json
@@ -21,7 +21,7 @@
   ],
   "engines": {
     "node": ">= 20",
-    "npm": ">= 9.6.4"
+    "npm": ">=9.6.4"
   },
   "repository": {
     "type": "git",
--- a/packages/cli-hooks/package.json
+++ b/packages/cli-hooks/package.json
@@ -21,7 +21,7 @@
   ],
   "engines": {
     "node": ">= 20",
-    "npm": ">= 8.6.0"
+    "npm": ">= 9.6.4"
   },
   "repository": {
     "type": "git",
--- a/packages/rtm-api/package.json
+++ b/packages/rtm-api/package.json
@@ -22,7 +22,7 @@
     "dist/**/*"
   ],
   "engines": {
-    "node": ">=20",
+    "node": ">=18",
     "npm": ">=8.6.0"
   },
   "repository": {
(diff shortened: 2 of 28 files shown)


packages/cli-hooks/package.json
```diff
@@ -20,7 +20,7 @@
     "src/start.js"
   ],
   "engines": {
-    "node": ">= 18",
+    "node": ">= 20",
```
zimeg, line 23: 🔭 note(non-blocking): We might want to find the LTS version of this minimum? Perhaps not for `20` but in future updates this might guarantee more stable app code:

> This release marks the transition of Node.js 24.x into Long Term Support (LTS).

🔗 https://nodejs.org/en/blog/release/v24.11.0

packages/rtm-api/package.json
zimeg, line 1: 🧊 suggestion: We might want to keep this package frozen?

> Beginning **March 31, 2025**, we will discontinue support for legacy custom bots.

🔗 https://docs.slack.dev/changelog/2024-09-legacy-custom-bots-classic-apps-deprecation/

packages/types/package.json
```diff
@@ -16,8 +16,8 @@
     "dist/**/*"
   ],
   "engines": {
-    "node": ">= 12.13.0",
-    "npm": ">= 6.12.0"
+    "node": ">= 20",
```
zimeg, line 19: ⭐ praise: I understand this is a significant change but should make maintenance much better ongoing!

🪬 thought: We might be confident that developers update to the latest package versions and using the latest Slack features are using more recent versions of `node` too?

packages/cli-hooks/package.json
```diff
@@ -20,7 +20,7 @@
...
     "npm": ">= 8.6.0"
```
zimeg:
```suggestion
    "npm": ">= 9.6.4"
```

🎁 suggestion: Let's match the version packaged here too!

🔗 https://nodejs.org/en/download/archive/v20.0.0

zimeg (approved):
@WilliamBergamin Huge thanks for getting this started! Let's note this toward https://github.com/slackapi/node-slack-sdk/issues/2359 🐙 

I'm approving now but left a few thoughts and comments. Before merging I'm most curious about:

- Freezing the `@slack/rtm-api` package
- Increasing the minimum `npm` version

The standard packages of these later versions are so exciting to unlock 🔏 🦕

packages/rtm-api/package.json
WilliamBergamin: Good point 💯

packages/types/package.json
WilliamBergamin: I'm not sure of the use case where this package is used standalone, from what I understand it should be used in combination with another one of our packages that now require node 20 🤔

packages/cli-hooks/package.json
WilliamBergamin: Good catch I did not think of this 🚀

Issue #775, opened 2019-04-17
WebAPICallResult and the error handling is not intuitive


lenkan:
### Description

This is not really a bug, but rather a feature of the API that caught me off guard. I think there is room for improvement for how errors are propagated back to the user from the WebClient. See "Steps to reproduce" for an explanation of what surprised me.

### What type of issue is this? (place an `x` in one of the `[ ]`)
- [ ] bug
- [ ] enhancement (feature request)
- [ ] question
- [ ] documentation related
- [ ] testing related
- [x] discussion

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/master/.github/contributing.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

---

### Bug Report

Filling out the following details about bugs will help us solve your issue sooner.

#### Packages:

Select all that apply:

- [x] `@slack/web-api`
- [ ] `@slack/events-api`
- [ ] `@slack/interactive-messages`
- [ ] `@slack/rtm-api`
- [ ] `@slack/webhooks`
- [ ] I don't know

#### Reproducible in:

package version: 4.12.0

node version: 10.15.3

OS version(s): Arch Linux 5.0.7

#### Steps to reproduce:

1. Call `users.lookupByEmail` for a team where the email is _not_ registered.

```javascript
const result = await client.users.lookupByEmail({ email: 'foo@bar.com' })
```

#### Expected result:

Since the interface `WepAPICallResult` has the following shape, in particular the `error` and `ok` properties: 

```javascript
export interface WebAPICallResult {
  ok: boolean;
  error?: string;
  response_metadata?: {
    warnings?: string[];
    next_cursor?: string;
    scopes?: string[];
    acceptedScopes?: string[];
    retryAfter?: number;
  };
  [key: string]: unknown;
}
```

I expect that the call to `lookupByEmail` should _resolve_ to an object similar to:

```
{
   error: 'users_not_found',
   ok: false,
   response_metadata: [...]
}
```

I.e., the `error` property indicates that there was an error executing the particular lookup.

#### Actual result:

The call _rejected_ to an _Error_ object containing:

```
{ 
  message: 'An API error occurred: users_not_found',
  code: 'slackclient_platform_error',
  data: { 
[...]

PR #2593, opened 2026-05-12
feat: replace error interfaces with proper Error subclasses

WilliamBergamin:
## Summary

These changes aim to replace error interfaces and unsafe factory functions with proper `Error` subclasses in `@slack/web-api` and `@slack/webhook`, enabling `instanceof` checks and better `TypeScript` narrowing.

Aims to resolves https://github.com/slackapi/node-slack-sdk/issues/775

## Motivation

Issue #775 identified a fundamental DX problem: the `WebAPICallResult` interface declared an `error?: string` field that was never populated on resolved values (since the client throws on `ok: false`). This led developers to write dead `if (result.error)` branches that `TypeScript` couldn't catch.

More broadly, the SDK's error types were implemented as plain interfaces with factory functions that used unsafe `as Partial<XxxError>` casting. This meant:
- `instanceof` never worked (users could only check `error.code === ErrorCode.X`)
- Error objects had `name: 'Error'` instead of descriptive class names
- The `attachOriginalToWebAPIRequestError` option existed solely to work around `axios`'s large error objects

## Changes
- **Error classes**: `SlackError` (web-api) and `SlackWebhookError` (webhook) abstract base classes with concrete subclasses for each error type
 - **`instanceof` support**: errors are real class instances with proper prototype chains (`Object.setPrototypeOf` for ES5 compat)
 - **`Error.cause`**: request errors set the standard `cause` property (Node 18+) alongside the `original` field for backward compatibility
 - **Structured webhook errors**: `IncomingWebhookHTTPError` now exposes `statusCode`, `statusMessage`, and `body` directly instead of burying them inside an `original` `axios` error
 - **New file upload error type**: `WebAPIFileUploadReadFileDataError` (the `ErrorCode` existed but had no corresponding type)

 ## Backward Compatibility

 | What still works | What breaks |
 |------------------|-------------|
 | `error.code === ErrorCode.PlatformError` | `instanceof` against the old interfaces (they were interfaces, so nobody could have used `instanceof` anyway) |
 | All `ErrorCode` enum string values | Factory functions removed (were internal, not exported) |
 | `error.original` on request errors | `IncomingWebhookHTTPError.original` removed — use `.statusCode`/`.body` instead |
 | | `error.name` is now the class name (e.g. `'WebAPIPlatformError'`) instead of `'Error'` |

### Requirements 


packages/web-api/src/errors.ts
```diff
@@ -23,119 +23,83 @@ export enum ErrorCode {
...
 
-export interface WebAPIHTTPError extends CodedError {
-  code: ErrorCode.HTTPError;
-  statusCode: number;
-  statusMessage: string;
-  headers: Record<string, string>;
-  // biome-ignore lint/suspicious/noExplicitAny: HTTP response bodies might be anything
-  body?: any;
-}
+export class WebAPIPlatformError extends SlackError {
```
zimeg, line 38: 🌟 praise: Goodness I forget if this was so clear before these changes but I find this a delight!

```js
(async () => {
  try {
    await client.usergroups.list();
  } catch (error) {
    if (error instanceof WebAPIPlatformError) {
      app.logger.error(error.data);
    } else {
      app.logger.error(error);
    }
  }
})();
```

```
[ERROR]  bolt-app {
  ok: false,
  error: 'missing_scope',
  needed: 'usergroups:read',
  provided: 'channels:history,chat:write,commands',
  response_metadata: {
    scopes: [ 'channels:history', 'chat:write', 'commands' ],
    acceptedScopes: [ 'usergroups:read' ]
  }
}
```

zimeg (approved):
@WilliamBergamin Superb changes and thanks for taking this on! I'm quite excited for the patterns this supports.

PR #2611, opened 2026-05-29
feat(web-api): expose public ts property on ChatStream

srtaalej:
### Summary
This PR adds a public read-only ts getter to ChatStreamer, enabling users to call chat.update as a fallback when the server closes a stream due to undocumented timeouts.

### Background

When Slack's server-side timeout kills a stream, both `chat.appendStream` and `chat.stopStream` return `message_not_in_streaming_state`. The message still exists but is stuck as a broken pill in the UI. The only recovery path is `chat.update({ ts: ... })`, but `ts` was previously a private field (`streamTs`).

### Changes
- `packages/web-api/src/chat-stream.ts`: Added `ts` getter (read-only, returns `string | undefined`)
- `packages/web-api/src/WebClient.test.ts`: Added test verifying `ts` is undefined before flush and set after
  
 
<details><summary> code snippet </summary>

```js
  import { WebClient } from '@slack/web-api';

  async function sendStreaming(channel: string, threadTs: string, teamId: string, userId: string, 
  chunks: AsyncIterable<string>) {
    const streamer = client.chatStream({
      channel,
      thread_ts: threadTs,
      recipient_team_id: teamId,
      recipient_user_id: userId,
    });

    let fullText = '';
    let streamAlive = true;

    for await (const chunk of chunks) {
      fullText += chunk;

      if (streamAlive) {
        try {
          await streamer.append({ markdown_text: chunk });
        } catch (error) {
          if (error.data?.error === 'message_not_in_streaming_state') {
            streamAlive = false;
          } else {
            throw error;
          }
        }
      }

      if (!streamAlive) {
        await client.chat.update({ channel, ts: streamer.ts, text: fullText });
      }
    }

    if (streamAlive) {
      await streamer.stop();
    } else {
      await client.chat.update({ channel, ts: streamer.ts, text: fullText });
    }
  }
```

</details>

### Requirements 

### DIFF
--- a/packages/web-api/src/chat-stream.ts
+++ b/packages/web-api/src/chat-stream.ts
@@ -55,6 +55,16 @@ export class ChatStreamer {
   }
 
   /**
+   * @description The message timestamp of the stream. Returns `undefined` until the first flush
+   * (when `chat.startStream` is called). Can be used with `chat.update` as a fallback if the
+   * stream expires server-side.
+   * @see {@link https://docs.slack.dev/reference/methods/chat.update}
+   */
+  get ts(): string | undefined {
+    return this.streamTs;
+  }
+
+  /**
    * Append to the stream.
    *
    * @description The "append" method appends to the chat stream being used. This method can be called multiple times. After the stream is stopped this method cannot be called.
--- a/packages/web-api/src/chat-stream.ts
+++ b/packages/web-api/src/chat-stream.ts
@@ -56,8 +56,7 @@ export class ChatStreamer {
 
   /**
    * @description The message timestamp of the stream. Returns `undefined` until the first flush
-   * (when `chat.startStream` is called). Can be used with `chat.update` as a fallback if the
-   * stream expires server-side.
+   * (when `chat.startStream` is called).
    * @see {@link https://docs.slack.dev/reference/methods/chat.update}
    */
   get ts(): string | undefined {
--- a/.changeset/expose-ts-on-chat-streamer.md
+++ b/.changeset/expose-ts-on-chat-streamer.md
@@ -1,0 +1,24 @@
+---
+"@slack/web-api": minor
+---
+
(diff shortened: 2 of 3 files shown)


packages/web-api/src/chat-stream.ts
```diff
@@ -54,6 +54,16 @@ export class ChatStreamer {
     this.streamArgs = args;
   }
 
+  /**
+   * @description The message timestamp of the stream. Returns `undefined` until the first flush
+   * (when `chat.startStream` is called). Can be used with `chat.update` as a fallback if the
+   * stream expires server-side.
+   * @see {@link https://docs.slack.dev/reference/methods/chat.update}
+   */
+  get ts(): string | undefined {
```
WilliamBergamin, line 62: I think this is one of the rare times we use the `get` operator in this project 🤔 I think it makes sense to use it here, it should give us the flexibly to add a setter if we ever need to and exposes the property as `readonly`

packages/web-api/src/chat-stream.ts
```diff
@@ -54,6 +54,16 @@ export class ChatStreamer {
     this.streamArgs = args;
   }
 
+  /**
+   * @description The message timestamp of the stream. Returns `undefined` until the first flush
+   * (when `chat.startStream` is called). Can be used with `chat.update` as a fallback if the
+   * stream expires server-side.
```
WilliamBergamin: What do you think about keeping this comment "simple"?

```suggestion
   * (when `chat.startStream` is called).
```

WilliamBergamin (approved):
Nice work 💯 changes seem good to me! I left one addressable non blocking comment

packages/web-api/src/chat-stream.ts
srtaalej: sure!

packages/web-api/src/chat-stream.ts
srtaalej: yes i agree - best to keep it readonly!

PR #2615, opened 2026-06-02
feat(webhook): add WebhookTrigger class for Workflow Builder triggers

zimeg:
### Summary

This pull request adds a minimal `WebhookTrigger` class to `@slack/webhook` for Workflow Builder webhook triggers.

The package previously only supported incoming webhooks (`IncomingWebhook`). Workflow Builder triggers accept arbitrary JSON payloads and return JSON responses, so consumers like `slack-github-action` currently fall back to raw `fetch`.

The implementation is intentionally small and mirrors `IncomingWebhook`:

- Constructor takes a URL and optional defaults (`timeout`, `agent`); rejects a missing or empty URL.
- `send(payload?)` POSTs the payload (defaults to `{}`) and resolves to the parsed JSON response body (`WebhookTriggerResult`, typed `{ ok, error? }`); HTTP and request failures reject.
- Reuses the existing error infrastructure and User-Agent instrumentation, adding `WebhookTrigger`-named error types.

#### Notes for reviewers

- **URL guard:** the constructor uses a falsy check (`!url`), so it rejects an empty string as well as `undefined`. This intentionally differs from `IncomingWebhook`, which only guards `undefined`.
- **Result typing:** `WebhookTriggerResult` is typed closed as `{ ok: boolean; error?: string }` — matching what the trigger endpoint returns (`{ ok: true }` or `{ ok: false, error }`). Widening this later is backward-compatible; tightening a released open type would not be.

### Preview

🔗 https://github.com/slackapi/node-slack-sdk/tree/eden/webhook-trigger/packages/webhook#trigger-a-workflow-builder-workflow

```js
const { WebhookTrigger } = require('@slack/webhook');
const url = process.env.SLACK_WEBHOOK_TRIGGER_URL;

const trigger = new WebhookTrigger(url);

(async () => {
  // Keys should match the variables your workflow expects
  const result = await trigger.send({
    customer_name: 'Ada Lovelace',
    order_id: '1024',
  });
})();
```

### Requirements 


WilliamBergamin (approved):
Gave this a quick look and it looks good 💯 

if we need to dry this up in the future we can look into it at that time

packages/webhook/src/errors.ts
```diff
@@ -16,16 +16,19 @@ export enum ErrorCode {
 }
 
 export type IncomingWebhookSendError = IncomingWebhookRequestError | IncomingWebhookHTTPError;
+export type WebhookTriggerSendError = WebhookTriggerRequestError | WebhookTriggerHTTPError;
```
zimeg, line 19: ⚠️ note: The error handling is shared between webhook techniques but so we export these values under a similar namespace.

packages/webhook/src/WebhookTrigger.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+      throw error;
+    }
+  }
+
+  /**
+   * Processes an HTTP response into a WebhookTriggerResult.
+   */
+  private buildResult(response: AxiosResponse): WebhookTriggerResult {
+    const body = typeof response.data === 'string' ? JSON.parse(response.data) : response.data;
```
zimeg: 🔬 note: This feels defensive to me and might be removed as `response` is unpacked to match standard "fetch"!

zimeg (commented):
🗣️ Words I think for the readers and reviewers!

packages/webhook/src/errors.ts
```diff
@@ -16,16 +16,19 @@ export enum ErrorCode {
...
 
 export interface IncomingWebhookRequestError extends CodedError {
   code: ErrorCode.RequestError;
   original: Error;
 }
+export type WebhookTriggerRequestError = IncomingWebhookRequestError;
```
WilliamBergamin, line 25: Safe move 🧠  I like it 💯

packages/webhook/src/WebhookTrigger.test.ts
```diff
@@ -0,0 +1,141 @@
...
+          // biome-ignore lint/suspicious/noExplicitAny: reading the wrapped axios response body
+          const response = (httpError.original as any).response;
+          assert.strictEqual(response.status, 401);
+          assert.deepStrictEqual(response.data, { ok: false, error: 'invalid_auth' });
+        }
+        scope.done();
+      });
+    });
+
+    describe('User-Agent header', () => {
```
WilliamBergamin, line 121: Nice 💯 

should we also add a tests that covers using `addAppMetadata` to expand the user-agent value?

packages/webhook/src/WebhookTrigger.ts
```diff
@@ -0,0 +1,106 @@
...
+import axios, { type AxiosInstance, type AxiosResponse } from 'axios';
+
+import { httpErrorWithOriginal, requestErrorWithOriginal } from './errors';
+import { getUserAgent } from './instrument';
+
+/**
+ * A client for Slack's Workflow Builder webhook triggers
+ * @see {@link https://slack.com/help/articles/360041352714-Build-a-workflow--Create-a-workflow-that-starts-outside-of-Slack}
+ */
+export class WebhookTrigger {
```
WilliamBergamin, line 12: Paise for mirroring existing behaviors 💯

WilliamBergamin (approved):
Nice work 💯

Left one comment on test coverage but its possible my comment is redundent

packages/webhook/src/WebhookTrigger.test.ts
zimeg: @WilliamBergamin Thanks for calling this out! We're introducing a test in https://github.com/slackapi/node-slack-sdk/pull/2615/changes/3bed3dcf060ca2eae57e25f9352999095afa6c5c for this.

packages/webhook/src/WebhookTrigger.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (error.response !== undefined) {
+        throw httpErrorWithOriginal(error);
+      }
+      if (error.request !== undefined) {
+        throw requestErrorWithOriginal(error);
+      }
+      throw error;
+    }
+  }
+}
```
zimeg, line 76: 👁️‍🗨️ note: Our return inlines the response values without more change to mirror implementations of the `@slack/web-api` package a bit more!

packages/webhook/src/WebhookTrigger.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export interface WebhookTriggerSendArguments {
+  [key: string]: string;
+}
+
+export interface WebhookTriggerResult {
+  ok: boolean;
+  error?: string;
+}
```
zimeg, line 94: 🪬 note: These values are more strict now too!

packages/webhook/README.md
zimeg, line 1: 📚 note: I hoped to generalize this package `README` for now with added information but think we might revise this ongoing?

zimeg (commented):
🔏 A few more comments of confidence!

packages/webhook/src/WebhookTrigger.ts
WilliamBergamin: This looks good 💯

If we need to add more fields in the future we can

WilliamBergamin (approved):
Ship it 🚀

zimeg:
@WilliamBergamin So exciting! Thanks for sharing big ideas and kind thoughts toward this 🧠 💡

PR #2624, opened 2026-06-10
fix(cli-test)!: remove default "--app deployed" global flag from commands

srtaalej:
### Summary

Skip passing a default `--app deployed` flag to commands.

### Notes

- CLI commit [c9f5bb](http://github.com/slackapi/slack-cli/pull/565/changes/c9f5bbf6706c4b3aab17f2b5bcdb6e43f07e9877) in [PR 565](http://github.com/slackapi/slack-cli/pull/565) motivates this change after introducing validation that rejects non-app-ID values (`local, deployed`) when using `--app` with `create`.

### Requirements 

### DIFF
--- a/.changeset/silver-papers-lick.md
+++ b/.changeset/silver-papers-lick.md
@@ -1,5 +1,5 @@
 ---
-"@slack/cli-test": patch
+"@slack/cli-test": major
 ---
 
 fix: remove default --app global flag from create command tests
--- a/.changeset/silver-papers-lick.md
+++ b/.changeset/silver-papers-lick.md
@@ -1,0 +1,5 @@
+---
+"@slack/cli-test": patch
+---
+
+fix: remove default --app global flag from create command tests
--- a/.changeset/silver-papers-lick.md
+++ b/.changeset/silver-papers-lick.md
@@ -2,4 +2,16 @@
 "@slack/cli-test": major
 ---
 
-fix: remove default --app global flag from create command tests
+fix: remove default "--app deployed" global flag from commands
+
+Commands running for a specific app must now provide the "app" argument:
+
+```diff
+  await SlackCLI.app.delete({
+    appPath: "my-app",
+    team: "T0123456789",
++   app: "deployed",
+  });
+```
+
+The options "local" or "deployed" or app ID all remain available to use.
--- a/packages/cli-test/src/cli/cli-process.test.ts
+++ b/packages/cli-test/src/cli/cli-process.test.ts
@@ -98,18 +98,22 @@ describe('SlackCLIProcess class', () => {
         await cmd.execAsync();
         sandbox.assert.calledWith(spawnProcessSpy, sinon.match.string, sinon.match.array.contains(['--skip-update']));
       });
-      it('should default to `--app deployed` but allow overriding that via the `app` parameter', async () => {
+      it('should only pass `--app` when explicitly provided via the `app` parameter', async () => {
         let cmd = new SlackCLIProcess(['help']);
         await cmd.execAsync();
-        sandbox.assert.calledWith(
+        sandbox.assert.neverCalledWith(
           spawnProcessSpy,
           sinon.match.string,
-          sinon.match.array.contains(['--app', 'deployed']),
+          sinon.match.array.contains(['--app']),
         );
         spawnProcessSpy.resetHistory();
         cmd = new SlackCLIProcess(['help'], { app: 'local' });
         await cmd.execAsync();
         sandbox.assert.calledWith(spawnProcessSpy, sinon.match.string, sinon.match.array.contains(['--app', 'local']));
+        spawnProcessSpy.resetHistory();
+        cmd = new SlackCLIProcess(['help'], { app: 'deployed' });
+        await cmd.execAsync();
+        sandbox.assert.calledWith(spawnProcessSpy, sinon.match.string, sinon.match.array.contains(['--app', 'deployed']));
       });
       it('should default to `--force` but allow overriding that via the `force` parameter', async () => {
         let cmd = new SlackCLIProcess(['help']);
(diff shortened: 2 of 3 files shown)


packages/cli-test/src/cli/cli-process.ts
```diff
@@ -132,10 +132,12 @@ export class SlackCLIProcess {
         cmd = cmd.concat(['--team', opts.team]);
       }
       // App instance; defaults to `deployed`
```
zimeg:
```suggestion
      // App instance; defaults to `deployed` - TODO(semver:major): remove fallback to "deployed" app
```

🔬 suggestion: Let's note that we want to move away from "deploy" or "local" options here since this default doesn't make sense for a handful of commands

zimeg: 🏁 ramble: Later CLI updates might start to error for unexpected flags too. I understand "app" is considered global at this time but it'd be nice to remain flexible for later updates. Not a blocker now if a fast fix is alright to merge?

zimeg (approved):
@srtaalej LGTM for a quick fix but we should follow up with fewer defaults overall to avoid similar errors soon

packages/cli-test/src/cli/cli-process.ts
```diff
@@ -132,10 +132,12 @@ export class SlackCLIProcess {
...
-      if (opts.app) {
-        cmd = cmd.concat(['--app', opts.app]);
-      } else {
-        cmd = cmd.concat(['--app', 'deployed']);
+      if (this.command[0] !== 'create') {
```
zimeg: 🦠 question: Oh does this block "app" flag from the "create" command altogetther?

zimeg (commented):
QQ before merge I might've tested just the current defaults

packages/cli-test/src/cli/cli-process.ts
srtaalej: hmm will keep that in mind. lets move ahead with a quick fix for now and come back to this in potential followups 🤔

packages/cli-test/src/cli/cli-process.ts
srtaalej: yes i believe it does! should it only block when the flag is passed with local and deployed?

zimeg: @srtaalej Hm I think we should remove `app` defaults altogether instead of extending the special cases to guard against. It's uncomfortable to add implementation logic to this package IMHO

srtaalej: sounds good 🚀

.changeset/silver-papers-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@slack/cli-test": patch
```
zimeg:
```suggestion
"@slack/cli-test": major
```

📣 suggestion: The removed default is worth calling out here! Using a major version should stop our CI from accepting this change without rebuilding a base image for a more seamless update?

📚 ramble: I'll include required steps following to make this update more clear in `CHANGELOG.md` for agents and kind readers alike.

.changeset/silver-papers-lick.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+fix: remove default --app global flag from create command tests
```
zimeg:
``````suggestion
fix: remove default "--app deployed" global flag from commands

Commands running for a specific app must now provide the "app" argument:

```diff
  await SlackCLI.app.delete({
    appPath: "my-app",
    team: "T0123456789",
+   app: "deployed",
  });
```

The options "local" or "deployed" or app ID all remain available to use.
``````

File: packages/cli-test/src/cli/cli-process.test.ts [hunk 5]
```diff
@@ -98,18 +98,22 @@ describe('SlackCLIProcess class', () => {
         await cmd.execAsync();
         sandbox.assert.calledWith(spawnProcessSpy, sinon.match.string, sinon.match.array.contains(['--skip-update']));
       });
-      it('should default to `--app deployed` but allow overriding that via the `app` parameter', async () => {
+      it('should only pass `--app` when explicitly provided via the `app` parameter', async () => {
         let cmd = new SlackCLIProcess(['help']);
         await cmd.execAsync();
+        sandbox.assert.neverCalledWith(spawnProcessSpy, sinon.match.string, sinon.match.array.contains(['--app']));
```
zimeg, line 104: 🧪 praise: Thanks for covering these cases!

packages/cli-test/src/cli/cli-process.ts
```diff
@@ -131,11 +131,9 @@ export class SlackCLIProcess {
       if (opts.team) {
         cmd = cmd.concat(['--team', opts.team]);
       }
-      // App instance; defaults to `deployed`
+      // App instance - TODO(semver:major): remove fallback to "deployed" app
```
zimeg:
```suggestion
      // App instance
```

🪓 suggestion: Let's remove this with these changes following!

packages/cli-test/src/cli/cli-process.ts
```diff
@@ -149,9 +147,10 @@ export class SlackCLIProcess {
         cmd = cmd.concat(['--verbose']);
       }
     } else {
-      cmd = cmd.concat(['--skip-update', '--force', '--app', 'deployed']);
```
zimeg, line 152: 🪓 praise: Nice!

🔮 note: I have hope to split "force" into more meaningful flags and perhaps we use an environment variable to skip updates in later changes. I'm hoping we avoid having defaults exposed for each command.

zimeg (approved):
🏁 @srtaalej A few suggestions and changes to the PR metadata to take this to a finished line. These changes are healthy for the package I think and I comment more on that too!

PR #2641, opened 2026-07-02
feat(webhook): add opt-in retries to IncomingWebhook and WebhookTrigger

zimeg:
### Summary

Adds opt-in retry support to `@slack/webhook`, mirroring the `retryConfig` convention already used by `@slack/web-api`'s `WebClient`. Stacked on top of #2615 (the `WebhookTrigger` class), so this PR is scoped to just the retry change.

- Both `IncomingWebhook` and `WebhookTrigger` accept a new `retryConfig?: RetryOptions` constructor option.
- Adds `retry-policies.ts` with the same named policies exported by `@slack/web-api` — `fiveRetriesInFiveMinutes`, `tenRetriesInAboutThirtyMinutes`, `rapidRetryPolicy` — re-exported from the package entry point.
- Each `send()` is wrapped in `p-retry` using the configured policy.

**Default:** no retries (`{ retries: 0 }`) when `retryConfig` is unset — preserves today's behavior; retries are opt-in.

**Retry conditions** (transient failures only):
- Server errors (`5xx`) → retry
- Network errors with no response → retry
- All client errors (`4xx`), including rate limits (`429`) → abort immediately

Unlike `@slack/web-api`, `429` responses are **not** retried here. That client blocks its request queue to honor the `Retry-After` header; the webhook clients have no such queue, so a blind retry would ignore the server's backoff. Rate-limit handling is left to the caller, and all `4xx` responses are treated the same.

`retryConfig` is a transport-only option and is stripped from the posted payload (like `agent`).

#### Notes for reviewers

- **429 abort is intentional and test-covered** — see the `does not retry a 429 even when a retry policy is set` cases in both `IncomingWebhook.test.ts` and `WebhookTrigger.test.ts`. The rationale is above (no request queue to honor `Retry-After`).
- **`@types/retry` is pinned to `0.12.0` against `retry ^0.13.1`** — this matches `@slack/web-api`'s own pinning. `0.12.0` is the latest published version of the type definitions, and the `OperationOptions` interface we extend for `RetryOptions` is unchanged across `0.12`/`0.13`.
- **`retry-policies.ts` mirrors `@slack/web-api`'s** rather than sharing it, matching the existing per-package convention in this monorepo.

Consumed downstream by slackapi/slack-github-action#630, which passes its `0/5/10/RAPID` input through to these policies.

### Requirements

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/webhook/src/retry-policies.ts
zimeg, line 1: ✂️ note: Matches the `@slack/web-api` implementation:

https://github.com/slackapi/node-slack-sdk/blob/f006ad409713741c5d398669c69ad98a7e090879/packages/web-api/src/retry-policies.ts

packages/webhook/package.json
```diff
@@ -37,13 +37,16 @@
...
+    "test:coverage": "npm run build && node --experimental-test-coverage --test-reporter=spec --test-reporter-destination=stdout --test-reporter=lcov --test-reporter-destination=lcov.info --test-reporter=junit --test-reporter-destination=test-results.xml --import tsx --test src/IncomingWebhook.test.ts src/WebhookTrigger.test.ts src/instrument.test.ts src/retry-policies.test.ts"
   },
   "dependencies": {
     "@slack/types": "^2.20.1",
     "@types/node": ">=18",
-    "axios": "^1.16.0"
+    "@types/retry": "0.12.0",
+    "axios": "^1.16.0",
+    "p-retry": "^4",
+    "retry": "^0.13.1"
```
zimeg: 🔍 note: These packages align with the `@slack/web-api` implementations:

https://github.com/slackapi/node-slack-sdk/blob/12b94a2c21268403c3f216e88232a841cfb47571/packages/web-api/package.json#L61-L62

packages/webhook/src/IncomingWebhook.ts
```diff
@@ -50,7 +58,9 @@ export class IncomingWebhook {
       },
     });
 
+    // Strip transport-only options so they do not leak into the posted payload.
     this.defaults.agent = undefined;
+    this.defaults.retryConfig = undefined;
```
zimeg, line 63: 📝 note: The default options passed to requests isn't ideal practice IMHO but we avoid regressions with this line.

zimeg (commented):
🔏 More words for the amazing reviewers.

packages/webhook/src/WebhookTrigger.test.ts
```diff
@@ -137,5 +138,67 @@ describe('WebhookTrigger', () => {
...
+        const trigger = new WebhookTrigger(url, { retryConfig: rapidRetryPolicy });
+        try {
+          await trigger.send({ key: 'value' });
+          assert.fail('expected rejection');
+        } catch (error) {
+          assert.strictEqual((error as CodedError).code, ErrorCode.HTTPError);
+        }
+        // Only one interceptor is registered; a retry would leave it unmatched
+        // and scope.done() would throw.
+        scope.done();
```
WilliamBergamin, line 185: Praise 💯

packages/webhook/src/WebhookTrigger.ts
```diff
@@ -59,19 +67,25 @@ export class WebhookTrigger {
...
+      try {
+        const response = await this.axios.post(this.url, payload);
+        return this.buildResult(response);
+        // biome-ignore lint/suspicious/noExplicitAny: errors can be anything
+      } catch (error: any) {
+        if (error.response !== undefined) {
+          const status: number = error.response.status;
+          const retryable = status === 429 || status >= 500;
+          const wrapped = httpErrorWithOriginal(error);
+          throw retryable ? wrapped : new AbortError(wrapped);
```
WilliamBergamin: Same here

```suggestion
          const wrapped = httpErrorWithOriginal(error);
          throw status >= 500 ? wrapped : new AbortError(wrapped);
```

packages/webhook/src/IncomingWebhook.ts
```diff
@@ -67,20 +77,26 @@ export class IncomingWebhook {
...
+        const response = await this.axios.post(this.url, payload);
+        return this.buildResult(response);
+        // biome-ignore lint/suspicious/noExplicitAny: errors can be anything
+      } catch (error: any) {
+        // Wrap errors in this packages own error types (abstract the implementation details' types)
+        if (error.response !== undefined) {
+          const status: number = error.response.status;
+          const retryable = status === 429 || status >= 500;
+          const wrapped = httpErrorWithOriginal(error);
+          throw retryable ? wrapped : new AbortError(wrapped);
```
WilliamBergamin: My mistake here, I may have mislead you on this 😅 

`429` status codes usually come with a retry header that informs the client on when it is appropriate to  send requests. The `web-api` handles this by [blocking the request queue of the client](https://github.com/slackapi/node-slack-sdk/blob/a795b86864cb3b37b49cc938cc73d33d0bd130a4/packages/web-api/src/WebClient.ts#L725-L746)

The `webhook` package does not have the concept of request queues, and no one asked for it 🤔 Currently `429` responses are handled by the user, since we aren't taking into account the `retry header` in this retry implementation what do you think about treating all `4**` responses the same way?

```suggestion
          const wrapped = httpErrorWithOriginal(error);
          throw status >= 500 ? wrapped : new AbortError(wrapped);
```

WilliamBergamin (changes requested):
This looks great already 🚀 

I think I might have mislead you around  handling 429 responses, left a comment about this 😅

packages/webhook/src/IncomingWebhook.ts
zimeg: @WilliamBergamin Ah I'm learning more about HTTP! Thanks for sharing this!

We're landing a change in https://github.com/slackapi/node-slack-sdk/pull/2641/commits/3d9a0af4e6c2c859dc55c1d05c689349d77727e5 to save retries for server errors at this time. I agree it'd be alright to wait for feedback before extending this more.

packages/webhook/src/WebhookTrigger.ts
zimeg: Similar changes to comment below we land this in https://github.com/slackapi/node-slack-sdk/pull/2641/commits/3d9a0af4e6c2c859dc55c1d05c689349d77727e5! 🚢 💨

packages/webhook/src/index.ts
```diff
@@ -20,6 +20,13 @@ export {
 
 export { addAppMetadata } from './instrument';
 
+export {
+  fiveRetriesInFiveMinutes,
+  RetryOptions,
+  rapidRetryPolicy,
+  tenRetriesInAboutThirtyMinutes,
+} from './retry-policies';
```
zimeg: 🌚 note: We might prefer to export this similar to how `@slack/web-api` is structured:

https://github.com/slackapi/node-slack-sdk/blob/b0e335b68e92c22a5195fe8ce6d05d3862690531/packages/web-api/src/index.ts#L17

packages/webhook/src/index.ts
```diff
@@ -20,6 +20,8 @@ export {
 
 export { addAppMetadata } from './instrument';
 
+export { default as retryPolicies, RetryOptions } from './retry-policies';
```
zimeg, line 23: 🎁 note: Here we match the export pattern of the `@slack/web-api` package:

https://github.com/slackapi/node-slack-sdk/blob/b0e335b68e92c22a5195fe8ce6d05d3862690531/packages/web-api/src/index.ts#L17

zimeg (commented):
🔭 Duplicated comments from changed code!

packages/webhook/src/IncomingWebhook.ts
```diff
@@ -67,20 +77,25 @@ export class IncomingWebhook {
...
+      try {
+        const response = await this.axios.post(this.url, payload);
+        return this.buildResult(response);
+        // biome-ignore lint/suspicious/noExplicitAny: errors can be anything
+      } catch (error: any) {
+        // Wrap errors in this packages own error types (abstract the implementation details' types)
+        if (error.response !== undefined) {
+          const status: number = error.response.status;
+          const wrapped = httpErrorWithOriginal(error);
+          throw status >= 500 ? wrapped : new AbortError(wrapped);
```
WilliamBergamin, line 90: Nice 💯  thanks for simplifying this 🚀

packages/webhook/src/index.ts
WilliamBergamin: Praise for matching the pattern 🚀

packages/webhook/src/WebhookTrigger.ts
```diff
@@ -59,19 +67,24 @@ export class WebhookTrigger {
...
+    return pRetry(async () => {
+      try {
+        const response = await this.axios.post(this.url, payload);
+        return response.data;
+        // biome-ignore lint/suspicious/noExplicitAny: errors can be anything
+      } catch (error: any) {
+        if (error.response !== undefined) {
+          const status: number = error.response.status;
+          const wrapped = httpErrorWithOriginal(error);
+          throw status >= 500 ? wrapped : new AbortError(wrapped);
```
WilliamBergamin, line 79: Praise 💯

WilliamBergamin (approved):
Ship it 🚀

zimeg:
@WilliamBergamin Will do! 🫡 I look forward to bringing this to the Slack GitHub Action after an upcoming release.

PR #2652, opened 2026-07-08
fix(web-api): redact response body in debug logs

### LINKED ISSUE #2651: Response body not redacted in debug logs
## Description

The `redact()` helper in `WebClient.ts` is applied to outgoing request bodies and headers before debug logging, but is **not** applied to incoming API response bodies. This means tokens or other sensitive values present in API responses are logged in plaintext when debug logging is enabled.

## Location

`packages/web-api/src/WebClient.ts`

- **Line 702** — request body is redacted: `this.logger.debug(\`http request body: \${JSON.stringify(redact(body))}\`)`
- **Line 721** — request headers are redacted: `this.logger.debug(\`http request headers: \${JSON.stringify(redact(allHeaders))}\`)`
- **Line 374** — response body is **not** redacted: `this.logger.debug(\`http request result: \${JSON.stringify(result)}\`)`

## Impact

When `LogLevel.DEBUG` is enabled, any tokens returned in API response bodies (e.g., `auth.test`, `oauth.v2.access`, or other methods that return tokens) are written to logs in plaintext. An attacker with access to debug logs could extract these credentials.

## Expected Behavior

The response body should be passed through `redact()` before being logged, consistent with how the request body and headers are handled.

## Suggested Fix

```typescript
this.logger.debug(`http request result: ${JSON.stringify(redact(result))}`);
```

The existing `redact()` function already handles redacting keys matching `.*token.*` and `Authorization`, so applying it to the response object should cover the relevant cases.

srtaalej:
## Summary

- Apply `redact()` to API response bodies before debug logging, consistent with how request bodies and headers are already handled
- Widen `redact()` parameter type from `Record<string, unknown>` to `object` so it accepts `WebAPICallResult` without casting
- Prevents tokens returned in API responses (e.g., `oauth.v2.access`, `openid.connect.token`, `tooling.tokens.rotate`) from being written to logs in plaintext when debug logging is enabled

## Test plan

- [x] `npm run build --workspace=packages/web-api` passes
- [x] `npm test --workspace=packages/web-api` — all 155 tests pass
- [ ] Verify manually: enable `LogLevel.DEBUG`, call `oauth.v2.access`, confirm `access_token` shows as `[[REDACTED]]` in output

Fixes #2651

### DIFF
--- a/packages/web-api/src/WebClient.test.ts
+++ b/packages/web-api/src/WebClient.test.ts
@@ -118,6 +118,21 @@ describe('WebClient', () => {
       // Calling #setName of the given logger is destructive
       assert.strictEqual((logger.setName as sinon.SinonStub).called, false);
     });
+
+    it('should redact tokens in API response bodies in debug logs', async () => {
+      const scope = nock('https://slack.com')
+        .post('/api/oauth.v2.access')
+        .reply(200, { ok: true, access_token: 'xoxb-secret-token', token_type: 'bot' });
+      const client = new WebClient(undefined, { logLevel: LogLevel.DEBUG, logger });
+      await client.apiCall('oauth.v2.access', { code: 'test-code' });
+      scope.done();
+
+      const debugCalls = (logger.debug as sinon.SinonStub).getCalls();
+      const resultLog = debugCalls.find((call) => call.args[0].includes('http request result'));
+      assert.ok(resultLog, 'expected a debug log for the http request result');
+      assert.ok(resultLog.args[0].includes('[[REDACTED]]'), 'expected access_token to be redacted');
+      assert.ok(!resultLog.args[0].includes('xoxb-secret-token'), 'expected raw token to not appear in logs');
+    });
   });
 
   describe('has an option to override the Axios timeout value', () => {
--- a/packages/web-api/src/WebClient.test.ts
+++ b/packages/web-api/src/WebClient.test.ts
@@ -122,7 +122,32 @@ describe('WebClient', () => {
     it('should redact tokens in API response bodies in debug logs', async () => {
       const scope = nock('https://slack.com')
         .post('/api/oauth.v2.access')
-        .reply(200, { ok: true, access_token: 'xoxb-secret-token', token_type: 'bot' });
+        .reply(200, {
+            "ok": true,
+            "access_token": "xoxb-secret-token",
+            "token_type": "bot",
+            "scope": "commands,incoming-webhook",
+            "bot_user_id": "U0KRQLJ9H",
+            "app_id": "A0KRD7HC3",
+            "expires_in": 43200,
+            "refresh_token": "xoxe-secret-token",
+            "team": {
+                "name": "Slack Softball Team",
+                "id": "T9TK3CUKW"
+             },
+             "enterprise": {
+               "name": "slack-sports",
+                "id": "E12345678"
+              },
+              "authed_user": {
+                "id": "U1234",
+                "scope": "chat:write",
+                "access_token": "xoxe.xoxp-secret-token",
+                "expires_in": 43200,
+                "refresh_token": "xoxe-secret-token",
+                "token_type": "user"
+              }
+            });
       const client = new WebClient(undefined, { logLevel: LogLevel.DEBUG, logger });
       await client.apiCall('oauth.v2.access', { code: 'test-code' });
       scope.done();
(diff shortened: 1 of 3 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -371,7 +371,7 @@ export class WebClient extends Methods {
       headers,
     );
     const result = await this.buildResult(response);
-    this.logger.debug(`http request result: ${JSON.stringify(result)}`);
+    this.logger.debug(`http request result: ${JSON.stringify(redact(result))}`);
```
WilliamBergamin, line 374: This is good 💯  but since this is in a critical path I think we might want to be safe and do something like the following, let me know what you think? 🤔 

```js
if (this.logger.getLevel() == LogLevel.DEBUG) {
    try{
         const redactedResults = JSON.stringify(redact(result))
         this.logger.debug(`http request result: ${redactedResults}`);
    } catch(error: any){
         this.logger.debug(`unable to log ressult`, error);
    }
}
```

WilliamBergamin (changes requested):
This is a solid change 💯 

I think we might want to gate the redaction logic behind a try catch so that it does not raise an error if it fails and rather logs the problem 🤔 let me know what you think of this

If we could add some tests for this it would be nice

srtaalej (commented):
thank you for reviewing @WilliamBergamin! i agree with these changes except that the existing debug lines on 702 and 721 don't include these guards. should we add them there too to stay consistent?

WilliamBergamin:
Ohh actually by bad, the object we pass to `redact` was already processed by `buildResult` it isn't the raw response 😅  
I don't think this guard is needed after all

What do you think about just adding a unit test to make sure the redaction works 🙏

srtaalej:
added unit test and left redact changes as is! can i get a quick approve @WilliamBergamin?

packages/web-api/src/WebClient.test.ts
```diff
@@ -118,6 +118,21 @@ describe('WebClient', () => {
       // Calling #setName of the given logger is destructive
       assert.strictEqual((logger.setName as sinon.SinonStub).called, false);
     });
+
+    it('should redact tokens in API response bodies in debug logs', async () => {
+      const scope = nock('https://slack.com')
+        .post('/api/oauth.v2.access')
+        .reply(200, { ok: true, access_token: 'xoxb-secret-token', token_type: 'bot' });
```
WilliamBergamin: What do you think about using a payload from [the docs](https://docs.slack.dev/reference/methods/oauth.v2.access/#response) for this test

```suggestion
        .reply(200, {
            "ok": true,
            "access_token": "xoxb-secret-token",
            "token_type": "bot",
            "scope": "commands,incoming-webhook",
            "bot_user_id": "U0KRQLJ9H",
            "app_id": "A0KRD7HC3",
            "expires_in": 43200,
            "refresh_token": "xoxe-secret-token",
            "team": {
                "name": "Slack Softball Team",
                "id": "T9TK3CUKW"
             },
             "enterprise": {
               "name": "slack-sports",
                "id": "E12345678"
              },
              "authed_user": {
                "id": "U1234",
                "scope": "chat:write",
                "access_token": "xoxe.xoxp-secret-token",
                "expires_in": 43200,
                "refresh_token": "xoxe-secret-token",
                "token_type": "user"
              }
            });
```

packages/web-api/src/WebClient.test.ts
```diff
@@ -118,6 +118,21 @@ describe('WebClient', () => {
...
+      const client = new WebClient(undefined, { logLevel: LogLevel.DEBUG, logger });
+      await client.apiCall('oauth.v2.access', { code: 'test-code' });
+      scope.done();
+
+      const debugCalls = (logger.debug as sinon.SinonStub).getCalls();
+      const resultLog = debugCalls.find((call) => call.args[0].includes('http request result'));
+      assert.ok(resultLog, 'expected a debug log for the http request result');
+      assert.ok(resultLog.args[0].includes('[[REDACTED]]'), 'expected access_token to be redacted');
+      assert.ok(!resultLog.args[0].includes('xoxb-secret-token'), 'expected raw token to not appear in logs');
```
WilliamBergamin: Similar with the asserts

```suggestion
      assert.ok(!resultLog.args[0].includes('xoxb-secret-token'), 'expected raw bot token to not appear in logs');
      assert.ok(!resultLog.args[0].includes('xoxe-secret-token'), 'expected raw refresh token to not appear in logs');
       assert.ok(!resultLog.args[0].includes('xoxe.xoxp-secret-token'), 'expected raw user token to not appear in logs');
      assert.ok(!resultLog.args[0].includes('xoxe-secret-token'), 'expected raw user token to not appear in logs');
```

WilliamBergamin (commented):
Looking good 💯  left one suggestion for the test 🙏

packages/web-api/src/WebClient.test.ts
srtaalej: sure 🚀

WilliamBergamin (approved):
Nice fix 💯

PR #2657, opened 2026-07-12
fix(socket-mode): observe reconnect rejections

### LINKED ISSUE #2656: SocketModeClient reconnect rejection is unhandled when disconnect races retry
### Description

`@slack/socket-mode@2.0.7` can terminate the Node process with `ERR_UNHANDLED_REJECTION` when a Socket Mode reconnect is stopped while it is still pending.

The rejection reason is `undefined`, matching the argument emitted by `State.Disconnected` during shutdown.

### Reproducible in

- `@slack/socket-mode`: 2.0.7 (latest)
- `@slack/bolt`: 4.7.2
- Node: 25.8.2 (also reproducible on 26.5.0)
- macOS 15 / arm64

### Relevant code path

In `SocketModeClient`:

1. The `close` listener calls `this.delayReconnectAttempt(this.start)` without observing the returned Promise.
2. `delayReconnectAttempt()` calls `cb.apply(this).then(res)` with no rejection handler.
3. If `disconnect()` races that reconnect, the reconnecting `start()` rejects from `State.Disconnected` with `undefined`.
4. The Promise returned by `.then(res)` is unhandled, so modern Node terminates the process.

### Minimal reproduction

This uses the public EventEmitter surface and replaces only the network attempt so the rejection path is deterministic:

```js
const { SocketModeClient } = require('@slack/socket-mode');

const client = new SocketModeClient({
  appToken: 'xapp-test',
  autoReconnectEnabled: true,
  clientPingTimeout: 1,
});

client.start = () => Promise.reject(undefined);
process.once('unhandledRejection', (reason) => {
  console.log({ reproduced: true, reasonIsUndefined: reason === undefined });
  process.exit(0);
});

client.emit('close');
setTimeout(() => process.exit(2), 100);
```

Output:

```text
{ reproduced: true, reasonIsUndefined: true }
```

### Production trigger observed

During a service restart, several Socket Mode connections opened but did not receive pong/hello before the client timeout. The application stopped the pending receiver after its startup deadline. The reconnect rejection then terminated the process; a supervisor restarted it and repeated the cycle until Slack connectivity recovered.

The characteristic log was:

```text
WebSocket was closed before the connection was established
UnhandledPromiseRejection: ... reason "undefined"
code: 'ERR_UNHANDLED_REJECTION'
```

### Expected behavior

Every reconnect attempt should have an observed rejection. Stopping during reconnect should settle/cancel the delayed reconnect without an unhandled Promise.

### Suggested fix

Two changes appear sufficient:

```js
void this.delayReconnectAttempt(this.start).catch((error) => {
  if (!this.shuttingDown) this.logger.error(`Failed to reconnect Socket Mode client: ${error}`);
});
```

and propagate callback rejection from the delayed Promise:

```js
  setTimeout(() => {
    if (this.shuttingDown) return resolve();
    this.emit(State.Reconnecting);
    cb.apply(this).then(resolve, reject);
  }, msBeforeRetry);
});
```

I searched existing issues and found related reconnect reports (#1243, #2094), but not this current 2.0.7 dropped-rejection path.

w3lld1:
### Summary

- propagate delayed reconnect callback rejections instead of leaving the returned promise pending
- observe failures from close-triggered reconnect attempts and log them when the client is still active
- settle a delayed reconnect cleanly when shutdown wins the race
- add a regression test for a rejected reconnect attempt

Fixes #2656

### Validation

- `npm run lint`
- `npm run build`
- `npm run test:unit --workspace=@slack/socket-mode` (21 tests passed)

### Requirements

### DIFF
--- a/packages/socket-mode/src/SocketModeClient.test.ts
+++ b/packages/socket-mode/src/SocketModeClient.test.ts
@@ -83,6 +83,24 @@ describe('SocketModeClient', () => {
     it('should resolve once Disconnected state emitted');
   });
 
+  describe('reconnect', () => {
+    it('should observe a rejected reconnect attempt', async () => {
+      const logger = new ConsoleLogger();
+      const errorLog = sandbox.stub(logger, 'error');
+      const client = new SocketModeClient({
+        appToken: 'xapp-',
+        clientPingTimeout: 1,
+        logger,
+      });
+      sandbox.stub(client, 'start').rejects(new Error('reconnect failed'));
+
+      client.emit('close');
+      await sleep(10);
+
+      sinon.assert.calledWith(errorLog, 'Failed to reconnect Socket Mode client: Error: reconnect failed');
+    });
+  });
+
   describe('onWebSocketMessage', () => {
     // While this method is protected and cannot be invoked directly, emitting the 'message' event directly invokes it
     describe('slash_commands messages', () => {
--- a/packages/socket-mode/src/SocketModeClient.test.ts
+++ b/packages/socket-mode/src/SocketModeClient.test.ts
@@ -99,6 +99,24 @@ describe('SocketModeClient', () => {
 
       sinon.assert.calledWith(errorLog, 'Failed to reconnect Socket Mode client: Error: reconnect failed');
     });
+
+    it('should not reconnect or log an error while shutting down', async () => {
+      const logger = new ConsoleLogger();
+      const errorLog = sandbox.stub(logger, 'error');
+      const client = new SocketModeClient({
+        appToken: 'xapp-',
+        clientPingTimeout: 1,
+        logger,
+      });
+      const start = sandbox.stub(client, 'start').resolves({});
+
+      client.emit('close');
+      await client.disconnect();
+      await sleep(10);
+
+      sinon.assert.notCalled(start);
+      sinon.assert.notCalled(errorLog);
+    });
   });
 
   describe('onWebSocketMessage', () => {
--- a/packages/socket-mode/src/SocketModeClient.ts
+++ b/packages/socket-mode/src/SocketModeClient.ts
@@ -18,6 +18,8 @@ import { SlackWebSocket, WS_READY_STATES } from './SlackWebSocket';
 import type { SocketModeDispatcher, SocketModeOptions } from './SocketModeOptions';
 import { UnrecoverableSocketModeStartError } from './UnrecoverableSocketModeStartError';
 
+class SocketModeClientShuttingDownError extends Error {}
+
 // Lifecycle events as described in the README
 enum State {
   Connecting = 'connecting',
@@ -159,7 +161,7 @@ export class SocketModeClient extends EventEmitter {
       // Underlying WebSocket connection was closed, possibly reconnect.
       if (!this.shuttingDown && this.autoReconnectEnabled) {
         void this.delayReconnectAttempt(this.start).catch((error) => {
-          if (!this.shuttingDown) {
+          if (!(error instanceof SocketModeClientShuttingDownError)) {
             this.logger.error(`Failed to reconnect Socket Mode client: ${error}`);
           }
         });
@@ -255,7 +257,7 @@ export class SocketModeClient extends EventEmitter {
         this.reconnectionTimer = undefined;
         if (this.shuttingDown) {
           this.logger.debug('Client shutting down, will not attempt reconnect.');
-          resolve(undefined as T);
+          reject(new SocketModeClientShuttingDownError('Socket Mode client is shutting down'));
         } else {
           this.logger.debug('Continuing with reconnect...');
           this.emit(State.Reconnecting);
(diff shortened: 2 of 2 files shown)


mwbrooks:
Hi @w3lld1 👋🏻 Thanks for the pull request! 🙇🏻 

❓ Can you please sign our CLA - once signed, we can begin reviewing your pull request and enabling the workflow checks.

packages/socket-mode/src/SocketModeClient.ts
```diff
@@ -142,7 +142,11 @@ export class SocketModeClient extends EventEmitter {
     this.on('close', () => {
       // Underlying WebSocket connection was closed, possibly reconnect.
       if (!this.shuttingDown && this.autoReconnectEnabled) {
-        this.delayReconnectAttempt(this.start);
+        void this.delayReconnectAttempt(this.start).catch((error) => {
+          if (!this.shuttingDown) {
+            this.logger.error(`Failed to reconnect Socket Mode client: ${error}`);
```
mwbrooks, line 165: praise: Nice - the `if (!this.shuttingDown)` guard here is great. It surfaces the real reconnect failure while keeping the log quiet during a normal `disconnect()` race, which is the whole point of the fix. Thanks for the care.

packages/socket-mode/src/SocketModeClient.ts
```diff
@@ -228,14 +232,15 @@ export class SocketModeClient extends EventEmitter {
     this.numOfConsecutiveReconnectionFailures += 1;
     const msBeforeRetry = this.clientPingTimeoutMS * this.numOfConsecutiveReconnectionFailures;
     this.logger.debug(`Before trying to reconnect, this client will wait for ${msBeforeRetry} milliseconds`);
-    return new Promise((res, _rej) => {
+    return new Promise((resolve, reject) => {
       setTimeout(() => {
         if (this.shuttingDown) {
           this.logger.debug('Client shutting down, will not attempt reconnect.');
```
mwbrooks, line 259: suggestion: We should handle this differently. Resolving with `undefined as T` works for the `close`-handler, but `retrieveWSSURL` also calls `delayReconnectAttempt` (line 276) and uses the result as a WebSocket URL.

If `disconnect()` races a retry inside `retrieveWSSURL`, this now returns `undefined` up through the recursion and `start()` ends up constructing `new SlackWebSocket({ url: undefined, ... })`. Not ideal.

A couple of options come to mind:

1. Short-circuit earlier so we never call into `start()` / return a URL when we know we're shutting down. This may match the approach throughout the package.
2. Reject with a specific "shutdown in progress" error and have callers (both the `close` handler and `retrieveWSSURL`) swallow it explicitly. That keeps the type honest and avoids the `undefined`-as-URL path.

packages/socket-mode/src/SocketModeClient.test.ts
```diff
@@ -39,6 +39,24 @@ describe('SocketModeClient', () => {
     it('should resolve once Disconnected state emitted');
   });
 
+  describe('reconnect', () => {
+    it('should observe a rejected reconnect attempt', async () => {
```
mwbrooks, line 87: praise: Really appreciate having a regression test alongside the fix - it locks in the main failure mode from #2656.

packages/socket-mode/src/SocketModeClient.test.ts
```diff
@@ -39,6 +39,24 @@ describe('SocketModeClient', () => {
...
+        appToken: 'xapp-',
+        clientPingTimeout: 1,
+        logger,
+      });
+      sandbox.stub(client, 'start').rejects(new Error('reconnect failed'));
+
+      client.emit('close');
+      await sleep(10);
+
+      sinon.assert.calledWith(errorLog, 'Failed to reconnect Socket Mode client: Error: reconnect failed');
```
mwbrooks, line 100: suggestion: We should  add a second test case here that covers the `!this.shuttingDown` guard.

mwbrooks (changes requested):
Thanks for the clean, focused fix - it's focused on https://github.com/slackapi/node-slack-sdk/issues/2656 and the code changes track it closely. I've left comments on two things that we'll need addressed along with the signed CLA. Thanks!

w3lld1:
Addressed both review points in `5c8a15dc`: the delayed reconnect now rejects with a typed shutdown signal instead of resolving `undefined`, the close handler suppresses only that expected signal, and a regression test covers the disconnect-during-delay guard.

Validation: the socket-mode unit suite passes (22 tests), the full repository Biome check passes, and `git diff --check` passes. The Salesforce CLA check is now green.

w3lld1:
Thanks for the review. The follow-up commit `5c8a15dc` addresses both requested changes:

- `delayReconnectAttempt()` now rejects with a dedicated `SocketModeClientShuttingDownError` instead of returning `undefined`, and the close handler suppresses only that expected shutdown error;
- the second regression test covers the shutdown race and verifies that `start()` is not called and no reconnect error is logged.

Fresh validation:
- `npm run build`
- `npm run lint`
- `npm run test:unit --workspace=@slack/socket-mode` — 22 passed
- `git diff --check`

w3lld1:
> Thanks for the contribution! Before we can merge this, we need @w3lld1 to [sign the Salesforce Inc. Contributor License Agreement](https://cla.salesforce.com/sign-cla?prUrl=https%3A%2F%2Fgithub.com%2Fslackapi%2Fnode-slack-sdk%2Fpull%2F2657).

Signed

PR #2662, opened 2026-07-20
chore(webhook): bump p-retry to v7 and fix Node version source of truth

WilliamBergamin:
### Summary

Dependabot opened #2655 to bump `p-retry` to v8, but v8 requires **Node >= 22** — higher than `@slack/webhook`'s supported floor of **`>= 20`** — so its CI failed. This PR resolves that instead of forcing a Node version bump on the SDK:

- **Bumps `p-retry` `^4` → `^7`.** v7 is the newest release whose Node requirement (`>= 20`) still matches our engine floor. (Regenerating the lockfile also fixes the `npm ci` failure from #2655.)
- **Ignores `p-retry` in Dependabot for `/packages/webhook`.** It's pure ESM and raises its minimum Node with every major, so automated bumps keep breaking CI — maintainers should upgrade it deliberately.
- **Fixes stale Node versions in `AGENTS.md`.** It claimed "Node 18+" and an outdated CI matrix; it now points to `package.json` `engines.node` and `ci-build.yml` as the sources of truth so the docs can't drift again.

Verified locally: webhook builds with no type errors and all 35 tests pass. Supersedes #2655.

### Requirements 

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -223,6 +223,11 @@ updates:
           - "@types/mocha"
           - "mocha"
     ignore:
+      # p-retry is pure ESM and raises its minimum Node.js version each major
+      # (v7 requires Node >= 20, v8 requires Node >= 22). Bumping it can force a
+      # change to this package's own supported Node floor, so maintainers must
+      # upgrade it deliberately rather than via automated PRs.
+      - dependency-name: "p-retry"
       - dependency-name: "@types/node"
         versions:
           - "19.x"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -225,8 +225,7 @@ updates:
     ignore:
       # p-retry is pure ESM and raises its minimum Node.js version each major
       # (v7 requires Node >= 20, v8 requires Node >= 22). Bumping it can force a
-      # change to this package's own supported Node floor, so maintainers must
-      # upgrade it deliberately rather than via automated PRs.
+      # change to this package's own supported Node floor.
       - dependency-name: "p-retry"
       - dependency-name: "@types/node"
         versions:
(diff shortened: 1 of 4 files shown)


AGENTS.md
```diff
@@ -78,14 +78,14 @@ npm run build --workspace=@slack/socket-mode
 2. **Request types ARE manually maintained** — `packages/web-api/src/types/request/` is hand-written code; edit these responsibly.
 3. **Build packages in dependency order** — see the dependency graph above.
 4. **Use Biome**, not ESLint or Prettier — config is in `biome.json` at repo root.
-5. **TypeScript 5.9.3**, Node 18+ (tested on Node 18, 20, 22, 24).
+5. **TypeScript 5.9.3**. The minimum supported Node.js version is the `engines.node` field in each package's `package.json`.
```
zimeg, line 81: 🔭 note: The `tsc` version might soon change without @dependabot updating this file. FWIW I like the referenced approach used for Node!

🔗 https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/

.github/dependabot.yml
```diff
@@ -223,6 +223,10 @@ updates:
           - "@types/mocha"
           - "mocha"
     ignore:
+      # p-retry is pure ESM and raises its minimum Node.js version each major
+      # (v7 requires Node >= 20, v8 requires Node >= 22). Bumping it can force a
+      # change to this package's own supported Node floor.
+      - dependency-name: "p-retry"
```
zimeg, line 229: 👾 question: This doesn't seem to have critical updates at the moment but next breaking minimum version change do we plan to un-ignore this?

☁️ ramble: I'm hoping context overall is enough with "@types/node" right below but want to note this as upcoming in some future release still!

zimeg (approved):
@WilliamBergamin Quick fixes and forward fixes are much appreciated! Thanks!

.github/dependabot.yml
WilliamBergamin: This dependency [requires node 22](https://github.com/sindresorhus/p-retry/blob/35681f6c70f8ca2bdcb9542281147679184269fa/package.json#L20), I'm not sure why dependabot tried to bump this to version 8 🤔

PR #2694, opened 2026-08-10
feat(socket-mode): support undici v8 as a peer dependency

WilliamBergamin:
### Summary

Expand `@slack/socket-mode` to support both `undici@^7` and `undici@^8` as a peer dependency.

- Widen `peerDependencies.undici` to `^7.0.0 || ^8.0.0`. The source needs no logic changes
- CI now pairs each Node version with an undici major via a `versions: [{ node, undici }]` matrix and forces the socket-mode workspace onto that major before running the suite.

### Requirements 


WilliamBergamin:
@zimeg I'll need to update the status checks if we want to merge this

packages/socket-mode/src/undiciVersion.test.ts
```diff
@@ -0,0 +1,33 @@
...
+const undiciVersion: string = require('undici/package.json').version;
+const undiciMajor = Number.parseInt(undiciVersion.split('.')[0], 10);
+const nodeMajor = Number.parseInt(process.versions.node.split('.')[0], 10);
+
+describe('undici peer dependency', () => {
+  it('never runs undici v8+ on Node older than 22', () => {
+    assert.ok(
+      undiciMajor < 8 || nodeMajor >= 22,
+      `undici@${undiciVersion} requires Node >=22, but tests are running on Node ${process.versions.node}`,
+    );
```
zimeg: ☕ note: Forgive me if logic is confusing but I find the short-circuit current a bit confusing when I think about running this test with `node24`. I might suggest instead:


```suggestion
      undiciMajor >= 8 && nodeMajor < 22,
      `undici@${undiciVersion} requires Node >=22, but tests are running on Node ${process.versions.node}`,
    );
```

packages/socket-mode/package.json
```diff
@@ -57,7 +57,7 @@
     "eventemitter3": "^5"
   },
   "peerDependencies": {
-    "undici": "^7.0.0"
+    "undici": "^7.0.0 || ^8.0.0"
```
zimeg, line 60: 🎯 praise: Thanks for keeping support consistent!

zimeg (approved):
@WilliamBergamin LGTM! I'm bummed CI wasn't calling this out with regular installations but this is a good fix to include.

🧪 Longterm I'm worried about testing compatible dependencies with unit tests but let's get this merged to address an issue!

PR #2696, opened 2026-08-12
feat(web-api): add webapi response changes as of 2026-08-21

zimeg:
### Summary

This pull request regenerates the Web API response types from the current `java-slack-sdk` samples, picking up newly-returned optional properties.

- Regenerates `packages/web-api/src/types/response/` via `scripts/generate-web-api-types.sh`.
- Adds newly-returned optional fields (e.g. a top-level `warning` on many responses, `Channel.properties` on `conversations.info`) and a new response type for `admin.users.getExpiration`. All additions are optional, so existing consumer code continues to type-check.
- **Fixes the generator script:** the final step ran `npm run lint:fix` from inside `packages/web-api`, but that package has no `lint`/`lint:fix` script since the repo moved to root-level Biome — so the script errored on every run (`Missing script: "lint:fix"`) and left generated files un-normalized. Now runs `npm run lint:fix` from the repo root. With that fixed, the regen diff drops from ~330 files (mostly formatting churn) to ~180 files of real property changes.

**Known artifact (not blocking):** quicktype infers a bogus `Tabz`/`tabz` shape on `conversations.info` from a typo'd key in the source sample. Left in for now — the real fix belongs upstream in the `java-slack-sdk` `json-logs/samples`, not a hand-edit here. Flagging for a follow-up.

Draft: opening for review of the generated diff before marking ready.

### Testing

- From the repo root, run `./scripts/generate-web-api-types.sh` and confirm it completes end-to-end (including the `lint:fix` step) without a `Missing script` error.
- Spot-check a regenerated type (e.g. `ConversationsInfoResponse.ts`) against the live `conversations.info` response to confirm the added optional fields match what the API returns.


packages/web-api/src/types/response/AdminAppsActivitiesListResponse.ts
```diff
@@ -15,6 +15,7 @@ export type AdminAppsActivitiesListResponse = WebAPICallResult & {
   ok?: boolean;
   provided?: string;
   response_metadata?: ResponseMetadata;
+  warning?: string;
```
zimeg, line 18: 👾 note: According to agentic search `229/330` methods have "warning" added. We might continue to search for it in upstream changes but this remains true to scripting at this time.

scripts/generate-web-api-types.sh
```diff
@@ -21,6 +21,4 @@ npm i
 ruby ./code_generator.rb
 popd
 # run lint fixing after type generation
-pushd packages/web-api
-npm i
```
zimeg, line 25: 🪓 note: This is removed with a **monorepo** setup towards linting!

packages/web-api/src/types/response/WorkflowsStepCompletedResponse.ts
zimeg, line 1: 📝 note: Responses to `workflows.step.*` are included to mirror logs upstream but these methods are not active at this time.

packages/web-api/src/types/response/AdminUsersGetExpirationResponse.ts
zimeg, line 1: 🗣️ note: We should include this change with https://github.com/slackapi/node-slack-sdk/pull/2451!

packages/web-api/src/types/response/ChatAppendStreamResponse.ts
```diff
@@ -1,3 +1,12 @@
+/////////////////////////////////////////////////////////////////////////////////////////
+//                                                                                     //
+// !!! DO NOT EDIT THIS FILE !!!                                                       //
+//                                                                                     //
+// This file is auto-generated by scripts/generate-web-api-types.sh in the repository. //
+// Please refer to the script code to learn how to update the source data.             //
+//                                                                                     //
+/////////////////////////////////////////////////////////////////////////////////////////
```
zimeg, line 8: 🤖 note: Some responses were handwritten and are now overwritten with automated response. This is a good change for these packages.

📝 note: I'll leave comments on WIP generations that might have changed or don't include expected response variables!

packages/web-api/src/types/response/SlackListsDownloadGetResponse.ts
```diff
@@ -1,10 +1,24 @@
...
+import type { WebAPICallResult } from '../../WebClient';
 export type SlackListsDownloadGetResponse = WebAPICallResult & {
+  download_url?: string;
   error?: string;
   needed?: string;
   ok?: boolean;
   provided?: string;
+  response_metadata?: ResponseMetadata;
   status?: string;
-  download_url?: string;
```
zimeg, line 9: 💾 note: This is moved into alphabetical order.

packages/web-api/src/types/response/SlackListsDownloadStartResponse.ts
```diff
@@ -1,9 +1,23 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////
 
+import type { WebAPICallResult } from '../../WebClient';
 export type SlackListsDownloadStartResponse = WebAPICallResult & {
   error?: string;
+  job_id?: string;
   needed?: string;
   ok?: boolean;
   provided?: string;
-  job_id?: string;
```
zimeg, line 8: 🧮 note: This is moved to alphabetical orderings.

packages/web-api/src/types/response/WorkflowsFeaturedListResponse.ts
```diff
@@ -1,11 +1,18 @@
...
 export type WorkflowsFeaturedListResponse = WebAPICallResult & {
-  featured_workflows: {
-    channel_id: string;
-    triggers: {
-      id: string;
-      title: string;
-    }[];
-  }[];
+  error?: string;
+  invalid_channel_ids?: string[];
```
zimeg, line 14: 👁️‍🗨️ note: This removed the `featured_workflows` attribute in earlier responses which makes me think the upstream generation tests need to be updated!

packages/web-api/src/types/response/ChatStopStreamResponse.ts
```diff
@@ -1,24 +1,16 @@
...
+
 import type { WebAPICallResult } from '../../WebClient';
 export type ChatStopStreamResponse = WebAPICallResult & {
-  channel?: string;
   error?: string;
   needed?: string;
   ok?: boolean;
   provided?: string;
-  ts?: string;
-  message?: ChatStopStreamResponseMessage;
```
zimeg, line 10: 🪬 note: Particular details on this response aren't found in responses and might require test changes as well.

packages/web-api/src/types/response/SlackListsCreateResponse.ts
```diff
@@ -1,14 +1,69 @@
-import type { WebAPICallResult } from '../../WebClient';
-import type { SlackListsSchemaColumnResponse } from '../request/slackLists';
```
zimeg, line 2: 🪓 note: We avoid importing types from `requests` for this method as part of the autogenerated change. I find this continues to support existing and now additional attributes but has some risk to using exports not from the top-level of the package in code.

👁️‍🗨️ note: I'll add a few more commits to remove additional request objects.

zimeg (commented):
🌲 A few changes are in progress for this PR but I leave the following comments after a first pass and will focus on changes to particular methods:

- `admin.users.getExpiration`
- `chat.stopStream`
- `workflows.featured.list`

The comment on `slackLists` is worth calling out again here and I plan to adjust the **changesets** to call out changed exports if it ships in a current state but I understand if this deserves more discussion 🚢

zimeg:
Refreshed the regen against the latest samples.

Since this was first opened (Aug 12 samples), a new upstream dump landed — [slackapi/java-slack-sdk#1629](https://github.com/slackapi/java-slack-sdk/pull/1629) *("webapi response changes as of 2026-08-19")*. Merged current `main` in and re-ran `./scripts/generate-web-api-types.sh` against it.

**New in this refresh (24 response types, all additive/optional):** new fields across `chat.*`, `conversations.history`/`replies`/`open`, `files.*`, `pins.list`, `reactions.*`, `search.all`/`files`/`messages`, `users.*`, plus a populated `workflows.featured.list` body (`featured_workflows[]` → `FeaturedWorkflow` → `Trigger`).

- No new required fields — every addition is optional, so existing consumers still type-check.
- The ~140 removed lines are quicktype reordering (alphabetization), not property removals — verified zero properties dropped.
- `@slack/web-api` builds (`tsc`) and lints (Biome) clean.

The `Tabz`/`tabz` shape on `conversations.info` is a **known, accepted** quicktype detection — leaving it as-is (no longer tracking as a follow-up).

Still draft pending review of the generated diff.

zimeg:
### Note: regenerating off the latest `java-slack-sdk` `api` branch recovers nearly all dropped properties — tracking the remainder

I re-ran the generator against the current `slackapi/java-slack-sdk` `api` branch (tip `dcfad72`, which includes the `2026-08-20` sample refresh) to confirm which of the properties this PR's regen dropped are now covered by upstream samples.

**Recovered** (a fresh regen off the current `api` restores these — they were only missing because the samples didn't previously exercise them):

- `SlackListsCreateResponse` / `SlackListsItemsInfoResponse` — `options.precision`, `options.date_format`, `options.emoji`, `options.max` (list column option variants; the samples now include number/date/rating columns).
- `SearchFilesResponse` / `SearchAllResponse` — `shares.public[].date_last_shared`, `shares.public[].latest_reply` (present in the upstream search sample fixtures).
- `FilesInfoResponse` — the reply share detail (`latest_reply`, `reply_count`, `reply_users`, `reply_users_count`).

**Still missing so far** (not yet in any `api` sample, so a regen would still omit these):

- `UsersListResponse` — top-level `arg` and `Member.is_connector_bot`. Neither is reliably reproducible in the remote-API test (`arg` is a request echo; `is_connector_bot` requires a connector bot in the test workspace), so these remain dropped for now.
- `SlackListsCreateResponse` `options` — `emoji_team_id`, `notify_users`, `default_value_typed`. These are covered by a pending `slacklists_Test` change on `api` but won't land in the sample until the next detection run executes it.

Net: after a regen against the latest `api`, the only residual drops versus `main` are the two `UsersListResponse` fields above (plus the three `options` fields once the next detection run captures them). Everything else this PR appeared to drop is restored.

zimeg:
**Re: removed properties vs base — verified, not a real property loss.** _(Updated to describe `arg` more accurately.)_

A property-level diff of the full PR against `main` (all 330 response types) is additive-only (**+1,715 properties, 0 removed**) *except* two fields dropped from `UsersListResponse`. Neither is a loss of a real success-response property:

- **`Member.is_connector_bot`** — not returned on the public user object by `users.list` (or `users.info`); it's surfaced on **`bots.info`**, where this PR correctly *adds* it (to `BotsInfoResponse`). Its presence on the `users.list` member type was inaccurate, so removing it is a correction.

- **`arg`** — this is an **error-response echo** field (the argument name reported on a failed call, alongside `error`/`needed`/`provided`), not a real field on a *successful* `users.list` body. The generator infers each type from recorded samples, so `arg` only lands on whichever methods had an error sample captured — currently just 3 of 330 types, vs `error`/`needed`/`provided` which appear on all 330. So its presence per-method is inconsistent sample-inference artifact, not a schema fact; this regen just shuffles which few types carry it. Dropping it from `users.list` brings that method in line with the other ~327 that don't carry it. (A cleaner long-term fix would hoist `arg` onto the shared `WebAPICallResult` base so it's consistent everywhere — out of scope here.)

Net: no legitimate existing property is lost. Will note both in the release notes.

.changeset/regen-web-api-response-types.md
```diff
@@ -0,0 +1,12 @@
+---
+"@slack/web-api": minor
+---
+
+feat(web-api): add webapi response changes as of 2026-08-21
+
+Regenerate `packages/web-api/src/types/response` from the latest `java-slack-sdk` samples, picking up new stable response properties (e.g. `agent_session` on message-bearing responses, `ListView` grouping/sort/filter fields, `bots.info` connector/workflow flags).
```
zimeg:
```suggestion
Regenerate response types with new stable response properties (e.g. `agent_session` on message-bearing responses, `ListView` grouping/sort/filter fields, `bots.info` connector/workflow flags).
```

📚 suggestion: To keep this focused on this package consumers!

zimeg (commented):
📣 A comment for release notes

PR #2710, opened 2026-08-26
fix(socket-mode): tear down leaked sockets on disconnect (#2709)

### LINKED ISSUE #2709: @slack/socket-mode 3.0.0: undici port dropped all force-close paths — sockets leak until the 10-connection cap blocks reconnection
### Summary

The 3.0.0 migration of `@slack/socket-mode` from `ws` to undici's WebSocket removed every mechanism that guaranteed a socket actually dies, while the surrounding code still assumes one exists. On a long-running process with occasional transport errors, connections accumulate to Slack's 10-connections-per-app limit, after which reconnection fails until Slack garbage-collects a stale connection server-side.

We observed this in production as a ~2 hour outage: reconnects refused every 10 minutes for 15 consecutive attempts, recovering at the exact moment the server-side connection count dropped from 10 to 9. Client-side we measured 8 leaked ESTABLISHED kernel sockets to `wss-primary.slack.com` from a single process.

### The defects

All in `dist/src/SlackWebSocket.js` unless noted.

**1. `disconnect()` on an errored socket waits forever.** The error path sends a close frame (`this.websocket.close(1000)`) and waits for the peer's reply. The comment in the CLOSING branch says:

> // A close frame was already sent but the peer hasn't responded. Force-terminate rather than
> // waiting for the ws library's closeTimeout (~30s) while the ping monitor logs repeated warnings.

But undici has **no close timeout at all** — there is no timer anywhere in undici's websocket implementation (checked against undici 7.28.0). The comment documents a safety net that stopped existing with the transport port. A dead peer never answers, `closeHandler` never fires, and `cleanup()` is never reached.

**2. `cleanup()` never closes the underlying TCP connection.** It removes listeners, unsubscribes the diagnostics channels, nulls `this.websocket`, and clears timers — but nothing closes the socket. undici's WebSocket has no `terminate()`; the `ws` version's `terminate()` was the actual teardown. The kernel socket remains ESTABLISHED indefinitely after the object is abandoned (undici keepalives keep it alive against a healthy peer).

**3. `SocketModeClient`'s `'close'` handler reconnects without deduplication** (`dist/src/SocketModeClient.js`). Every `SlackWebSocket` instance emits `'close'` on the shared client emitter, the handler calls `delayReconnectAttempt()` unconditionally, and `delayReconnectAttempt()` overwrites `this.reconnectionTimer` without clearing the previous one. A stale instance's late `'close'` while the current connection is healthy spawns a duplicate live connection — a second path to the same 10-connection cap.

### Reproduction

Harness for defects 1+2 — a server that accepts the websocket and then detaches its data listeners, so the client's close frame is never processed (simulating a dead/unresponsive peer):

```js
const { WebSocketServer } = require('ws');
const { SlackWebSocket } = require('@slack/socket-mode/dist/src/SlackWebSocket');
const { EventEmitter } = require('node:events');

let serverTcpClosed = false;
const wss = new WebSocketServer({ host: '127.0.0.1', port: 0 });
wss.on('connection', (ws) => {
  const raw = ws._socket;
  raw.removeAllListeners('data');       // no frame, incl. close, is ever processed
  raw.on('data', () => {});             // keep the stream flowing, discard everything
  raw.on('close', () => { serverTcpClosed = true; });
  raw.on('error', () => {});
});
wss.on('listening', () => {
  const client = new EventEmitter();
  let clientClosed = false;
  client.on('close', () => { clientClosed = true; });
  client.on('error', () => {});
  const sws = new SlackWebSocket({
    url: `ws://127.0.0.1:${wss.address().port}/`, client, logLevel: 'error',
    clientPingTimeoutMS: 60000, serverPingTimeoutMS: 60000, pingInterval: 60000,
  });
  sws.connect();
  setTimeout(() => {
    sws.disconnect();   // sends close frame; peer never answers
    const t0 = Date.now();
    setInterval(() => {
      if (clientClosed && serverTcpClosed) {
        console.log(`cleanup fired and TCP closed after ${Date.now() - t0}ms`); process.exit(0);
      } else if (Date.now() - t0 > 12000) {
        console.log(`LEAK: closeEmitted=${clientClosed} tcpClosed=${serverTcpClosed}`); process.exit(1);
      }
    }, 250);
  }, 1500);
});
```

Unpatched 3.0.0: `LEAK: closeEmitted=false tcpClosed=false` — no `'close'` event ever fires (so the client never knows to reconnect) and the TCP connection stays ESTABLISHED forever.

### Suggested fixes

- `disconnect()`: add a close-handshake timeout that forces cleanup, restoring the invariant the `ws` version provided via `closeTimeout`/`terminate()`.
- `cleanup()`: actually tear down the connection. Since undici's WebSocket exposes no terminate and `agent.destroy()` does not reach a socket once the upgrade detaches it from the pool (verified), one working approach is holding the raw socket: create the WebSocket with a dedicated `Agent` whose `connect` wraps `buildConnector` and captures the `net.Socket`; `cleanup()` destroys it. We run this as a local patch and confirmed cleanup + real TCP teardown in ~5s against the harness above.
- `SocketModeClient` `'close'` handler: ignore the event when the current websocket `isActive()`, and don't schedule a second reconnect while `reconnectionTimer` is pending.

### Environment

- `@slack/socket-mode` 3.0.0 (via `@slack/bolt` 5.0.0)
- undici 7.28.0, Node.js 24.18.0, Linux (arm64)
- Long-running Socket Mode process (days of uptime), single app/workspace

### Requirements

- [x] I've read and understood the [Contributing guidelines](https://github.com/slackapi/node-slack-sdk/blob/main/.github/contributing.md) and have done my best effort to follow them.
- [x] I've searched for any related issues and avoided creating a duplicate issue.

WilliamBergamin:
### Summary

Fixes #2709.

Porting `@slack/socket-mode` from the `ws` library to undici's `WebSocket` in 3.0.0 silently dropped some **force-close** mechanism `ws` provided. 😅  When a peer stops responding, the underlying TCP socket is never torn down and accumulates in `ESTABLISHED` state, pushing apps toward Slack's ~10-connections-per-app cap until Socket Mode stops connecting.

1. **`disconnect()` waited forever for the close handshake.**
    1. `SlackWebSocket.close(1000)` writes a CLOSE frame with no timer, so a dead peer that never replies leaves the socket stuck in `CLOSING` and `cleanup()` never runs.
    2. 🟢  A `CLOSE_HANDSHAKE_TIMEOUT_MS` (30s) timeout now arms when the close frame is sent and forces `cleanup()` if the handshake stalls.

2. **`cleanup()` never destroyed the underlying TCP socket.**
    1. undici's `WebSocket` exposes no public `.socket`, `terminate()`, or abort signal. The raw socket lives on a private handler.
    2. 🟢 `SlackWebSocket` now creates its own dispatcher that captures the raw socket, and destroys it during `cleanup()`. A user-supplied dispatcher owns its own socket and cannot be force-closed here; it falls back to the close-handshake timeout above. This limitation is documented in the `dispatcher` option's JSDoc.

3. **Reconnect storm on `close`.** `SocketModeClient`'s `'close'`
    1.  handler scheduled a reconnect on every `close`, with no guard against stale closes fired while a connection is still active, or against multiple closes stacking multiple reconnects.
    2. 🟢 The handler now returns early if the socket is still active or a reconnect is already scheduled.

**Known limitation:** socket-level force-close only applies when the client creates its own `Agent` (no user dispatcher). This is intentional and documented.

### Requirements 


packages/socket-mode/src/SlackWebSocket.ts
```diff
@@ -1,14 +1,18 @@
...
-import { CloseEvent, type Dispatcher, ErrorEvent, MessageEvent, ping, WebSocket } from 'undici';
+import { Agent, buildConnector, CloseEvent, type Dispatcher, ErrorEvent, MessageEvent, ping, WebSocket } from 'undici';
 
 import { SMWebsocketError } from './errors';
 import log, { type Logger, LogLevel } from './logger';
 import type { SocketModeDispatcher } from './SocketModeOptions';
 
 export const WS_READY_STATES = ['CONNECTING', 'OPEN', 'CLOSING', 'CLOSED'];
 
+const CLOSE_HANDSHAKE_TIMEOUT_MS = 30_000;
```
zimeg, line 14: 👁️‍🗨️ thought: This might be nice to surface as an option although I understand it's for default `dispatcher` so might be confusing in some cases...

packages/socket-mode/src/SlackWebSocket.ts
```diff
@@ -250,9 +291,14 @@ export class SlackWebSocket {
...
     this.pingHandler = null;
     this.pongHandler = null;
+    if (this.defaultSocket && !this.defaultSocket.destroyed) {
+      this.defaultSocket.destroy();
+    }
+    this.defaultSocket = null;
     this.websocket = null;
     clearTimeout(this.serverPingTimeout);
     clearInterval(this.clientPingTimeout);
+    clearTimeout(this.closeHandshakeTimeout);
```
zimeg, line 301: 🧠 praise: Clever pattern to avoid confused logs!

zimeg (approved):
@WilliamBergamin LGTM! Impressive patterns to this change I think. I'm leaving a comment about added options but nothing to stop this from next release 🚢 💨

packages/socket-mode/src/SlackWebSocket.ts
WilliamBergamin: Yeah I thought about this also 🤔 But I think we can add this in later if we need to

PR #2712, opened 2026-08-27
fix: revert internal packages requiring esm loading for backward compatible support

### LINKED ISSUE #2711: @slack/web-api 8.1.0: CJS build requires ESM-only p-queue@9 and p-retry@8, ERR_REQUIRE_ESM on Node 20
### Package

@slack/web-api

### SDK Version

@slack/web-api@8.1.0

### Node.js Version

v20.20.2

### Operating System

macOS 26.6

### Steps to Reproduce

`mkdir slack-esm-repro && cd slack-esm-repro`
`npm init -y`
`npm install @slack/web-api@8.1.0`
Note: Node 20.19+ and 22.12+ mask this via require(esm). Disable it to get the behaviour of every earlier Node in the declared ">= 20" support range:
`node --no-experimental-require-module -e "require('@slack/web-api')"`

Control, same command, same Node:
`npm install @slack/web-api@8.0.0`
`node --no-experimental-require-module -e "require('@slack/web-api')"`  -> loads fine

### Expected Result

`require('@slack/web-api')` succeeds on every Node version in the package's declared engines.node: ">= 20" range, as it does on 8.0.0.

### Actual Result

node_modules/@slack/web-api/dist/WebClient.js:45
const p_queue_1 = __importDefault(require("p-queue"));
                                                              ^
Error [ERR_REQUIRE_ESM]: require() of ES Module node_modules/p-queue/dist/index.js
from node_modules/@slack/web-api/dist/WebClient.js not supported.
Instead change the require of index.js in .../WebClient.js to a dynamic import()
which is available in all CommonJS modules.
        at Object.<anonymous> (.../WebClient.js:45:35) {
     code: 'ERR_REQUIRE_ESM'
}

zimeg:
### Summary

This pull request restores CommonJS-loadable dependency versions so `require('@slack/web-api')` and `require('@slack/webhook')` work on every Node.js version in the declared `engines.node: ">= 20"` range. Fixes #2711.

- `@slack/web-api@8.1.0` throws `ERR_REQUIRE_ESM` when `require()`d on Node 20 versions before `require(esm)` was unflagged (i.e. Node 20.0–20.18; unflagged by default only in [20.19+](https://nodejs.org/en/blog/release/v20.19.0) / [22.12+](https://nodejs.org/en/blog/release/v22.12.0)). `@slack/web-api@8.0.0` loads fine because it shipped with the older CommonJS deps.
- Root cause: `p-queue` and `p-retry` each went **pure ESM** (`"type": "module"`) at a major that our ranges silently crossed:
  - `p-queue` — CJS through `6.6.2`; ESM-only from [**v7.0.0**](https://github.com/sindresorhus/p-queue/releases/tag/v7.0.0). We were on `^9`.
  - `p-retry` — CJS through `4.6.2`; ESM-only from [**v5.0.0**](https://github.com/sindresorhus/p-retry/releases/tag/v5.0.0). We were on `^8` ([v8.0.0](https://github.com/sindresorhus/p-retry/releases/tag/v8.0.0) also raised the engine floor to Node 22).
- Fix: pin both back to their last CommonJS releases — **`p-queue@^6.6.2`** and **`p-retry@^4.6.2`** — matching what shipped in the working `@slack/web-api@8.0.0`. `6.6.2` and `4.6.2` are the terminal versions of their CJS majors, so the caret ranges resolve to exactly those releases while staying major-gated against the ESM-only successors. Matching Dependabot `ignore` rules are added for `web-api` (mirroring the existing rule for `webhook`) so a future bump can't silently re-break CJS loading.
- **Not a breaking change for consumers:** `p-queue`/`p-retry` are internal implementation details — never re-exported. The public `RetryOptions` type is defined via the [`retry`](https://www.npmjs.com/package/retry) package, not `p-retry`, and the SDK only uses options common to these versions (`AbortError`, `retries`, `factor`, `minTimeout`, `maxTimeout`, `randomize`). Consumers on Node 20.19+/22.12+ see no change; consumers on plain Node 20 go from broken → working.

### Testing

Simulate the pre-`require(esm)` Node.js behavior with the flag from the issue repro:

```sh
# From packages/web-api, after npm run build:
node --no-experimental-require-module test/integration/commonjs-project/index.js
# → ✅ CJS project integration test succeeded!

# And directly, mirroring the issue:
node --no-experimental-require-module -e "require('@slack/web-api')"   # loads (was ERR_REQUIRE_ESM)
node --no-experimental-require-module -e "require('@slack/webhook')"   # loads (was ERR_REQUIRE_ESM)
```

### Requirements

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -186,10 +186,7 @@ updates:
           - "@types/mocha"
           - "mocha"
     ignore:
-      # p-queue (v7+) and p-retry (v5+) are pure ESM and raise their minimum
-      # Node.js version each major. Bumping either breaks `require()` of this
-      # package's CommonJS build on the declared Node >= 20 range (ERR_REQUIRE_ESM,
-      # see #2711) and can force a change to its supported Node floor.
+      # Avoid pure ESM p-queue (v7+) / p-retry (v5+) to prevent CJS require() breakage on Node >= 20 (see #2711).
       - dependency-name: "p-queue"
       - dependency-name: "p-retry"
       - dependency-name: "@types/node"
@@ -221,10 +218,7 @@ updates:
           - "@types/mocha"
           - "mocha"
     ignore:
-      # p-retry (v5+) is pure ESM and raises its minimum Node.js version each
-      # major (v7 requires Node >= 20, v8 requires Node >= 22). Bumping it breaks
-      # `require()` of this package's CommonJS build (ERR_REQUIRE_ESM, see #2711)
-      # and can force a change to its supported Node floor.
+      # Avoid pure ESM p-retry (v5+) to prevent CJS require() breakage on Node >= 20 (see #2711).
       - dependency-name: "p-retry"
       - dependency-name: "@types/node"
         versions:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -186,6 +186,12 @@ updates:
           - "@types/mocha"
           - "mocha"
     ignore:
+      # p-queue (v7+) and p-retry (v5+) are pure ESM and raise their minimum
+      # Node.js version each major. Bumping either breaks `require()` of this
+      # package's CommonJS build on the declared Node >= 20 range (ERR_REQUIRE_ESM,
+      # see #2711) and can force a change to its supported Node floor.
+      - dependency-name: "p-queue"
+      - dependency-name: "p-retry"
       - dependency-name: "@types/node"
         versions:
           - "21.x"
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
```diff
@@ -186,6 +186,12 @@ updates:
           - "@types/mocha"
           - "mocha"
     ignore:
+      # p-queue (v7+) and p-retry (v5+) are pure ESM and raise their minimum
+      # Node.js version each major. Bumping either breaks `require()` of this
+      # package's CommonJS build on the declared Node >= 20 range (ERR_REQUIRE_ESM,
+      # see #2711) and can force a change to its supported Node floor.
```
WilliamBergamin: Can we make this comment a one liner like 
```suggestion
      # Avoid pure ESM versions of p-queue (v7+) and p-retry (v5+) 
      # prevent CJS require() breakage on Node >= 20.
```

packages/web-api/package.json
```diff
@@ -53,8 +53,8 @@
     "@types/node": ">=20",
     "@types/retry": "0.12.5",
     "eventemitter3": "^5.0.1",
-    "p-queue": "^9",
-    "p-retry": "^8",
+    "p-queue": "^6.6.2",
+    "p-retry": "^4.6.2",
```
WilliamBergamin, line 57: Thank you 🙏

WilliamBergamin (approved):
Thanks for working on this 🙏  

Left one none blocking comment

.github/dependabot.yml
zimeg: @WilliamBergamin A much appreciated callout! We're landing this in https://github.com/slackapi/node-slack-sdk/pull/2712/changes/f57d7236059e1e0f4628cee71a66df7c5aae9c9c to revisit later!

PR #2722, opened 2026-09-02
fix: correct SlackListsItemFieldMessage type

### LINKED ISSUE #2598: `SlackListsItem   FieldMessage.message` type   does not match API response    shape
### Package

@slack/types

### SDK Version

@slack/types 2.21.1

### Node.js Version

v24.12.0

### Operating System

_No response_

### Steps to Reproduce

`SlackListsItemFieldMessage` in `packages/web-api/src/types/request/slackLists.ts` defines:

  ```typescript
export interface SlackListsItemFieldMessage {
    column_id: string;
    message: string[];
}
```

 The Slack API can return the message field as either:
 - A single message object: `"message": {"text":"hello", "ts": "123.456"}`
 - An array of message objects: `"message":[{"text": "a"}, {"text":"b"}]`

The current typing is inaccurate in two ways:
1. It assumes `message` is always an array, but the API may return a single object
2. It types elements as `string` when they areactually message objects

See [java-slack-sdk#1587 ](https://github.com/slackapi/java-slack-sdk/issues/1587)for the same API inconsistency on the Java side.

**Suggested fix**

```typescript
interface SlackListsItemMessage {
    text?: string;
    ts?: string;
    user?: string;
    team?: string;
    type?: string;
}

export interface SlackListsItemFieldMessage {
    column_id: string;
    message: SlackListsItemMessage | SlackListsItemMessage[];
  }


export interface SlackListsItemFieldMessage {
    column_id: string;
    message: SlackListsItemMessage | SlackListsItemMessage[];
}
```

This types it as a union so consumers handle both shapes.

**Impact**

No runtime crash (JavaScript is untyped at runtime), but developers relying on the type will write incorrect code, like calling .map() on a non-array or treating objects as strings.

**References**

- slackapi/java-slack-sdk#1587
- https://docs.slack.dev/reference/methods/slackLists.items.create#field-types

### Expected Result

N/A

### Actual Result

N/A

dajiaohuang:
## Summary

The SlackListsItemFieldMessage interface incorrectly defined the message field as string[]. However, the Slack API can return the message field as either a single message object or an array of message objects.

## Fix

This PR introduces a new SlackListsItemMessage interface and updates SlackListsItemFieldMessage to use a union type.

## References

Fixes #2598

### DIFF
--- a/packages/web-api/src/types/request/slackLists.ts
+++ b/packages/web-api/src/types/request/slackLists.ts
@@ -58,11 +58,22 @@ export interface SlackListsItemFieldLink {
 }
 
 /**
+ * @description Message object in Slack Lists message field.
+ */
+export interface SlackListsItemMessage {
+  text?: string;
+  ts?: string;
+  user?: string;
+  team?: string;
+  type?: string;
+}
+
+/**
  * @description Message field with message URLs.
  */
 export interface SlackListsItemFieldMessage {
   column_id: string;
-  message: string[];
+  message: SlackListsItemMessage | SlackListsItemMessage[];
 }
 
 /**


vegeris:
Hi there, 

This update was tried before but broke the request side: https://github.com/slackapi/node-slack-sdk/issues/2598#issuecomment-4840812085

PR #2724, opened 2026-09-02
fix: safely extract error message from unknown error type

dajiaohuang:
## Summary

The code in `packages/oauth/src/install-provider.ts` assumed that caught errors (`e`) are always Error objects and accessed `e.message` directly. If `e` was a primitive value (string, number) or undefined, this could result in undefined being passed to `GenerateInstallUrlError`.

## Problem

At line 407:
```typescript
throw new GenerateInstallUrlError((e as any).message);
```

If `e` is not an Error object (e.g., `throw "something went wrong"` or `throw null`), then `(e as any).message` would be `undefined`.

## Fix

Used the same safe pattern already used elsewhere in this file (line 289):
```typescript
const errorMessage = e instanceof Error ? e.message : String(e);
throw new GenerateInstallUrlError(errorMessage);
```

This ensures a valid string is always passed to `GenerateInstallUrlError`, whether `e` is an Error object or a primitive value.

## Testing

- [ ] Run existing tests: `npm test --workspace=packages/oauth`

### DIFF
--- a/packages/web-api/src/WebClient.ts
+++ b/packages/web-api/src/WebClient.ts
@@ -799,7 +799,12 @@ export class WebClient extends Methods {
       // if it isn't a Gzip response but is from the admin.analytics.getFile request,
       // decode the ArrayBuffer to JSON read the error
       const buffer = await response.arrayBuffer();
-      data = JSON.parse(new TextDecoder().decode(buffer));
+      try {
+        data = JSON.parse(new TextDecoder().decode(buffer));
+      } catch (_) {
+        // failed to parse the response body as JSON
+        data = { ok: false, error: new TextDecoder().decode(buffer) };
+      }
     } else {
       const text = await response.text();
       try {
--- a/packages/oauth/src/install-provider.ts
+++ b/packages/oauth/src/install-provider.ts
@@ -401,10 +401,10 @@ export class InstallProvider {
         res.end(body);
       }
     } catch (e: unknown) {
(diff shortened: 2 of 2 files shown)


packages/web-api/src/WebClient.ts
```diff
@@ -799,7 +799,12 @@ export class WebClient extends Methods {
       // if it isn't a Gzip response but is from the admin.analytics.getFile request,
       // decode the ArrayBuffer to JSON read the error
       const buffer = await response.arrayBuffer();
-      data = JSON.parse(new TextDecoder().decode(buffer));
+      try {
+        data = JSON.parse(new TextDecoder().decode(buffer));
+      } catch (_) {
+        // failed to parse the response body as JSON
+        data = { ok: false, error: new TextDecoder().decode(buffer) };
+      }
```
vegeris, line 807: I see this is a duplicate of your other PR, https://github.com/slackapi/node-slack-sdk/pull/2723/changes. I'm going to close the other one

vegeris: Same nit as posted to the other PR, `could move const text = new TextDecoder().decode(buffer); above to avoid decoding twice on error`

vegeris: And the need for signing the CLA

vegeris:
Closing as a duplicate of https://github.com/slackapi/node-slack-sdk/pull/2725/changes#r3927012436

PR #2725, opened 2026-09-02
fix: use consistent URL validation pattern in IncomingWebhook

dajiaohuang:
## Summary

The URL validation in `IncomingWebhook` used `if (url === undefined)` which only catches undefined values. Meanwhile, `WebhookTrigger` in the same package uses the more robust `if (!url)` which catches undefined, null, and empty string.

## Problem

At line 74 in `IncomingWebhook.ts`:
```typescript
if (url === undefined) {
  throw new Error('Incoming webhook URL is required');
}
```

This check would not catch `null` or empty string `""` values.

## Fix

Changed to use the same pattern as `WebhookTrigger.ts`:
```typescript
if (!url) {
  throw new Error('Incoming webhook URL is required');
}
```

This ensures consistent validation across the webhook package and catches all falsy values.

## Testing

- [ ] Run existing tests: `npm test --workspace=packages/webhook`

### DIFF
--- a/packages/webhook/src/IncomingWebhook.ts
+++ b/packages/webhook/src/IncomingWebhook.ts
@@ -71,7 +71,7 @@ export class IncomingWebhook {
       timeout: 0,
     },
   ) {
-    if (url === undefined) {
+    if (!url) {
       throw new Error('Incoming webhook URL is required');
     }
 
--- a/packages/oauth/src/install-provider.ts
+++ b/packages/oauth/src/install-provider.ts
@@ -401,10 +401,10 @@ export class InstallProvider {
         res.end(body);
       }
     } catch (e: unknown) {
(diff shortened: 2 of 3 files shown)


zimeg:
👋 Hey @dajiaohuang! Before we review this PR I'm wondering if it'd be possible to sign the CLA above? 🏛️ ✨

packages/webhook/src/IncomingWebhook.ts
```diff
@@ -71,7 +71,7 @@ export class IncomingWebhook {
       timeout: 0,
     },
   ) {
-    if (url === undefined) {
+    if (!url) {
```
vegeris, line 74: I see this PR is also a duplicate of your other PR, https://github.com/slackapi/node-slack-sdk/pull/2724. I will close the other one. In the future, do not spam our repo with multiple PRs for the same changes.

